PR #1, opened 2018-10-03
Initial actor structure

djfr:
Initial actor structure draft.

I separated the actors into different DLLs this time, because here performance will be a __thing__, so we have slightly faster actor ramp up time at the cost of more LoC.


src/EventHandlerActor/ActorEventSource.cs
```diff
@@ -0,0 +1,165 @@
+﻿namespace EventHandlerActor
+{
+    using System;
+    using System.Diagnostics.Tracing;
+    using System.Threading.Tasks;
+    using Microsoft.ServiceFabric.Actors.Runtime;
+
+    [EventSource(Name = "MyCompany-CaptainHook-EventHandlerActor")]
+    internal sealed class ActorEventSource : EventSource
```
VaclavK: I would start weaning this off...

djfr: I will when I plug BB there!

PR #11, opened 2019-02-06
Feature/updating to reminders

dguerin:


src/CaptainHook.EventReaderActor/EventReaderActor.cs
```diff
@@ -169,9 +164,19 @@ internal void ReadEvents(object _)
         /// <remarks>
         /// Do nothing by design. We just need to make sure that the actor is properly activated.
         /// </remarks>>
-        public Task Run()
+        public async Task Run()
```
arturek: Maybe it would be better to replace Run with Enable and Disable calls which respectively create or remove the reminder? Even for troubleshooting proposes (and to remove remarks section above)

PR #12, opened 2019-02-15
Feature/dynamic routing

dguerin:


src/CaptainHook.Common/Configuration/EventHandlerConfig.cs
```diff
@@ -21,13 +21,9 @@ public WebhookConfig()
         public string Name { get; set; }
 
         //todo implement this on the calls to the webhook to select http verb
-        public string Verb { get; set; }
+        public string HttpVerb { get; set; }
```
VaclavK: enum?

src/CaptainHook.Common/Configuration/EventHandlerConfig.cs
```diff
@@ -74,24 +60,86 @@ public class EventParser
...
+    }
 
-        public QueryLocation QueryLocation { get; set; }
+    public enum DataType
+    {
+        Property = 1,
+        HttpContent = 2,
+        HttpStatusCode = 3,
+        Model = 4,
+        String = 5
```
VaclavK: what is the diff between string vs. property?

src/CaptainHook.Common/Nasty/ModelParser.cs
```diff
@@ -1,49 +1,79 @@
...
+        public static JToken GetJObject(object payload, DataType destinationType = DataType.Model)
+        {
+            if (destinationType == DataType.String)
+            {
+                return payload as string;
+            }
+
+            if (payload is string payloadAsString)
+            {
+                return JObject.Parse(payloadAsString);
```
VaclavK: exception handling/telemetry for invalid json?

src/CaptainHook.EventHandlerActor/Handlers/StringExtensions.cs
```diff
@@ -0,0 +1,23 @@
...
+        public static int LastIndexOfSafe(this string value, char searchCharacter)
+        {
+            try
+            {
+                var position = value.LastIndexOf("/", StringComparison.Ordinal);
+                return position;
+            }
+            catch
+            {
+                // ignored
```
VaclavK: hmmmmm :-D

src/CaptainHook.EventReaderActor/EventReaderActor.cs
```diff
@@ -92,12 +92,11 @@ protected override async Task OnActivateAsync()
                     new RetryExponential(TimeSpan.FromMilliseconds(100), TimeSpan.FromMilliseconds(500), 3),
                     BatchSize);
 
-                _wakeupReminder = await this.RegisterReminderAsync(
-                    WakeUpReminderName,
+                _poolTimer = new Timer(
+                    ReadEvents,
                     null,
-                    TimeSpan.FromMilliseconds(100),
+                    TimeSpan.FromMilliseconds(1000),
```
VaclavK: why not make these (and there was another instance) above part of service config package?

src/CaptainHook.EventReaderActor/EventReaderActor.cs
```diff
@@ -172,9 +171,13 @@ internal void ReadEvents()
...
         /// Do nothing by design. We just need to make sure that the actor is properly activated.
         /// </remarks>>
-        public Task Run()
+        public async Task Run()
         {
-            return Task.CompletedTask;
+            _wakeupReminder = await this.RegisterReminderAsync(
+                WakeUpReminderName,
+                null,
+                TimeSpan.FromMinutes(15),
```
VaclavK: config?

src/CaptainHook.EventReaderActor/EventReaderActor.cs
```diff
@@ -197,7 +200,7 @@ public Task ReceiveReminderAsync(string reminderName, byte[] state, TimeSpan due
         {
             if (reminderName.Equals(WakeUpReminderName, StringComparison.OrdinalIgnoreCase))
             {
-                ReadEvents();
+                ReadEvents(null);
```
VaclavK: default param value?

src/CaptainHook.Common/Configuration/EventHandlerConfig.cs
dguerin: yeah I can change this. Config builder didn't support enums previously but it does now so I'll change all them below for the ones which I have not changed already.

src/CaptainHook.EventReaderActor/EventReaderActor.cs
dguerin: As a general rule, anything that is not in EventHandlerConfig is getting a light touch. This will all change in V1 so I do not want to spend to much time on V0 stuff which will get blown away.

src/CaptainHook.EventReaderActor/EventReaderActor.cs
dguerin: This is just a temp fix to keep the actor alive polling the subscription. After the 1 hour default going to sleep and waking up was a problem for the actor. This reminder keeps them alive indefinitely so they are always available to poll the subscriptions.

src/CaptainHook.Common/Nasty/ModelParser.cs
dguerin: sure

src/CaptainHook.Common/Configuration/EventHandlerConfig.cs
dguerin: property can be bool, int, enum etc etc. String is not. Can make this clearer. I have a bunch of docs to write up around this work. But I won't get to that until I have the configuration api done and see what that looks like when I connect the api to this builder. It should be clearer then.

PR #18, opened 2019-04-11
Feature/timeout extensions

dguerin:


src/CaptainHook.Common/Configuration/EventHandlerConfig.cs
```diff
@@ -24,6 +24,12 @@ public WebhookConfig()
         public HttpVerb HttpVerb { get; set; }
 
         public List<WebhookRequestRule> WebhookRequestRules { get; set; }
+
+        /// <summary>
+        /// Request duration maximum timeout in seconds
+        /// Left at 100 seconds as the default value for the http client timeout
+        /// </summary>
+        public int TimeoutInSeconds { get; set; } = 100;
```
arturek: maybe instead of "int TimeoutInSeconds" use "TimeSpan Timeout"?

src/CaptainHook.PoolManagerActor/PoolManagerActor.cs
```diff
@@ -47,47 +47,58 @@ public PoolManagerActor(ActorService actorService, ActorId actorId, IBigBrother
         /// </summary>
         protected override async Task OnActivateAsync()
         {
-            _free = await StateManager.TryGetStateAsync<HashSet<int>>("free");
+            _bigBrother.Publish(new ActorActivated(this));
+            var free = await StateManager.TryGetStateAsync<HashSet<int>>(nameof(_free));
```
arturek: if the name of a private variable is used here, than any renaming will cause issues but won't be flagged as a problem

src/CaptainHook.PoolManagerActor/PoolManagerActor.cs
```diff
@@ -47,47 +47,58 @@ public PoolManagerActor(ActorService actorService, ActorId actorId, IBigBrother
...
             else
             {
-                _free = new ConditionalValue<HashSet<int>>(true, new HashSet<int>());
-                _busy = new ConditionalValue<Dictionary<Guid, MessageHook>>(true, new Dictionary<Guid, MessageHook>());
+                _free = new HashSet<int>(NumberOfHandlers);
+                _busy = new Dictionary<Guid, MessageHook>(NumberOfHandlers);
 
                 for (var i = 0; i < NumberOfHandlers; i++)
                 {
                     ActorProxy.Create<IEventHandlerActor>(new ActorId(i)); // TODO: this probably isn't needed here since we're not invoking the actor at this point - REVIEW
```
arturek: it doesn't create an actor. only a proxy which is GCed before being used.

src/CaptainHook.PoolManagerActor/PoolManagerActor.cs
```diff
@@ -47,47 +47,58 @@ public PoolManagerActor(ActorService actorService, ActorId actorId, IBigBrother
...
+        }
+
         public async Task<Guid> DoWork(string payload, string type)
         {
             // need to handle the possibility of the resources in the pool being all busy!
 
             try
             {
-                var handlerId = _free.Value.First();
+                var handlerId = _free.First();
```
arturek: may _free be ever empty?

dguerin: yes it will, hence the comment above. the logic here needs to be addressed such that certain connections can be throttle here while others can be concurrent. it's a much bigger piece of work but not addressed here.

src/CaptainHook.Common/Configuration/EventHandlerConfig.cs
dguerin: this gets pulled in from KV config builder so the value there will be in seconds. Why would you want to change that?

src/CaptainHook.PoolManagerActor/PoolManagerActor.cs
arturek: So maybe throwing some other exception (or handling it in other way) rather than a general one from First() would be better?

src/CaptainHook.PoolManagerActor/PoolManagerActor.cs
dguerin: yeah I've cleaned this up. Nice one

src/CaptainHook.PoolManagerActor/PoolManagerActor.cs
dguerin: added an explicit exception, it's something we have to come back to

src/CaptainHook.Common/Configuration/EventHandlerConfig.cs
dguerin: I've changed this but I just want to test this out a little before we merge

dguerin: works a treat.

PR #19, opened 2019-04-17
Bug fix and pool manger tests

dguerin:
* Fixing some bugs in the pool manger flow
* Added some actor tests
* Changed event name

### DIFF
--- a/src/CaptainHook.PoolManagerActor/PoolManagerActor.cs
+++ b/src/CaptainHook.PoolManagerActor/PoolManagerActor.cs
@@ -24,7 +24,7 @@ namespace CaptainHook.PoolManagerActor
     public class PoolManagerActor : Actor, IPoolManagerActor
     {
         private readonly IBigBrother _bigBrother;
-        private readonly IActorProxyFactory _mockActorProxyFactory;
+        private readonly IActorProxyFactory _actorProxyFactory;
         private HashSet<int> _free; // free pool resources
         private Dictionary<Guid, MessageHook> _busy; // busy pool resources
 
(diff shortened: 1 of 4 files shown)


src/CaptainHook.PoolManagerActor/PoolManagerActor.cs
```diff
@@ -24,6 +24,7 @@ namespace CaptainHook.PoolManagerActor
     public class PoolManagerActor : Actor, IPoolManagerActor
     {
         private readonly IBigBrother _bigBrother;
+        private readonly IActorProxyFactory _mockActorProxyFactory;
```
arturek: why 'mock...'? It's a standard actor proxy factory. It's mocked only when a test is executed.

src/CaptainHook.PoolManagerActor/PoolManagerActor.cs
```diff
@@ -64,14 +67,20 @@ protected override async Task OnActivateAsync()
                 _free = new HashSet<int>(NumberOfHandlers);
                 _busy = new Dictionary<Guid, MessageHook>(NumberOfHandlers);
 
-                for (var i = 0; i < NumberOfHandlers; i++)
+                for (var i = 1; i < NumberOfHandlers + 1; i++)
```
arturek: maybe `Enumerable.Range(1, 5).ToHashSet()`

src/CaptainHook.PoolManagerActor/PoolManagerActor.cs
dguerin: let me check, had it there for another reason but might be able to remove it.

src/CaptainHook.PoolManagerActor/PoolManagerActor.cs
dguerin: lol.

PR #21, opened 2019-04-18
Feature/better failure flows

dguerin:


dguerin:
@arturek could you take a look at this when you get a chance? per out conversations before I actually think I introduced a few bugs where the pool manager locks up the flow into the handlers because they were not a one to one mapping. Some exceptions in the handlers were also locking them up but those paths should be handled now to release a lock in the pool.

PR #24, opened 2019-06-21
Authentication Config Selection

dguerin:
* Update to select authentication scheme depending on message body selection.
* Move HttpClient creation out of IOC into cached factory for better reuse.
* Parsing config to create just a single http client per host, including parsing multi route destinations


dguerin:
I was going to refactor out the httpclientfactory I have put in but given the a webhook has a number of routes I need to know which route is being selected based on the contents of the message and then select the appropriate webhook config for it. Not going to clean this up now V2 has a replacement for all of this so would be wasted effort. @arturek @djfr @VaclavK

dguerin:
> I was going to refactor out the httpclientfactory I have put in but given the a webhook has a number of routes I need to know which route is being selected based on the contents of the message and then select the appropriate webhook config for it. Not going to clean this up now V2 has a replacement for all of this so would be wasted effort. @arturek @djfr @VaclavK

Actually nvm, I've thought of a cleaner way to do it now.

dguerin:
@VaclavK @arturek need to push on with this so going to merge it. If you wish to discuss anything do let me know.

PR #27, opened 2019-08-07
Cleanup up logging

dguerin:


dguerin:
> for filtering purposes, should we expose the code as a field + correlation id?

What I'm aiming for here @VaclavK is the telemetry work @arturek is doing with Actors can be extended to the whole flow for CH. This would include emitting the correlation vector for AI here too.

At the moment if you put it in, it will be useless as there is no link back to the original published message.

PR #30, opened 2019-08-26
Feature/logging request request

dguerin:
* Bug fix for correlation id which was breaking AFD
* More tests
* Refactor a bunch of flows
* Added more logging to failure flows


src/CaptainHook.EventHandlerActor/Handlers/HttpClientBuilder.cs
```diff
@@ -0,0 +1,60 @@
...
+        {
+            var uri = new Uri(config.Uri);
+
+            if (!_httpClients.TryGetValue(uri.Host, out var httpClient))
+            {
+                throw new ArgumentNullException(nameof(httpClient), $"HttpClient for {uri.Host} was not found");
+            }
+
+            httpClient.DefaultRequestHeaders.Remove(Constants.Headers.CorrelationId);
+            httpClient.DefaultRequestHeaders.Add(Constants.Headers.CorrelationId, correlationId);
```
hudo: Think this could be lead to problems since DefaultHeaders will be used for all requests for this host (not sure whats the life scope of this HttpClient...). 
We should maybe set HttpRequestMessage.Headers here?

src/CaptainHook.EventHandlerActor/Handlers/RequestLogger.cs
```diff
@@ -0,0 +1,71 @@
...
+                messageData.CorrelationId));
+        }
+
+        private static async Task<string> GetPayloadAsync(HttpResponseMessage response)
+        {
+            if (response?.Content == null)
+            {
+                return string.Empty;
+            }
+            var result = await response.Content.ReadAsStringAsync();
```
hudo: It would be cool to move from passing strings around to passing Streams or maybe even that new Span<string> if that even makes sense. Have a feeling that we're putting lot of preassure on GC, but would be good to measure before we plan to do anything ... Just a mental note, nothing to change here...

src/CaptainHook.EventHandlerActor/Handlers/HttpClientBuilder.cs
VaclavK: only commenting on the http client - in conjunction with http client factory - the client is - as per MS recommendations - very transient resource, and only a lightweight wrapper around handlers, which is where the heavy lifting is done and are (pool) managed by the factory

dguerin: I suspect in the older term when telemetry is correctly done - this will just get blown away. At the moment the HTTPClient is long lived per each endpoint - reuse and all that good stuff.

I was looking at putting in the httpclient factory but this requires aspnet packages which don't have a place in the actor framework.

src/CaptainHook.EventHandlerActor/Handlers/RequestLogger.cs
dguerin: Yeah I had thought about that but we would need to see how AI handles this, Ideally the memory footprint of a message should be light. In this case its the failure flow so it's expected failure is lighter than success paths so the memory footprint is just lighter. 

I'll add application memory to the dashboards so we can get a feel for memory and CPU footprint when I get a chance and we can see what we need to optimise there - should we need to.

src/CaptainHook.EventHandlerActor/Handlers/HttpClientBuilder.cs
dguerin: just looked at the package dependencies for the httpclientfactory again and I think it can go in without the dependencies I mentioned. I'll take a look at that next.

dguerin: That being said I was basing the implication for a persisted client on this https://docs.microsoft.com/en-us/aspnet/web-api/overview/advanced/calling-a-web-api-from-a-net-client#create-and-initialize-httpclient

And this https://aspnetmonsters.com/2016/08/2016-08-27-httpclientwrong/

VaclavK: you still have to consider DNS cache and long live instances becoming out of sync so some of these suggestions above can help with socket but not the DNS part

I think we all agree that factory does implement correct lifecycle/pool management for handlers

dguerin: yeah it is and it does

dguerin: There doesn't seem to be a good way of putting this into a Service Fabric Actor. Most examples I've come across suggust using the IHostBuilder but that's a much larger change. I'll backlog something for this and we can come back to it again.

dguerin:
I'm merging this in. If there is any other issue than the one mentioned above about the httpclientfactory let me know and we can discuss.

PR #33, opened 2019-10-03
Auth401 investigation

VaclavK:
changing the advance expiry timespan to 50seconda
logging the token in case of 401s for diagnostic purposes

I was initially using getter on http client default request headers authorization field but it occassionally throws InvalidCastException as MS code expects it always to be a header object but for some reason it is occassionally List<object> .. weird I know

{"Message":"Unable to cast object of type 'System.Collections.Generic.List`1[System.Object]' to type 'System.Net.Http.Headers.AuthenticationHeaderValue'.","ClassName":"System.InvalidCastException","CallerLineNumber":"0","HResult":"-2147467262","Source":"System.Net.Http","RemoteStackIndex":"0","StackTraceString":"at System.Net.Http.Headers.HttpRequestHeaders.get_Authorization()\r\n at CaptainHook.EventHandlerActor.Handlers.RequestLogger.LogAsync(HttpClient httpClient, HttpResponseMessage response, MessageData messageData, Uri uri, HttpVerb httpVerb) in F:\\17\\s\\src\\CaptainHook.EventHandlerActor\\Handlers\\RequestLogger.cs:line 0\r\n at CaptainHook.EventHandlerActor.Handlers.GenericWebhookHandler.CallAsync[TRequest](TRequest request, IDictionary`2 metadata, CancellationToken cancellationToken) in F:\\17\\s\\src\\CaptainHook.EventHandlerActor\\Handlers\\GenericWebhookHandler.cs:line 66\r\n at CaptainHook.EventHandlerActor.EventHandlerActor.InternalHandle(Object state) in F:\\17\\s\\src\\CaptainHook.EventHandlerActor\\EventHandlerActor.cs:line 109"}


src/CaptainHook.EventHandlerActor/Handlers/RequestLogger.cs
```diff
@@ -27,17 +27,28 @@ public RequestLogger(IBigBrother bigBrother)
             HttpResponseMessage response,
             MessageData messageData,
             Uri uri,
-            HttpVerb httpVerb
+            HttpMethod httpMethod
         )
         {
+            string token = string.Empty;
+
+            if (response.StatusCode == HttpStatusCode.Unauthorized)
```
dguerin: What is the plan for future here? I don't think we'll need this when we have all the issues resolved?

src/CaptainHook.EventHandlerActor/Handlers/RequestLogger.cs
```diff
@@ -27,17 +27,28 @@ public RequestLogger(IBigBrother bigBrother)
...
+            {
+                var values = httpClient.DefaultRequestHeaders.GetValues("Authorization");
```
dguerin: Is this populated anymore given how the request headers are populated on the httprequestmessage?

src/CaptainHook.EventHandlerActor/Handlers/RequestLogger.cs
VaclavK: with the other PR, this PR can be closed **but** I think that logging the header in case of 401s is a potentially useful feature to transfer for diagnostics.... @dguerin @arturek comments?

VaclavK: so the use case here I would consider is

especially with API, if the hook is misconfigured from security perspective, we would be able to "reverse engineer" the header (jwt.io and the likes) and prove why the flow is broken etc.,

src/CaptainHook.EventHandlerActor/Handlers/RequestLogger.cs
VaclavK: yeah if we want to keep the logging - as the suggestion in the other comment - this would have to be adjusted

src/CaptainHook.Common/Authentication/OidcAuthenticationConfig.cs
```diff
@@ -35,6 +35,6 @@ public OidcAuthenticationConfig()
         /// <summary>
         /// Refresh interval before the token expires
         /// </summary>
-        public int RefreshBeforeInSeconds { get; set; } = 10;
+        public int RefreshBeforeInSeconds { get; set; } = 50;
```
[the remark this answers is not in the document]
VaclavK: I think we are ok now..

my thinking is that 5 seconds is defo too low - given this is all fully async - and even async within async - task chains - we do not have precise control over when is the request actually sent over the wire and as such the check for expiration could succeed but the request itself not.....

re the 50 vs. 10.. the diff is IMHO minimal - 40/3600 :D.. any extra pressure on STS is negligible IMHO...

having said that... we can always adjust in the future so not worried about this one now - now that we know the mechanics of the 401 failures experienced before

PR #34, opened 2019-10-03
Feature/http heads refactor

dguerin:
* Added mechanism to add http headers dynamically
* Http Clients are now created on the fly and saved for reuse
* Authentication Factory saves on per endpoint basis.


src/CaptainHook.Common/Constants.cs
```diff
@@ -5,6 +5,16 @@ public struct Constants
         public struct Headers
         {
             public const string CorrelationId = "X-Correlation-ID";
+
+            public const string ContentType = "Content-Type";
+
+            public const string EventType = "Esw-Event-Type";
+
+            public const string EventDeliveryId = "Esw-Delivery";
```
VaclavK: why not include Id in the header name just like for correlation id?

src/CaptainHook.EventHandlerActor/Handlers/Authentication/MmAuthenticationHandler.cs
```diff
@@ -39,11 +39,19 @@ public override async Task GetTokenAsync(HttpClient client, CancellationToken to
...
-            client.DefaultRequestHeaders.TryAddWithoutValidation("client_secret", OidcAuthenticationConfig.ClientSecret);
+            var httpClient = HttpClientFactory.Get(OidcAuthenticationConfig.Uri);
 
-            var authProviderResponse = await client.PostAsync(OidcAuthenticationConfig.Uri, new StringContent("", Encoding.UTF32, "application/json-patch+json"), token);
+            var headers = new WebHookHeaders();
+            headers.AddContentHeader(Constants.Headers.ContentType, "application/json-patch+json");
+            headers.AddRequestHeader("client_id", OidcAuthenticationConfig.ClientId);
+            headers.AddRequestHeader("client_secret", OidcAuthenticationConfig.ClientSecret);
+
+            var authProviderResponse = await httpClient.SendRequestReliablyAsync(
```
VaclavK: I think this can throw exceptions even within polly (e.g. socket exceptions from the http client itself) or can't?

src/CaptainHook.EventHandlerActor/Handlers/HttpClientExtensions.cs
```diff
@@ -18,127 +18,39 @@ public static class HttpClientExtensions
...
-
-                case HttpVerb.Patch:
-                    return await client.PatchAsJsonReliably(uri, payload, logger, contentType, token);
+            var request = new HttpRequestMessage(httpMethod, uri);
 
-                default:
-                    throw new ArgumentOutOfRangeException(nameof(httpVerb), httpVerb, "no valid http verb found");
+            if (httpMethod != HttpMethod.Get)
+            {
+                request.Content = new StringContent(payload, Encoding.UTF8, webHookHeaders.ContentHeaders[Constants.Headers.ContentType]);
```
VaclavK: and I assume it will be present all the time?

src/CaptainHook.EventHandlerActor/Handlers/RequestBuilder.cs
```diff
@@ -224,5 +226,18 @@ public WebhookConfig SelectWebhookConfig(WebhookConfig webhookConfig, string pay
...
         }
+
+        /// <inheritdoc />
+        public WebHookHeaders GetHeaders(WebhookConfig webhookConfig, MessageData messageData)
+        {
+            var headers = new WebHookHeaders();
+
+            headers.AddContentHeader(Constants.Headers.ContentType, webhookConfig.ContentType);
+            headers.AddRequestHeader(Constants.Headers.CorrelationId, messageData.CorrelationId);
+            headers.AddRequestHeader(Constants.Headers.EventDeliveryId, messageData.CorrelationId);
```
VaclavK: so why do we introduce new header being a copy?

src/CaptainHook.Common/Constants.cs
dguerin: It's based on this https://devportal.eshopworld.com/reference/order-confirmation-webhook#section-delivery-header - which is already predefined. i agree though otherwise

src/CaptainHook.EventHandlerActor/Handlers/Authentication/MmAuthenticationHandler.cs
dguerin: it can. it's just a simple wrapper around some polly transient tries. more work will be done on this later I suspect to allow different configurable reties, so I won't spend too much time on it now.

src/CaptainHook.EventHandlerActor/Handlers/HttpClientExtensions.cs
dguerin: By default yes - will always be application/json unless specified otherwise. FYI => https://github.com/eShopWorld/captain-hook/pull/34/files#diff-81abfae5d4577c3574f7b577b61e2921R39

src/CaptainHook.EventHandlerActor/Handlers/RequestBuilder.cs
dguerin: not sure what you mean by this one.

VaclavK: two headers with same value

dguerin: That's intention for now until the EventDeliveryId is wired up correctly with the publisher flow. Additional the correlationid will be removed when the telemetry work for SF is done.

PR #37, opened 2019-10-14
log auth token on failure

VaclavK:


VaclavK:
@dguerin I spotted this in RequestLogger - 

_bigBrother.Publish(new FailedWebHookEvent(
                httpClient.DefaultRequestHeaders.ToString(),

that is no longer correct right?

PR #40, opened 2019-11-11
Return to Sender aka DLQ processing

VaclavK:


src/CaptainHook.Common/IWrapperPayloadContract.cs
```diff
@@ -0,0 +1,21 @@
+﻿using System.Net;
+using Newtonsoft.Json.Linq;
+
+namespace CaptainHook.Common
+{
+    public interface IWrapperPayloadContract
+    {
+        JObject Payload { get; set; }        
+        string MessageId { get; set; }
```
dguerin: Do we need this? Won't the DLQ message have it's one MessageId?

VaclavK: SB message does but it would get lost when sent to the webook over HTTP/JSON ... hence I include it for tracing and other potential purposes

src/CaptainHook.Common/Configuration/EventHandlerConfig.cs
```diff
@@ -56,6 +56,15 @@ public WebhookConfig()
         /// Left at 100 seconds as the default value for the http client timeout
         /// </summary>
         public TimeSpan Timeout { get; set; } = new TimeSpan(0, 0, 100);
+
+        public PayloadContractTypeEnum PayloadTransformation { get; set; } = PayloadContractTypeEnum.Raw;
+    }
+
+    //signifies what kind of contract to use for webhook
```
arturek: Could you transform this and other comments in this class to summary comments?

src/CaptainHook.EventReaderService/EventReaderService.cs
```diff
@@ -181,9 +177,9 @@ private async Task SetupServiceBus()
         {
             ServicePointManager.DefaultConnectionLimit = 100;
 
-            await _serviceBusManager.CreateAsync(_settings.AzureSubscriptionId, _settings.ServiceBusNamespace, _subscriberName, _eventType);
-            
-            var messageReceiver = _serviceBusManager.CreateMessageReceiver(_settings.ServiceBusConnectionString, _eventType, _subscriberName);
+            await _serviceBusManager.CreateAsync(_settings.AzureSubscriptionId, _settings.ServiceBusNamespace, _initData.SubscriptionName, _initData.EventType);
+
+            var messageReceiver = _serviceBusManager.CreateMessageReceiver(_settings.ServiceBusConnectionString, _initData.EventType, _initData.SubscriptionName, _initData.DLQMode!=null);
```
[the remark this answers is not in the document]
VaclavK: that was already there before my changes and I did not touch that

PR #41, opened 2019-12-02
Telemetry

VaclavK:


src/CaptainHook.EventReaderService/Program.cs
```diff
@@ -35,22 +38,21 @@ private static async Task Main()
...
                 builder.Register<IActorProxyFactory>(_ => new ActorProxyFactory());
-                builder.RegisterServiceFabricSupport();
+
+                builder.SetupFullTelemetry(settings.InstrumentationKey);
                 builder.RegisterStatefulService<EventReaderService>(ServiceNaming.EventReaderServiceType);
 
-                using (builder.Build())
+                using (var container = builder.Build())
                 {
+                    var initializers = container.Resolve<IEnumerable<ITelemetryInitializer>>().ToList();
```
arturek: it can be deleted

src/CaptainHook.Interfaces/FabricTransport.cs
```diff
@@ -1,3 +1,7 @@
 ﻿[assembly: Microsoft.ServiceFabric.Actors.Remoting.FabricTransport.FabricTransportActorRemotingProvider(
```
arturek: Shouldn't the version ..WithTelemetry be used into both cases here? It only modifies creation of listeners so it should be safe for all projects.

src/nuget.config
```diff
@@ -0,0 +1,13 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<configuration>
+  <config>
+    <add key="repositoryPath" value="..\packages" />
+  </config>
+  <packageSources>
+    <add key="nuget.org" value="https://api.nuget.org/v3/index.json" protocolVersion="3" />
+    <add key="eswmyget-nuget-dev" value="https://eshopworld.myget.org/F/nuget-dev/api/v3/index.json" />
```
dguerin: I'm not sure it's ok to have all of these listed as such, won't they conflict if there is a dev version of the package in one feed and a release in another. Which ever is resolved first it would get used?

src/CaptainHook.EventHandlerActor/PackageRoot/ServiceManifest.xml
```diff
@@ -10,7 +10,7 @@
             <ReplicatorConfigSection Name="EventHandlerActorServiceReplicatorConfig" />
             <ReplicatorSecurityConfigSection Name="EventHandlerActorServiceReplicatorSecurityConfig" />
             <StoreConfigSection Name="EventHandlerActorServiceLocalStoreConfig" />
-            <ServiceEndpointV2_1 Name="EventHandlerActorServiceEndpointV2_1" />
+            <ServiceEndpointV2 Name="EventHandlerActorServiceEndpointV2" />
```
dguerin: SF likes renaming this on you.

arturek: The V1 compatibility mode is no longer supported so the endpoints are only V2.

VaclavK: yeah we can remove those... VS/tooling got it "out of sync"

src/nuget.config
VaclavK: I will exclude this for now...the intention was to enable the "beta" package flow which - when then promoted - we would not push to dev feeds...

PR #42, opened 2019-12-16
Dlq event type change and more

VaclavK:
use the platform events package for wrapper contract
fix auth config lookup for dlq subscriber
also fix the event type header send for DLQ


src/CaptainHook.EventHandlerActor/Handlers/RequestBuilder.cs
```diff
@@ -248,7 +249,7 @@ public WebHookHeaders GetHttpHeaders(WebhookConfig webhookConfig, MessageData me
             headers.AddContentHeader(Constants.Headers.ContentType, webhookConfig.ContentType);
             headers.AddRequestHeader(Constants.Headers.CorrelationId, messageData.CorrelationId);
             headers.AddRequestHeader(Constants.Headers.EventDeliveryId, messageData.CorrelationId);
-            headers.AddRequestHeader(Constants.Headers.EventType, webhookConfig.EventType);
+            headers.AddRequestHeader(Constants.Headers.EventType, webhookConfig.PayloadTransformation==PayloadContractTypeEnum.WrapperContract ? typeof(NewtonsoftDeliveryStatusMessage).FullName.ToLowerInvariant(): webhookConfig.EventType);
```
VaclavK: @dguerin so we need to decide what value will the header have for the wrapper contract.. right now I use the NewtonsoftDeliveryStatusMessage but I have some reservations around it... future proofing  mostly....

dguerin: yeah I think something more intuitive would make sense here. and it make make sense to have two, one for the callback and one for DLQ? 

Thinking here when/if the endpoint for receiving a message is the same endpoint used for both types of messages.

VaclavK: remember that what it is is in the payload (wrapper).. whether callback or DLQ call

PR #44, opened 2020-02-20
Fix hook and callback delivery failure handling

VaclavK:
This changes the flow for new definition of codes that are considered "failed delivery"  - 5XX and 429

also repointed packages folder to structure to match what is in the sfproj file

### DIFF
--- a/src/CaptainHook.EventHandlerActor/Handlers/HttpClientExtensions.cs
+++ b/src/CaptainHook.EventHandlerActor/Handlers/HttpClientExtensions.cs
@@ -66,21 +66,21 @@ namespace CaptainHook.EventHandlerActor.Handlers
         private static async Task<HttpResponseMessage> RetryRequest(
             Func<Task<HttpResponseMessage>> makeTheCall)
         {
+            
             //todo the retry status codes need to be customisable from the webhook config api
             var response = await Policy.HandleResult<HttpResponseMessage>(
-                    message =>
-                        message.StatusCode == HttpStatusCode.ServiceUnavailable ||
-                        message.StatusCode == HttpStatusCode.TooManyRequests)
+                    message => message.IsDeliveryFailure())                      
 
                 .WaitAndRetryAsync(new[]
                 {
                     //todo config this + jitter
                     TimeSpan.FromSeconds(20),
                     TimeSpan.FromSeconds(30)
-
                 }).ExecuteAsync(makeTheCall.Invoke);
-
+        
             return response;
         }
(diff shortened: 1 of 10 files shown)


src/CaptainHook.EventHandlerActor/Handlers/HttpClientExtensions.cs
```diff
@@ -66,21 +66,21 @@ public static class HttpClientExtensions
         private static async Task<HttpResponseMessage> RetryRequest(
             Func<Task<HttpResponseMessage>> makeTheCall)
         {
+
```
mszymeshop: Are those extra blank lines for a reason? Personally prefer when they are not there, your call though.

src/CaptainHook.EventHandlerActor/Handlers/HttpResponseMessageExtensions.cs
```diff
@@ -0,0 +1,15 @@
...
+namespace CaptainHook.EventHandlerActor.Handlers
+{
+    public static class HttpResponseMessageExtensions
+    {
+        /// <summary>
+        /// define "delivery failure" for the EDA context
+        /// </summary>
+        /// <param name="response">response object</param>
+        /// <returns>true if response indicates EDA delivery failure</returns>
+        public static bool IsDeliveryFailure(this HttpResponseMessage response) => response !=null && ((response.StatusCode >= HttpStatusCode.InternalServerError && response.StatusCode <= HttpStatusCode.NetworkAuthenticationRequired) || response.StatusCode == HttpStatusCode.TooManyRequests);
```
dguerin: The only one that does not sit well with me is accepting that 429s are failed delivery. For now in the context of this PR it makes sense. However, longer term it may make sent create a feature here such that if we are getting 429s from an endpoint, it means they are playing ball but can't answer the request just yet. 

As such if the retry after header is present we may choose to delay failing the message until a later time, keeping the message alive with lock renewals for a longer period than the one minute default lock duration. This may also link into the message consumption from service bus such that we slow down the consumption of messages for an endpoint when this occurs. 

Thoughts?

VaclavK: we discussed this and we need to track this - I am on board 

AHA feature?

dguerin: cool. Yeah I'll add something in for it.

PR #46, opened 2020-02-26
55965 http method load from kv

VaclavK:


src/CaptainHook.Common/Configuration/EventHandlerConfig.cs
```diff
@@ -36,6 +36,19 @@ public WebhookConfig()
...
+        /// <summary>
+        /// due to limitations of binder (simple POCOs), route the config through this property
+        /// </summary>
+        public string HttpVerb
+        {
+            get => HttpMethod.Method;
+            set
+            {
+                if (!string.IsNullOrWhiteSpace(value))
+                    HttpMethod = new HttpMethod(value);
```
dguerin: annoying but necessary

dguerin: This resolves the issues with the config builder then?

VaclavK: indeed it does... the in memory configs see the verb as flowing in through KV with this change

PR #47, opened 2020-03-06
#57443 - introduce CosmosDB communication project

mszymeshop:
### What
Introduction of a project, which is going to be moved to a separate package later, that is going to take care of CosmosDB communication for us.

### Why
Captain Hook is going to start relying on CosmosDB shortly and we need this capability.

### Remarks
The usage of this code is going to come in the following PRs.

### DIFF
--- a/src/CaptainHook.Database/CaptainHook.Database.csproj
+++ b/src/CaptainHook.Database/CaptainHook.Database.csproj
@@ -1,0 +1,16 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp2.2</TargetFramework>
+    <DebugType>Full</DebugType>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Autofac" Version="4.9.2" />
+    <PackageReference Include="Eshopworld.Telemetry" Version="2.1.5" />
+    <PackageReference Include="Microsoft.Azure.DocumentDB.Core" Version="2.6.0" />
+    <PackageReference Include="Microsoft.Extensions.Configuration.Binder" Version="2.2.4" />
+    <PackageReference Include="Microsoft.Extensions.Options" Version="2.2.0" />
+  </ItemGroup>
+
+</Project>
--- a/src/CaptainHook.Database/CaptainHook.Database.csproj
+++ b/src/CaptainHook.Database/CaptainHook.Database.csproj
@@ -6,9 +6,9 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Autofac" Version="4.9.2" />
-    <PackageReference Include="Eshopworld.Telemetry" Version="2.1.5" />
-    <PackageReference Include="Microsoft.Azure.DocumentDB.Core" Version="2.6.0" />
+    <PackageReference Include="Autofac" Version="4.9.4" />
+    <PackageReference Include="Eshopworld.Telemetry" Version="2.3.0-telemetry-beta4" />
+    <PackageReference Include="Microsoft.Azure.DocumentDB.Core" Version="2.10.1" />
     <PackageReference Include="Microsoft.Extensions.Configuration.Binder" Version="2.2.4" />
     <PackageReference Include="Microsoft.Extensions.Options" Version="2.2.0" />
   </ItemGroup>
--- a/src/CaptainHook.Database/CosmosDB/CosmosDbRepository.cs
+++ b/src/CaptainHook.Database/CosmosDB/CosmosDbRepository.cs
@@ -1,0 +1,202 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Net;
+using System.Threading.Tasks;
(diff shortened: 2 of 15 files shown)


src/CaptainHook.Database/CaptainHook.Database.csproj
```diff
@@ -0,0 +1,16 @@
...
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp2.2</TargetFramework>
+    <DebugType>Full</DebugType>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Autofac" Version="4.9.2" />
+    <PackageReference Include="Eshopworld.Telemetry" Version="2.1.5" />
+    <PackageReference Include="Microsoft.Azure.DocumentDB.Core" Version="2.6.0" />
```
VaclavK: this is quite old.. can we refresh this?

VaclavK: and whatever else?

src/CaptainHook.Database/CosmosDB/CosmosDbRepository.cs
```diff
@@ -0,0 +1,202 @@
...
+
+                    return true;
+                }
+                catch (DocumentClientException ex)
+                {
+                    if (ex.StatusCode == HttpStatusCode.NotFound && IsErrorMessageForDocumentResource(ex.Error.Message))
+                        return false;
+
+                    throw ;
+                }
```
tabman83: can use the `catch when` syntax there?

src/CaptainHook.Database/CosmosDB/ICosmosDbRepository.cs
```diff
@@ -0,0 +1,100 @@
...
+    /// </summary>
+    public interface ICosmosDbRepository
+    {
+        /// <summary>
+        /// Defines instance of a client for cosmos db
+        /// </summary>
+        IDocumentClient DbClient { get; }
+
+        /// <summary>
+        /// Defines the collection (and from which databsae) that should be used by the repository.
```
VaclavK: typo

src/CaptainHook.Database/CosmosDB/CosmosDbRepository.cs
```diff
@@ -0,0 +1,202 @@
...
+        private async Task<TResult> ExecuteFunction<TResult>(Func<Task<TResult>> func)
+        {
+            while (true)
+                try
+                {
+                    return await func();
+                }
+                catch (DocumentClientException ex) when (ex.StatusCode == HttpStatusCode.PreconditionFailed)
+                {
+                    throw new StaleDataException();
```
VaclavK: I do not particularly like this pattern of signalling by exception.. if you are returning some result object.. could we simply attach an error flag etc. and the reason?

src/CaptainHook.Database/CaptainHook.Database.csproj
mszymeshop: will do

src/CaptainHook.Database/CosmosDB/ICosmosDbRepository.cs
mszymeshop: Fixed

src/CaptainHook.Database/CaptainHook.Database.csproj
mszymeshop: Dependencies updated, however I kept them consolidated with the other dependencies in the solution. We can do another swipe when we migrate to .NET Core 3.1

src/CaptainHook.Database/CosmosDB/ICosmosDbRepository.cs
```diff
@@ -0,0 +1,100 @@
...
+        /// <param name="data">New version of the document data</param>
+        /// <returns></returns>
+        public static Task<Document> ReplaceAsync<T>(this ICosmosDbRepository repo, string id, T data)
+        {
+            return repo.ReplaceAsync (id, data, null);
+        }
+    }
+
+    /// <summary>
+    /// Defines basic ineterface for cosmos db-based repository
```
VaclavK: Found one more typo

src/CaptainHook.Database/CosmosDB/CosmosDbRepository.cs
mszymeshop: I have now added an enabler to address this: ADO #57823

src/CaptainHook.Database/CosmosDB/ICosmosDbRepository.cs
mszymeshop: Fixed as well, but well spot :100:

src/CaptainHook.Database/CosmosDB/CosmosDbRepository.cs
mszymeshop: @tabman83 I will refrain from changing this now, however this is a valid remark. We are going to rebuild exception handling as part of ADO #57823 (see @VaclavK comment), so this is likely to be addressed then.

PR #48, opened 2020-03-06
add event model definitions

VaclavK:


src/CaptainHook.Common/Models/V2/Event.cs
```diff
@@ -0,0 +1,27 @@
...
+        /// </summary>
+        public string Name { get; set; }
+        /// <summary>
+        /// underlying event (or message) full type name
+        /// </summary>
+        public string Type { get; set; }
+        /// <summary>
+        /// scope of visibility for the event instance
+        /// </summary>
+        public EventVisibilityEnum Visibility { get; set; } = EventVisibilityEnum.Internal;
```
tabman83: not a big fan of having the enum suffix for enums
also ref: https://rules.sonarsource.com/csharp/tag/convention/RSPEC-2344

VaclavK: cool... some teams use them.. some don't.. renamed

PR #49, opened 2020-03-08
#57444 - configure connection with CosmosDB

mszymeshop:
### What
Configure CosmosDB via Environment variables.

### Why
Temporarily we want to use Env variables to configure connection string to CosmosDB since the local emulator will be used. Long term, an Azure instance will be used and configured through KeyVaults. The run of the code is disabled until a feature flag is introduced.

### DIFF
--- a/src/CaptainHook.Fabric/ApplicationParameters/Local.1Node.xml
+++ b/src/CaptainHook.Fabric/ApplicationParameters/Local.1Node.xml
@@ -1,4 +1,4 @@
-﻿<?xml version="1.0" encoding="utf-8"?>
+<?xml version="1.0" encoding="utf-8"?>
 <Application xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" Name="fabric:/CaptainHook" xmlns="http://schemas.microsoft.com/2011/01/fabric">
   <Parameters>
     <Parameter Name="KeyVault_Base_Uri" Value="https://esw-tooling-ci-we.vault.azure.net/" />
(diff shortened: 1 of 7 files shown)


src/CaptainHook.Fabric/ApplicationParameters/Local.1Node.xml
```diff
@@ -9,5 +9,7 @@
     <Parameter Name="CaptainHook.DirectorService_DefaultMinReplicaSetSize" Value="1"/>
     <Parameter Name="CaptainHook.DirectorService_DefaultTargetReplicaSetSize" Value="1"/>
     <Parameter Name="CaptainHook.DirectorService_DefaultPartitionCount" Value="1"/>
+    <Parameter Name="CaptainHook.DirectorService_DefaultPartitionCount" Value="1"/>
+    <Parameter Name="EDA_CosmosDb_ConnectionString" Value=""/>
```
mszymeshop: This is where you specify your local emulator connection string.

VaclavK: Why the different naming structure? This will fall under Director service so can we use the same prefix?

mszymeshop: For some reason the value is not loaded when I go for something like `CaptainHook.DirectorService_CosmosDbConnectionString`. It does load when the name is as above. Didn't analyse why, maybe the `.` in the name? @VaclavK have you seen this situation before?

Anyway, I think it is only temporal until we work out the KV for this setting anyway. Mainly for local development.

VaclavK: Let’s do screen share on this today

VaclavK: Further to the point, there config would normally go to KV

mszymeshop: Ok, this is now changes so that KV is used and we can still rely on Env for local config.

PR #52, opened 2020-03-24
introduce test fixture and sample test to spike the flow

VaclavK:


VaclavK:
@tabman83 @mszymeshop please note that the int tests (l1) will continue to fail till we deploy new version of peter pan

src/Tests/CaptainHook.Tests/Web/FlowTests/E2EFlowTestsFixture.cs
```diff
@@ -0,0 +1,116 @@
...
+#else
+            //for local testing to bypass KV load
+
+            var instrKey = "";
+            var sbConnString = "";
+            var subId = "";
+            peterPanUrlBase = "";
+#endif
+
+            _bb = new BigBrother(instrKey, instrKey);
```
tabman83: Aren't we going to use the new Telemetry package & BigBrother constructor?

src/Tests/CaptainHook.Tests/Web/FlowTests/ProcessedEventModel.cs
```diff
@@ -0,0 +1,28 @@
+﻿using System;
+using Newtonsoft.Json;
+
+namespace CaptainHook.Tests.Web.FlowTests
+{
+    public class ProcessedEventModel
+    {
+        [JsonProperty( "eventType")]
```
tabman83: Same question as for peter pan, we do we use per-property Json attribute instead of global one.

src/Tests/CaptainHook.Tests/Web/FlowTests/E2EFlowTestsFixture.cs
VaclavK: there is IMHO no value of telemetry package pre-configured BB usage here since none of the initializers apply here really

src/Tests/CaptainHook.Tests/Web/FlowTests/ProcessedEventModel.cs
VaclavK: can you elaborate?

src/Tests/CaptainHook.Tests/Web/FlowTests/E2EFlowTestsFixture.cs
tabman83: Thanks that was just for me to understand.

src/Tests/CaptainHook.Tests/Web/FlowTests/ProcessedEventModel.cs
tabman83: JsonSerializer.Deserialize takes an option where you can specify PropertyNameCaseInsensitive = true even if I think it will take this by default. Anyway this will avoid you specifying a JsonProperty for each model property.

VaclavK: I get you now.. I kinda like the option of controlling naming (not just casing) per field but I understand that may be personal preference

PR #54, opened 2020-04-06
new test flows - routed matched and unmatched flow- and associated ch…

VaclavK:
…anges


VaclavK:
> 
> 
> Looks nice to me, just wondering why we have some hardcoded stuff like scope and tenant code and whether they can be extracted someway or at least defined as constants.

these are tests so that makes them self encapsulating and isolates them... there will be no real change in values either

PR #56, opened 2020-04-08
#61975 - migrate Captain Hook to .NET Core 3.1

mszymeshop:
### What
Captain Hook is migrated to .NET Core 3.1 on top of that missing Service Fabric Telemetry is moved to Captain Hook solution (along with unit tests) not to lose any features (we might consider moving that to a separate package later on).

### Why
We are migrating our applications to .NET Core 3.1.


mszymeshop:
@tabman83 I have now added the `global.json` file

PR #57, opened 2020-04-08
#62771 - fix failing integration tests due to invalid Messaging reference

mszymeshop:
### What
Reference EShopworld.Messaging package directly as it should have been from the beginning. Right now the old version of 1.1.1 is referenced causing issues in runtime for L1 tests.

### DIFF
--- a/src/Tests/CaptainHook.Tests/CaptainHook.Tests.csproj
+++ b/src/Tests/CaptainHook.Tests/CaptainHook.Tests.csproj
@@ -17,6 +17,7 @@
   </PropertyGroup>
 
   <ItemGroup>
+    <PackageReference Include="Eshopworld.Messaging" Version="2.1.2" />
     <PackageReference Include="Eshopworld.Telemetry" Version="3.1.2" />
     <PackageReference Include="Eshopworld.Tests.Core" Version="2.0.3" />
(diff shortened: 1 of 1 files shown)


VaclavK:
@mszymeshop this is tied to.net 31 upgrade I assume?

mszymeshop:
> @mszymeshop this is tied to.net 31 upgrade I assume?

Only partially as I think the upgrade might have surfaced another problem. I had to upgrade our packages and Messenger failed to create an instance due to the use of methods from EShopWorld.Core which did not exist. Check this out for more info: https://dev.azure.com/eshopworld/Github%20build/_releaseProgress?_a=release-environment-logs&releaseId=1361&environmentId=4086

VaclavK:
why would you have explicit dependency if the test project includes other solution projects that actually consume this.. so in that sense you avoid versioning problems

VaclavK:
> 
> 
> why would you have explicit dependency if the test project includes other solution projects that actually consume this.. so in that sense you avoid versioning problems

never mind, we covered that on the call

VaclavK:
@mszymeshop the messaging comes to tests via platform events

image

mszymeshop:
Thank you @VaclavK, didn't notice it as it does not start with EShopWorld. Thanks! Still, the change in the PR is valid then.

PR #59, opened 2020-04-09
.NET Core SDK version downgraded to 3.1.101

dawidkeshop:


tabman83:
why the downgrade?

dawidkeshop:
> why the downgrade?

It was just an idea to test, but Oisin shows us a proper approach.

dawidkeshop:
Closed as non-relevant.

PR #62, opened 2020-04-15
improves code, more test diagnostics

VaclavK:
unblock test parallelization as original code was not running under xunit scheduling and was hanging on system with less cores


VaclavK:
I can walk you through why this is done

PR #63, opened 2020-04-28
#65429 - retry service bus setup when connection is dropped

tabman83:
On EventReaderService.cs the catch block for the ServiceBusCommunicationException has been augmented with a retry policy on the Service Bus setup method to allow for a recovery when the connection drop lasts longer. The strategy is an exponential backoff.

A custom AI event is logged (ServiceBusReconnectionAttemptEvent) carrying the retry count so that appropriate alerting can be later set up.

### DIFF
--- a/src/CaptainHook.EventReaderService/EventReaderService.cs
+++ b/src/CaptainHook.EventReaderService/EventReaderService.cs
@@ -64,7 +64,12 @@ namespace CaptainHook.EventReaderService
         internal ConcurrentDictionary<Guid, MessageReceiverWrapper> _messageReceivers = new ConcurrentDictionary<Guid, MessageReceiverWrapper>();
         internal MessageReceiverWrapper _activeMessageReader;
 
-        private Func<int, TimeSpan> exponentialBackoff = x => TimeSpan.FromSeconds(Math.Pow(2, x));
+        private static TimeSpan retryCeiling = TimeSpan.FromSeconds(60);
+        private readonly Func<int, TimeSpan> exponentialBackoff = x =>
+        {
+            var value = TimeSpan.FromSeconds(Math.Pow(2, x));
+            return retryCeiling > value ? retryCeiling : value;
+        };
 
         /// <summary>
         /// Default ctor used at runtime
@@ -274,7 +279,6 @@ namespace CaptainHook.EventReaderService
             }
             catch (Exception e)
             {
-                // HttpRequestException
                 BigBrother.Write(e.ToExceptionEvent());
             }
         }
--- a/src/CaptainHook.EventReaderService/EventReaderService.cs
+++ b/src/CaptainHook.EventReaderService/EventReaderService.cs
@@ -253,22 +253,25 @@ namespace CaptainHook.EventReaderService
                     }
                     catch (ServiceBusCommunicationException sbCommunicationException)
                     {
-                        BigBrother.Write(sbCommunicationException.ToExceptionEvent());
+                        BigBrother.Write(sbCommunicationException);
 
-                        await Policy.Handle<HttpRequestException>()
+                        await Policy.Handle<Exception>()
                             .WaitAndRetryForeverAsync(exponentialBackoff,
                                 (exception, retryCount, timeSpan) =>
                                 {
+                                    BigBrother.Write(exception);
+
                                     _bigBrother.Publish(new ServiceBusReconnectionAttemptEvent
                                     {
-                                        RetryCount = retryCount
+                                        RetryCount = retryCount,
+                                        FabricId = $"{this.Context.ServiceName}:{this.Context.ReplicaId}"
                                     });
                                 }
                             ).ExecuteAsync(SetupServiceBus);
                     }
-                    catch (Exception e)
+                    catch (Exception exception)
                     {
-                        BigBrother.Write(e.ToExceptionEvent());
+                        BigBrother.Write(exception);
                     }
                 }
 
--- a/src/CaptainHook.EventReaderService/EventReaderService.cs
+++ b/src/CaptainHook.EventReaderService/EventReaderService.cs
@@ -6,6 +6,7 @@ using System.Fabric;
 using System.IO;
 using System.Linq;
 using System.Net;
+using System.Net.Http;
 using System.Reflection;
 using System.Text;
 using System.Threading;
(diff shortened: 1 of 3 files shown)


src/CaptainHook.EventReaderService/EventReaderService.cs
```diff
@@ -259,10 +271,10 @@ protected override async Task RunAsync(CancellationToken cancellationToken)
                 }
 
                 _bigBrother.Publish(new Common.Telemetry.CancellationRequestedEvent { FabricId = $"{Context.ServiceName}:{Context.ReplicaId}" });
-             
             }
             catch (Exception e)
             {
+                // HttpRequestException
```
VaclavK, line 277: ??

tabman83: just a note I forgot to remove

src/CaptainHook.EventReaderService/EventReaderService.cs
```diff
@@ -62,7 +64,7 @@ public class EventReaderService : StatefulService, IEventReaderService
         internal ConcurrentDictionary<Guid, MessageReceiverWrapper> _messageReceivers = new ConcurrentDictionary<Guid, MessageReceiverWrapper>();
         internal MessageReceiverWrapper _activeMessageReader;
 
-
+        private Func<int, TimeSpan> exponentialBackoff = x => TimeSpan.FromSeconds(Math.Pow(2, x));
```
VaclavK, line 67: should we not have a ceiling too?

src/CaptainHook.EventReaderService/EventReaderService.cs
```diff
@@ -250,7 +252,17 @@ protected override async Task RunAsync(CancellationToken cancellationToken)
                     catch (ServiceBusCommunicationException sbCommunicationException)
                     {
                         BigBrother.Write(sbCommunicationException.ToExceptionEvent());
-                        await SetupServiceBus();
+
+                        await Policy.Handle<HttpRequestException>()
+                            .WaitAndRetryForeverAsync(exponentialBackoff,
+                                (exception, retryCount, timeSpan) =>
```
VaclavK, line 258: you could probably use just the   (exception, timespan) => signature here as the context (3rd one as per doco) is not use anyway

tabman83: I need the retry count because I want to report it through AI. That is so we can setup alerting after n retries.

src/CaptainHook.EventReaderService/EventReaderService.cs
tabman83: I have added the ceiling to the exponential backoff function.

src/CaptainHook.EventReaderService/EventReaderService.cs
tabman83: Removed.

src/CaptainHook.EventReaderService/EventReaderService.cs
VaclavK: I misread this, makes sense

src/CaptainHook.EventReaderService/EventReaderService.cs
```diff
@@ -62,7 +64,12 @@ public class EventReaderService : StatefulService, IEventReaderService
         internal ConcurrentDictionary<Guid, MessageReceiverWrapper> _messageReceivers = new ConcurrentDictionary<Guid, MessageReceiverWrapper>();
         internal MessageReceiverWrapper _activeMessageReader;
 
-
+        private static TimeSpan retryCeiling = TimeSpan.FromSeconds(60);
+        private readonly Func<int, TimeSpan> exponentialBackoff = x =>
+        {
+            var value = TimeSpan.FromSeconds(Math.Pow(2, x));
+            return retryCeiling > value ? retryCeiling : value;
```
VaclavK, line 71: this could be simplified by using https://docs.microsoft.com/en-us/dotnet/api/system.math.clamp?view=netcore-3.1

tabman83: It was a quick a nice exercise, thanks & done.

src/CaptainHook.EventReaderService/EventReaderService.cs
```diff
@@ -250,7 +257,17 @@ protected override async Task RunAsync(CancellationToken cancellationToken)
                     catch (ServiceBusCommunicationException sbCommunicationException)
                     {
                         BigBrother.Write(sbCommunicationException.ToExceptionEvent());
-                        await SetupServiceBus();
+
+                        await Policy.Handle<HttpRequestException>()
```
VaclavK, line 258: I would suggest handling perhaps Exception  itself (is that exception truly the only one thrown out of setupservicebus under all conditions)?

for connection attempt, I think we should we also push the exception itself into the AI (as it is available)

src/CaptainHook.EventReaderService/EventReaderService.cs
```diff
@@ -250,7 +254,17 @@ protected override async Task RunAsync(CancellationToken cancellationToken)
...
+                            .WaitAndRetryForeverAsync(exponentialBackoff,
+                                (exception, retryCount, timeSpan) =>
+                                {
+                                    _bigBrother.Publish(new ServiceBusReconnectionAttemptEvent
```
dguerin, line 262: Could we also publish the name of the reader? this will give us the name of the subscription and topic which is having the difficulty

tabman83: It's been added.

src/CaptainHook.EventReaderService/EventReaderService.cs
```diff
@@ -253,22 +253,25 @@ protected override async Task RunAsync(CancellationToken cancellationToken)
...
                             .WaitAndRetryForeverAsync(exponentialBackoff,
                                 (exception, retryCount, timeSpan) =>
                                 {
+                                    BigBrother.Write(exception);
+
                                     _bigBrother.Publish(new ServiceBusReconnectionAttemptEvent
                                     {
-                                        RetryCount = retryCount
+                                        RetryCount = retryCount,
+                                        FabricId = $"{this.Context.ServiceName}:{this.Context.ReplicaId}"
```
dguerin, line 267: What is the actual value of this.Context.ServiceName? 

I was thinking more in terms of having a property or two on this event so that we can query for it in AI for example. Properties for: 
_initData.EventType, 
_initData.SubscriptionName

That may be all we need right now. If we go to multiple consumers later for higher through put we would need the instance id but that's something for later.

src/CaptainHook.EventReaderService/EventReaderService.cs
tabman83: This has been implemented as well.

src/CaptainHook.EventReaderService/EventReaderService.cs
tabman83: I see. It's a quick change thanks.

PR #64, opened 2020-04-30
Abstract configuration loading

tabman83:


src/CaptainHook.Common/Configuration/Configuration.cs
```diff
@@ -0,0 +1,96 @@
...
+
+        private Configuration()
+        {
+        }
+
+        /// <summary>
+        /// Load configuration settings and domain events 
+        /// </summary>
+        /// <returns>An instance holding the configuration settings and domain events</returns>
+        public static Configuration Load()
```
chomick, line 36: Any reason this is static?

src/CaptainHook.Common/Configuration/Configuration.cs
```diff
@@ -0,0 +1,96 @@
...
+using Microsoft.Azure.Services.AppAuthentication;
+using Microsoft.Extensions.Configuration;
+using Microsoft.Extensions.Configuration.AzureKeyVault;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+
+namespace CaptainHook.Common.Configuration
+{
+    public sealed class Configuration
```
dawidkeshop, line 11: Wouldn't be better to separate configuration data and loading logic instead of placing it in one class?
It will give us more flexibility with testing (e.g. we could use test configuration object without necessity to mock lots of stuff) as well as allow to keep things short and clear. Now once we add CosmosDb integration, this class will contain more and more dependencies.

dguerin, line 11: does this need to be sealed?

tabman83: I don't think whoever is using the configuration class should be able to extend it.

tabman83: The aim for this task is to remove the dependency from the keyvault from every single service/actor and this step accomplishes it. We are going to have more visibility on the goal in the next tasks and that will give us the opportunity to improve this if need be. Cosmos Db may not need to be part of this class as it will come from outside.

dawidkeshop: OK, sure.

PR #66, opened 2020-05-04
62065 powershell generation

dawidkeshop:


chomick:
Probably no point changing now but originally I thought we will simply load the Json file and, using JObject instances, output (recursively) the content into PS format - you would process everything with a generic code, rather than handling each section individually.

PR #72, opened 2020-05-14
#67466 - add web api to Captain Hook

mszymeshop:
### What
Changes which add Web Api to Captain Hook and host it within Service Fabric. The Api supports /probe endpoint only at this stage.

### Why
The Web API is going to be used to control Captain Hook configuration in the long term.

### Proof of life
image

### DIFF
--- a/src/CaptainHook.Api/Startup.cs
+++ b/src/CaptainHook.Api/Startup.cs
@@ -14,6 +14,7 @@ using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
 using System.Reflection;
 using Eshopworld.Telemetry.Configuration;
+using Eshopworld.Telemetry.Processors;
 using Microsoft.AspNetCore.Mvc.Filters;
 using Microsoft.OpenApi.Models;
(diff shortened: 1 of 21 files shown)


src/CaptainHook.Api/Startup.cs
```diff
@@ -0,0 +1,170 @@
...
+        /// ConfigureServices is where you register dependencies. This gets
+        /// called by the runtime before the ConfigureContainer method, below.
+        /// </summary>
+        /// <remarks>See https://docs.autofac.org/en/latest/integration/aspnetcore.html#asp-net-core-3-0-and-generic-hosting</remarks>
+        /// <param name="builder">The <see cref="ContainerBuilder"/> to configure</param>
+        public void ConfigureContainer(ContainerBuilder builder)
+        {
+            builder.RegisterInstance(_telemetrySettings).SingleInstance();
+            builder.RegisterModule<TelemetryModule>();
+        }
```
tabman83, line 54: Can we add the SuccessfulProbeFilter here?

mszymeshop: Now added

PR #75, opened 2020-05-18
60287 pass subscriber configuration from Reader to Handler

dawidkeshop:


src/CaptainHook.DirectorService/DirectorService.cs
```diff
@@ -81,29 +82,7 @@ protected override async Task RunAsync(CancellationToken cancellationToken)
...
-                                PartitionSchemeDescription = new SingletonPartitionSchemeDescription(),
-                                ServiceTypeName = ServiceNaming.EventReaderServiceType,
-                                ServiceName = new Uri(readerServiceNameUri),
-                                InitializationData = initializationData,
-                                PlacementConstraints = _defaultServiceSettings.DefaultPlacementConstraints
-                            },
-                            TimeSpan.FromSeconds(30),
-                            cancellationToken);
-                    }
+                    await CreateServiceForSubscriber(subscriber, serviceList, cancellationToken);
```
tabman83, line 85: This has been discussed on a separate project - as per MS conventions we should consider adding the -Async suffix to any async method to improve clarity.

dawidkeshop: Sure

src/CaptainHook.Common/MessageData.cs
```diff
@@ -51,5 +53,14 @@ public MessageData(string payload, string type, string subscriberName, string re
...
 
+        /// <summary>
+        /// The subscriber configuration related to the message.
+        /// </summary>
+        public SubscriberConfiguration SubscriberConfig { get; set; }
+
+        /// <summary>
+        /// The Webhook configuration related to the message.
+        /// </summary>
+        public WebhookConfig WebhookConfig { get; set; }
```
mszymeshop, line 64: We can probably get rid of that and start relying on `SubscriberConfig` entirely. To be added as a Tech Debt as discussed offline.

src/CaptainHook.Common/ServiceModels/EventReaderInitData.cs
```diff
@@ -41,6 +45,19 @@ public static EventReaderInitData FromSubscriberConfiguration(SubscriberConfigur
             };
         }
 
+        public static EventReaderInitData FromSubscriberConfiguration(SubscriberConfiguration subscriberConfiguration, WebhookConfig webhookConfig)
```
mszymeshop, line 48: The above method is probably no longer needed or both can be merged together.

src/CaptainHook.EventHandlerActor/Handlers/EventHandlerFactory.cs
```diff
@@ -57,10 +51,10 @@ public IHandler CreateEventHandler(string eventType, string subscriberName)
                     _authenticationHandlerFactory,
                     _requestLogger,
                     _bigBrother,
-                    subscriberConfig);
+                    messageData.SubscriberConfig);
             }
 
-            return CreateWebhookHandler(subscriberConfig.Name);
+            return CreateWebhookHandler(messageData.WebhookConfig);
```
mszymeshop, line 57: Potentially you could just pass the WebhookConfig from the subscriber here. If that is out of scope let's include it in the tech debt which was discussed.

src/CaptainHook.Common/MessageData.cs
chomick: Yes, we will have to refactor the models over time but would minimise changes for now.

src/CaptainHook.DirectorService/DirectorService.cs
```diff
@@ -144,6 +124,67 @@ protected override async Task RunAsync(CancellationToken cancellationToken)
...
+                   TimeSpan.FromSeconds(30),
+                   cancellationToken);
+            }
+        }
+
+        private static (string newName, string oldName) FindServiceNames(ICollection<string> serviceList, string readerServiceNameUri)
+        {
+            string newName = $"{readerServiceNameUri}-a", oldName = null;
+
+            if (serviceList.Contains(readerServiceNameUri))
```
mszymeshop, line 163: I would consider simplifying the algorithm. My suggestion: find the first occurrence of `uri, uri-b, uri-a` in `serviceList` and use it as `oldName`. If `oldName` matches `newName` use `uri-b` instead as newName. What do you think @dawidkeshop ?

dawidkeshop: Yes, you're right. Especially that there is possibility that more than one service should be deleted.

mszymeshop:
After the compatibility commit, it looks good ✔️

PR #79, opened 2020-05-23
#67489 implement API controller for config refresh

tabman83:


src/CaptainHook.Api/Helpers/FakeAuthBuilderExtensions.cs
```diff
@@ -0,0 +1,23 @@
+﻿using System;
+using Microsoft.AspNetCore.Builder;
+
+namespace CaptainHook.Api.Helpers
+{
+    /// <summary>
+    /// Extension methods to add fake authentication capabilities to an HTTP application pipeline.
+    /// </summary>
+    public static class FakeAuthBuilderExtensions
```
mszymeshop, line 9: What's the need for that? Was there a problem with STS Client Secret or the Certificate?

tabman83: I took the idea from BF where we use conditional compilation to disable auth when running locally, thus making debugging quicker.

src/CaptainHook.Api.Client/ApiClient/CaptainHookClient.cs
```diff
@@ -412,5 +413,120 @@ public async Task<HttpOperationResponse> GetProbeWithHttpMessagesAsync(Dictionar
...
+        /// <param name='cancellationToken'>
+        /// The cancellation token.
+        /// </param>
+        /// <exception cref="HttpOperationException">
+        /// Thrown when the operation returned an invalid status code
+        /// </exception>
+        /// <return>
+        /// A response object containing the response body and response headers.
+        /// </return>
+        public async Task<HttpOperationResponse> RefreshConfigForEventWithHttpMessagesAsync(CaptainHookApiModelsRefreshConfigRequest body = default(CaptainHookApiModelsRefreshConfigRequest), Dictionary<string, List<string>> customHeaders = null, CancellationToken cancellationToken = default(CancellationToken))
```
dawidkeshop, line 434: I think this method is quite long. Maybe would be valuable to split it into pieces instead of adding comments describing each section?

tabman83: This is the client generated by auto rest so I guess we don't want to mess with it in case we're going to re-generate it.

dawidkeshop: Sure, you right. I didn't noticed that this auto-generated in GitHub. Sorry for confusion.

PR #80, opened 2020-05-25
#67491 implement integration tests to validate config refresh endpoint

tabman83:
Implement integration tests to validate the refresh configuration endpoint.

### DIFF
--- a/src/CaptainHook.Api.Client/scripts/GenerateClient.ps1
+++ b/src/CaptainHook.Api.Client/scripts/GenerateClient.ps1
@@ -76,7 +76,7 @@ function Install-Tooling
 	param([string] $ToolsPath)
 
 	Write-Output "Installing or updating Swashbuckle CLI"
-	dotnet tool install swashbuckle.aspnetcore.cli --version 5.4.1 --tool-path $ToolsPath
+	dotnet tool update swashbuckle.aspnetcore.cli --version 5.4.1 --tool-path $ToolsPath
 
 	Write-Output "Installing or updating AutoRest"
 	npm install autorest@3.0.6187 --no-audit --prefix $ToolsPath
--- a/src/CaptainHook.Api.Tests/appsettings.CI.json
+++ b/src/CaptainHook.Api.Tests/appsettings.CI.json
@@ -1,0 +1,6 @@
+{
+  "StsApiEndpoint": "https://security-sts-api.ci.eshopworld.net",
+  "RefreshingTokenProviderOptions": {
+    "Issuer": "https://security-sts.ci.eshopworld.net"
+  }
+}
--- a/src/CaptainHook.Api.Tests/appsettings.PROD.json
+++ b/src/CaptainHook.Api.Tests/appsettings.PROD.json
@@ -1,6 +1,0 @@
-{
-  "StsApiEndpoint": "https://security-sts-api.production.eshopworld.com",
-  "RefreshingTokenProviderOptions": {
-    "Issuer": "https://security-sts.production.eshopworld.com"
-  }
-}
--- a/src/CaptainHook.Api.Tests/appsettings.PROD.json
+++ b/src/CaptainHook.Api.Tests/appsettings.PROD.json
@@ -1,0 +1,6 @@
+{
+  "StsApiEndpoint": "https://security-sts-api.production.eshopworld.com",
+  "RefreshingTokenProviderOptions": {
+    "Issuer": "https://security-sts.production.eshopworld.com"
+  }
+}
--- a/src/CaptainHook.Api.Tests/appsettings.json
+++ b/src/CaptainHook.Api.Tests/appsettings.json
@@ -1,0 +1,8 @@
+{
+  "RefreshingTokenProviderOptions": {
+    "ClientId": "tooling.eda.api.client",
+    "Scopes": [
+      "tooling.eda.api.all"
+    ]
+  }
+}
--- a/src/CaptainHook.Api.Tests/config/ApiClientFixture.cs
+++ b/src/CaptainHook.Api.Tests/config/ApiClientFixture.cs
@@ -14,17 +14,6 @@ namespace CaptainHook.Api.Tests.Config
             return new CaptainHookClient(CaptainHookTestUri, AnonymousCredential.Instance);
         }
 
-        //private TokenCredentials CreateCredentials()
-        //{
-        //    return new TokenCredentialsBuilder()
-        //        // Loaded by default from appsettings.json
-        //        //.AddAudience("mydomain.api")
-        //        .AddClientId("tooling.eda.api.client")
-        //        .AddScopes("tooling.eda.api.all")
-        //        // Build
-        //        .Build();
-        //}
-
         public ICaptainHookClient GetApiClient()
         {
             var token = new TokenCredentialsBuilder().Build();
--- a/src/CaptainHook.Api.Tests/config/ApiClientFixture.cs
+++ b/src/CaptainHook.Api.Tests/config/ApiClientFixture.cs
@@ -1,16 +1,34 @@
 ﻿using System;
 using CaptainHook.Api.Client;
+using EShopworld.Security.Services.Testing.Token;
 using Microsoft.Rest;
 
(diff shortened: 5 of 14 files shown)


src/CaptainHook.Api.Tests/config/ApiClientFixture.cs
```diff
@@ -1,16 +1,34 @@
...
+        private static Uri CaptainHookTestUri = new Uri("https://localhost:24010");
+
         public ICaptainHookClient GetApiUnauthenticatedClient()
         {
-            var url = new Uri("https://localhost:24010");
+            return new CaptainHookClient(CaptainHookTestUri, AnonymousCredential.Instance);
+        }
 
-            return new CaptainHookClient(url, AnonymousCredential.Instance);
+        //private TokenCredentials CreateCredentials()
```
dawidkeshop, line 17: I think it's better to remove commented code block.

tabman83: Thanks just forgot.

src/CaptainHook.Api.Client/scripts/GenerateClient.ps1
```diff
@@ -76,7 +76,7 @@ function Install-Tooling
 	param([string] $ToolsPath)
 
 	Write-Output "Installing or updating Swashbuckle CLI"
-	dotnet tool install swashbuckle.aspnetcore.cli --version 5.4.1 --tool-path $ToolsPath
+	dotnet tool update swashbuckle.aspnetcore.cli --version 5.4.1 --tool-path $ToolsPath
```
mszymeshop, line 79: :+1:

src/CaptainHook.Api.Tests/appsettings.CI.json
```diff
@@ -0,0 +1,6 @@
+{
+  "StsApiEndpoint": "https://security-sts-api.ci.eshopworld.net",
```
mszymeshop, line 2: I have doubts if we should be exposing these publically. My preference would be to use KeyVault to store these instead. @chomick @dguerin @tabman83 what do you think?

VaclavK: already exposed :D

src/CaptainHook.Api.Tests/appsettings.json
```diff
@@ -0,0 +1,8 @@
+{
+  "RefreshingTokenProviderOptions": {
+    "ClientId": "tooling.eda.api.client",
```
mszymeshop, line 3: Should we change the client id name? Kevin has mentioned this is too similar to the tooling.eda.client but I am not sure if there is a better name. What do you think?

VaclavK: I personally have no problem decyphering the difference and actually the similar nature makes sense in terms of tokenisation.... my 2 c

src/CaptainHook.Api.Tests/appsettings.CI.json
mszymeshop: @VaclavK haha, well yeah :) Still, what do you think here?

VaclavK: hiding the url like that does not IMHO add real security - it is essentially security by obscurity..

either we have a hole (and in that sense hiding the URL (already known I am sure even from elsewhere) is not going to help or we do not and we are fine...

tabman83: I don't think a URL is a leak as it's an already publicly available resource, but I am happy to take your guidance here.

src/CaptainHook.Api.Tests/appsettings.json
tabman83: I don't have a strong opinion here so I am happy to take your guidance here as well.

src/CaptainHook.Api.Tests/appsettings.CI.json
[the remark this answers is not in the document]
chomick: needed for token content

src/CaptainHook.Api.Tests/appsettings.PROD.json
```diff
@@ -0,0 +1,6 @@
+{
+  "StsApiEndpoint": "https://security-sts-api.production.eshopworld.com",
```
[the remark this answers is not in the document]
tabman83: I have removed settings file for both PROD and SAND environments.

src/CaptainHook.DirectorService/DirectorService.cs
```diff
@@ -146,6 +146,11 @@ protected override async Task RunAsync(CancellationToken cancellationToken)
 
         public Task ReloadConfigurationForEventAsync(string eventName)
         {
+            if(string.IsNullOrWhiteSpace(eventName))
```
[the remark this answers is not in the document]
tabman83: I have removed the event name validation for now.

src/CaptainHook.Api.Tests/config/ApiClientFixture.cs
```diff
@@ -1,16 +1,23 @@
 ﻿using System;
 using CaptainHook.Api.Client;
+using EShopworld.Security.Services.Testing.Token;
 using Microsoft.Rest;
 
 namespace CaptainHook.Api.Tests.Config
 {
     public class ApiClientFixture
     {
+        private static Uri CaptainHookTestUri = new Uri("https://localhost:24010");
```
chomick, line 10: Should that not be configured per env?

src/CaptainHook.Api/Controllers/RefreshConfigController.cs
```diff
@@ -31,6 +35,10 @@ public async Task<IActionResult> RefreshConfigForEvent([FromBody]RefreshConfigRe
 
                 return Ok();
             }
+            catch(Exception exception) when (exception.InnerException is ArgumentNullException)
```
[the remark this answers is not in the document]
tabman83: This has been removed following our conversation.

src/CaptainHook.Api/Controllers/RefreshConfigController.cs
```diff
@@ -31,6 +35,10 @@ public async Task<IActionResult> RefreshConfigForEvent([FromBody]RefreshConfigRe
...
+            {
+                return NotFound();
+            }
             catch
```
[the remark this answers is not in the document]
tabman83: The generic exception is now being logged.

src/CaptainHook.Api.Tests/config/ApiClientFixture.cs
tabman83: Test is marked as IsDev now as suggested by @mszymeshop

src/CaptainHook.Api/Controllers/RefreshConfigController.cs
```diff
@@ -31,8 +47,9 @@ public async Task<IActionResult> RefreshConfigForEvent([FromBody]RefreshConfigRe
 
                 return Ok();
             }
-            catch
+            catch(Exception exception)
```
chomick, line 50: Normally would not agree that this is the correct behaviour - if the exception happens that we do not understand, we should fail with 5xx or simply let exception to bubble up - no reason to catch this for the sake of logging and returning 400.
But to move on with this PR and as this is temp endpoint, I think we are ok. Just a note for future work.

src/CaptainHook.Api.Tests/config/ApiClientFixture.cs
chomick: ok, so we do not expect htem to run on CI? I think this is not correct, we should to ensure the deployment is healthy?
Again, happy to close it for now but we need a significant re-do of the testing in CH - it is very messy right now - we should follow the same standards as every other team/project follows, in relation to naming conventions, folder structures and test attributes.

src/CaptainHook.Api/Controllers/RefreshConfigController.cs
tabman83: Noted, thanks.

PR #84, opened 2020-05-29
69911 configuration comparison

dawidkeshop:


src/CaptainHook.DirectorService/SubscriberConfigurationComparer.cs
```diff
@@ -0,0 +1,46 @@
...
+            }
+
+            return new Result(added, removed, changed);
+        }
+
+        public class Result
+        {
+            public IDictionary<string, SubscriberConfiguration> Added { get; }
+            public IDictionary<string, SubscriberConfiguration> Removed { get; }
+            public IDictionary<string, SubscriberConfiguration> Changed { get; }
```
tabman83, line 35: I would make those IReadOnlyDictionary

dawidkeshop: Sure!

src/CaptainHook.DirectorService/SubscriberConfigurationComparer.cs
```diff
@@ -30,12 +30,13 @@ public Result Compare(IDictionary<string, SubscriberConfiguration> oldConfig, ID
...
+            public IReadOnlyDictionary<string, SubscriberConfiguration> Removed { get; }
+            public IReadOnlyDictionary<string, SubscriberConfiguration> Changed { get; }
             public bool HasChanged => this.Added.Any() || this.Removed.Any() || this.Changed.Any();
 
-            public Result(IDictionary<string, SubscriberConfiguration> added, IDictionary<string, SubscriberConfiguration> removed, IDictionary<string, SubscriberConfiguration> changed)
+            public Result(IReadOnlyDictionary<string, SubscriberConfiguration> added, 
+                IReadOnlyDictionary<string, SubscriberConfiguration> removed, IReadOnlyDictionary<string, SubscriberConfiguration> changed)
             {
                 Added = added;
                 Removed = removed;
```
tabman83: Don't mean to sound picky but I feel it makes more sense to leave the constructor with the Dictionary and then get the readonly version upon assignment:
`Added = added.ToReadOnlyDictionary()`
and so on.

dawidkeshop: Done using the default ReadOnlyDictionary constructor

src/CaptainHook.DirectorService/SubscriberConfigurationComparer.cs
```diff
@@ -0,0 +1,47 @@
+﻿using System.Collections.Generic;
+using System.Collections.ObjectModel;
+using System.Linq;
+using CaptainHook.Common.Configuration;
+using Newtonsoft.Json;
+
+namespace CaptainHook.DirectorService
+{
+    public class SubscriberConfigurationComparer
```
chomick, line 9: You will need some interface to unit test code that depends on this guy

src/CaptainHook.DirectorService/SubscriberConfigurationComparer.cs
```diff
@@ -0,0 +1,47 @@
...
+    {
+        public Result Compare(IDictionary<string, SubscriberConfiguration> oldConfig, IDictionary<string, SubscriberConfiguration> newConfig)
```
chomick, line 11: Add comments on what is the intended usage/what gets compared etc

src/Tests/CaptainHook.Tests/Director/SubscriberConfigurationComparerTests.cs
```diff
@@ -0,0 +1,156 @@
...
+using FluentAssertions;
+using Xunit;
+
+namespace CaptainHook.Tests.Director
+{
+    public class SubscriberConfigurationComparerTests
+    {
+        private readonly Dictionary<string, SubscriberConfiguration> oldConfigs = new Dictionary<string, SubscriberConfiguration>
+        {
+            ["event1-captain-hook"] = new SubscriberConfigurationBuilder().WithType("event1").WithSubscriberName("captain-hook").Create(),
```
chomick, line 15: I like the fact you guys added 'builders' concept

src/CaptainHook.DirectorService/SubscriberConfigurationComparer.cs
dawidkeshop: I'll improve during work on unit tests task.

src/CaptainHook.DirectorService/SubscriberConfigurationComparer.cs
chomick: What about webhook configiurations? I beleive they are not part of those dictionaries?

dawidkeshop: In fact they are. Both webhook and subscribers configurations are built from same data read from KV. The difference is that webhook store callbacks as separate items. In fact it's data duplication and we already have a tech debt task to remove webhooks.

Anyway, if something has been changed in KV, we only have to compare subscribers to detect the change.

chomick: ok, if that's the case, we are good then.
So when we load KV data, webhooks gets added to the subscriber list as a first subscriber?

dawidkeshop: Not exactly - both these collections are being created at the same time, just when subscriber has Callback then it's added separately to the webhooks list.
You can see it here: https://github.com/eShopWorld/captain-hook/blob/master/src/CaptainHook.Common/Configuration/Configuration.cs#L71

chomick: So what will happen if I have no Callback and update my single webhook's URL in the KV? How will you detect this change?

chomick: ok - discussed with Dawid that webhooks are added to the subscribers list on config load.

PR #85, opened 2020-05-29
#69914 abstract reader creation

tabman83:


src/Tests/CaptainHook.Tests/Configuration/KeyVaultProviderTests.cs
```diff
@@ -12,7 +12,7 @@ namespace CaptainHook.Tests.Configuration
 {
     public class KeyVaultProviderTests
     {
-        [Fact]
+        [Fact(Skip = "Superseded by ConfigLoaderTests.ConfigNotEmpty")]
```
chomick, line 15: Do we need that test/code for anything if duplicated?

tabman83: Nope. I just thought it was good practice to deprecate by skipping it before removing it entirely. Can be removed now or in the next PR.

src/CaptainHook.DirectorService/DirectorService.cs
```diff
@@ -58,132 +51,28 @@ protected override async Task RunAsync(CancellationToken cancellationToken)
...
-                        cancellationToken);
-                }
-
-                foreach (var (key, subscriber) in _subscriberConfigurations)
-                {
-                    if (cancellationToken.IsCancellationRequested) return;
-
-                    await RecreateServiceForSubscriberAsync(subscriber, serviceList, cancellationToken);
+                    var description = new ServiceCreationDescription(ServiceNaming.EventHandlerServiceFullName, ServiceNaming.EventHandlerActorServiceType);
+                    await _fabricClientWrapper.CreateServiceAsync(description, cancellationToken);
```
chomick, line 60: Should that not happen for each subscriber config?

src/CaptainHook.DirectorService/FabricClientWrapper.cs
```diff
@@ -0,0 +1,68 @@
...
+using System.Threading.Tasks;
+
+namespace CaptainHook.DirectorService
+{
+    /// <summary>
+    /// A simple wrapper around FabricClient
+    /// </summary>
+    public class FabricClientWrapper : IFabricClientWrapper
+    {
+        private readonly FabricClient fabricClient;
```
chomick, line 18: Can you use _ prefix for local fields?

src/CaptainHook.DirectorService/SubscriptionManager.cs
```diff
@@ -0,0 +1,64 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace CaptainHook.DirectorService
+{
+    public class SubscriptionManager
+    {
+        private readonly IFabricClientWrapper fabricClientWrapper;
```
chomick, line 13: Please use "_" prefix

src/CaptainHook.DirectorService/FabricClientWrapper.cs
dawidkeshop: Not a problem. Would be nice to define one convention in a project anyway.

PR #86, opened 2020-06-02
69916 recreate reader services based on diff result

dawidkeshop:


src/CaptainHook.DirectorService/DirectorService.cs
```diff
@@ -60,8 +62,8 @@ protected override async Task RunAsync(CancellationToken cancellationToken)
                     await _fabricClientWrapper.CreateServiceAsync(description, cancellationToken);
                 }
 
-                var manager = new SubscriptionManager(_fabricClientWrapper, serviceList, _webhookConfigurations);
-                await manager.CreateAsync(_subscriberConfigurations, cancellationToken);
+                var manager = new ReaderServicesManager(_fabricClientWrapper, serviceList, _webhookConfigurations);
+                await manager.CreateAsync(new ReadOnlyDictionary<string, SubscriberConfiguration>(_subscriberConfigurations), cancellationToken);
```
tabman83, line 66: I don't see why the ReaderServicesManager cannot be an injected dependency here.

src/CaptainHook.DirectorService/DirectorService.cs
```diff
@@ -70,10 +72,20 @@ protected override async Task RunAsync(CancellationToken cancellationToken)
...
             var configuration = Configuration.Load();
-            return Task.CompletedTask;
+
+            var comparisonResult = new SubscriberConfigurationComparer().Compare(_subscriberConfigurations, configuration.SubscriberConfigurations);
+
+            var serviceList = await _fabricClientWrapper.GetServiceUriListAsync();
+            var manager = new ReaderServicesManager(_fabricClientWrapper, serviceList, new ReadOnlyCollection<WebhookConfig>(configuration.WebhookConfigurations));
+            await manager.CreateAsync(comparisonResult.Added, CancellationToken.None);
+            await manager.DeleteAsync(comparisonResult.Removed, CancellationToken.None);
+            await manager.RefreshAsync(comparisonResult.Changed, CancellationToken.None);
```
tabman83, line 85: This is gonna take a huge amount of time. Have you been able to run some checks locally with the whole KV reload to see if we hit some remoting timeout?

src/CaptainHook.DirectorService/SubscriptionManager.cs
```diff
@@ -1,27 +1,36 @@
...
+using CaptainHook.Common.Configuration;
+using CaptainHook.Common.ServiceModels;
 
 namespace CaptainHook.DirectorService
 {
-    public class SubscriptionManager
+    /// <summary>
+    /// Allows to create, delete and refresh Reader Services instances.
+    /// </summary>
+    public class ReaderServicesManager
```
tabman83, line 14: @chomick suggested [here](https://github.com/eShopWorld/captain-hook/pull/85#pullrequestreview-422623460) that the use of the word "manager" is somewhat a code smell so may be you want to use a different name again.

src/CaptainHook.DirectorService/SubscriptionManager.cs
```diff
@@ -1,27 +1,36 @@
...
+        /// <param name="serviceList">List of currently deployed services.</param>
+        /// <param name="webhookConfigurations">Recent list of webhook configurations</param>
+        public ReaderServicesManager(IFabricClientWrapper fabricClientWrapper, IList<string> serviceList, IList<WebhookConfig> webhookConfigurations)
         {
             _fabricClientWrapper = fabricClientWrapper;
             _serviceList = serviceList;
             _webhookConfigurations = webhookConfigurations;
         }
 
-        public async Task CreateAsync(IDictionary<string, SubscriberConfiguration> subscriberConfigurations, CancellationToken cancellationToken)
```
tabman83, line 24: For create, delete and refresh you are using the discard pattern meaning you are only interested in the SubscriberConfiguration part from the subscriberConfigurations parameter, so may be you can change the signature of those methods to only accept a collection of SubscriberConfiguration elements.

tabman83:
Also,

- SubscriptionsManager file needs to be renamed accordingly to the class inside
- ReaderServicesManager/SubscriptionsManager needs to implement an interface (that's on me - I meant to extract if afterwards and then I forgot)

src/CaptainHook.DirectorService/DirectorService.cs
```diff
@@ -70,10 +72,20 @@ protected override async Task RunAsync(CancellationToken cancellationToken)
             }
         }
 
-        public Task ReloadConfigurationForEventAsync(string eventName)
+        public async Task ReloadConfigurationForEventAsync(string eventName)
         {
             var configuration = Configuration.Load();
-            return Task.CompletedTask;
+
+            var comparisonResult = new SubscriberConfigurationComparer().Compare(_subscriberConfigurations, configuration.SubscriberConfigurations);
```
tabman83, line 79: we could store SubscriberConfigurationComparer in a static field to avoid the initialization cost

src/CaptainHook.DirectorService/DirectorService.cs
dawidkeshop: No, I didn't test whole KV reload that way. But in practice it won't be a whole KV reload, only the changed subscribers. I believe that there won't be more than few of these at once, so it shouldn't be a problem.
What is the timeout?

src/CaptainHook.DirectorService/SubscriptionManager.cs
dawidkeshop: I remember, but at that point of time it made sense to name this class `ReaderServicesCreator` since it had only option to create. Now it also deletes and refreshes, so `Creator` would be confusing.

I don't like `Manager` also, but I couldn't figure out anything better for this particular scenario. On the other hand, we use `ServiceManager` from FabricClient library, so maybe it's not that bad name? 

Anyway, I'm open for suggestions.

src/CaptainHook.DirectorService/DirectorService.cs
tabman83: Should be 5 minutes ([reference](https://docs.microsoft.com/en-us/dotnet/api/microsoft.servicefabric.services.remoting.fabrictransport.fabrictransportremotingsettings.operationtimeout?view=azure-dotnet))

src/CaptainHook.DirectorService/DirectorService.cs
chomick: This should be injected as a dependency to allow for better testability - we would need to introduce a interface as well.
The cost of instantiating a class is zero for what we are doing here but if it was a dependency, it would be even better.

src/CaptainHook.DirectorService/DirectorService.cs
chomick: Yes, the delta should be relatively low.
Is there any other way, Nino, that you think we can handle that?

src/CaptainHook.DirectorService/DirectorService.cs
```diff
@@ -70,10 +72,20 @@ protected override async Task RunAsync(CancellationToken cancellationToken)
...
+
+            var serviceList = await _fabricClientWrapper.GetServiceUriListAsync();
+            var manager = new ReaderServicesManager(_fabricClientWrapper, serviceList, new ReadOnlyCollection<WebhookConfig>(configuration.WebhookConfigurations));
+            await manager.CreateAsync(comparisonResult.Added, CancellationToken.None);
```
chomick, line 83: Should we not have a cancellation token provided here somehow?

src/CaptainHook.DirectorService/DirectorService.cs
```diff
@@ -70,10 +72,20 @@ protected override async Task RunAsync(CancellationToken cancellationToken)
...
+            await manager.DeleteAsync(comparisonResult.Removed, CancellationToken.None);
+            await manager.RefreshAsync(comparisonResult.Changed, CancellationToken.None);
+
+            _subscriberConfigurations = configuration.SubscriberConfigurations;
```
chomick, line 87: What will happen when the reload method gets called while another reload is taking place?
Should this be atomic operation?

src/CaptainHook.DirectorService/SubscriptionManager.cs
```diff
@@ -36,6 +45,37 @@ public async Task CreateAsync(IDictionary<string, SubscriberConfiguration> subsc
...
+        public async Task RefreshAsync(IReadOnlyDictionary<string, SubscriberConfiguration> subscriberConfigurations, CancellationToken cancellationToken)
+        {
+            foreach (var (_, subscriber) in subscriberConfigurations)
+            {
+                var (newName, oldNames) = FindServiceNames(subscriber);
+                var initializationData = BuildInitializationData(subscriber);
+                var description = new ServiceCreationDescription(newName, ServiceNaming.EventReaderServiceType, initializationData);
+
+                await _fabricClientWrapper.CreateServiceAsync(description, cancellationToken);
+                foreach (var oldName in oldNames.Where(n => n != null))
```
chomick, line 72: Someone mentioned that we have to create and then delete.
Is there any reason why we have to do it in that order? Not sure if it matters hugely at the moment but if we needed to optimise/refactor  the code in any way, would we have freedom to perform those operations in any order or there is a valid reason for certain ordering?

src/CaptainHook.DirectorService/SubscriptionManager.cs
```diff
@@ -36,6 +45,37 @@ public async Task CreateAsync(IDictionary<string, SubscriberConfiguration> subsc
             }
         }
 
+        public async Task DeleteAsync(IReadOnlyDictionary<string, SubscriberConfiguration> subscriberConfigurations, CancellationToken cancellationToken)
+        {
+            foreach (var (_, subscriber) in subscriberConfigurations)
```
chomick, line 50: Are we logging what happens with the services anywhere?
What logs would we want to see if we were to debug CH?
We could have:
- readers that are deleted (at least their names but maybe we also want to see what config they are initialised with?)
- readers that are added
- readers that are updated (again: deleted/added)
- maybe some stats at the beginning on how many readers we plan to add/delete/update (or which of them?)

src/CaptainHook.DirectorService/SubscriptionManager.cs
chomick: We can keep it as is

src/CaptainHook.DirectorService/SubscriptionManager.cs
dawidkeshop: Good idea, thanks!

src/CaptainHook.DirectorService/SubscriptionManager.cs
dawidkeshop: If we delete first, then for a some time (probably dozens of seconds) before new instance starts, messages won't be processed, so we will loose continuity of message delivery. 
Architects insisted on this particular order, so I think better to talk with them if we want to change.

src/CaptainHook.DirectorService/DirectorService.cs
dawidkeshop: Sure, will fix it.

src/CaptainHook.DirectorService/SubscriptionManager.cs
dawidkeshop: Sure, will add logging.

src/CaptainHook.DirectorService/DirectorService.cs
dawidkeshop: I've added it mostly for consistency. I'm not sure if we can provide a token here, so maybe better to remove it.

src/CaptainHook.DirectorService/DirectorService.cs
dawidkeshop: Actually, this one can be even static, it's a stable implementation without any state. 
The real problem with testability here is `Configuration.Load()`, so I think I must redesign this part a bit, maybe this code ends up in "Manager" class.

src/CaptainHook.DirectorService/ReaderServicesManager.cs
```diff
@@ -0,0 +1,140 @@
...
+        /// <returns></returns>
+        public async Task RefreshReadersAsync(Configuration newConfiguration, IDictionary<string, SubscriberConfiguration> currentSubscribers, IList<string> serviceList)
+        {
+            var comparisonResult = new SubscriberConfigurationComparer().Compare(currentSubscribers, newConfiguration.SubscriberConfigurations);
+
+            _bigBrother.Publish(new RefreshSubscribersEvent(comparisonResult));
+
+            await CreateAsync(comparisonResult.Added.Values, serviceList, newConfiguration.WebhookConfigurations);
+            await DeleteAsync(comparisonResult.Removed.Values, serviceList, newConfiguration.WebhookConfigurations);
+            await RefreshAsync(comparisonResult.Changed.Values, serviceList, newConfiguration.WebhookConfigurations, CancellationToken.None);
```
tabman83, line 63: This should accept a cancellation token and pass it down to all three methods

src/CaptainHook.DirectorService/DirectorService.cs
```diff
@@ -74,18 +76,21 @@ protected override async Task RunAsync(CancellationToken cancellationToken)
 
         public async Task ReloadConfigurationForEventAsync(string eventName)
         {
-            var configuration = Configuration.Load();
-
-            var comparisonResult = new SubscriberConfigurationComparer().Compare(_subscriberConfigurations, configuration.SubscriberConfigurations);
+            await _semaphore.WaitAsync();
```
tabman83, line 79: you can use a lock on an object ([here](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/lock-statement))

src/CaptainHook.DirectorService/Events/ServiceCreatedEvent.cs
```diff
@@ -0,0 +1,18 @@
...
+    {
+        public string Message { get; set; }
+        public string ReaderName { get; set; }
+        public string Configuration { get; set; }
+
+        public ServiceCreatedEvent(string readerName, string configuration)
+        {
+            ReaderName = readerName;
+            Configuration = configuration;
+            Message = $"Created Service: {readerName}";
```
tabman83, line 15: The message is redundant, to keep consumption down we may only send the reader name and configuration (if ever needed) - also we can infer the event type from the class name hence why no need for the string message.

src/CaptainHook.DirectorService/Events/ServiceDeletedEvent.cs
```diff
@@ -0,0 +1,16 @@
...
+    class ServiceDeletedEvent : TelemetryEvent
+    {
+        public string Message { get; set; }
+        public string ReaderName { get; set; }
+
+        public ServiceDeletedEvent(string readerName)
+        {
+            ReaderName = readerName;
+            Message = $"Created Service: {readerName}";
+        }
```
tabman83, line 14: Same as above, logging the reader name is enough

src/CaptainHook.DirectorService/Events/RefreshSubscribersEvent.cs
```diff
@@ -0,0 +1,21 @@
...
+        public string ReadersToDelete { get; set; }
+        public string ReadersToRefresh { get; set; }
+
+        public RefreshSubscribersEvent(SubscriberConfigurationComparer.Result result)
+        {
+            Message = $"Number of Readers to add: {result.Added.Count} to delete: {result.Removed.Count} and to refresh: {result.Changed.Count}";
+            ReadersToAdd = string.Join(',', result.Added.Keys);
+            ReadersToDelete = string.Join(',', result.Removed.Keys);
+            ReadersToRefresh = string.Join(',', result.Changed.Keys);
+        }
```
tabman83, line 19: we could compact this log message a little bit by skipping the text message, having a count key for each (added/removed/changed) and an array of names for each.

src/CaptainHook.DirectorService/IReaderServicesManager.cs
```diff
@@ -0,0 +1,30 @@
...
+        /// <summary>
+        /// Compares newly read Configuration with list of currently deployed subscribers and based on that create new, delete old
+        /// and refresh (by pair of create and delete operation) existing readers.
+        /// </summary>
+        /// <param name="newConfiguration">Target Configuration to be deployed</param>
+        /// <param name="serviceList">List of currently deployed services names</param>
+        /// <param name="currentSubscribers">List of currently deployed subscribers</param>
+        /// <returns></returns>
+        Task RefreshReadersAsync(Configuration newConfiguration, IDictionary<string, SubscriberConfiguration> currentSubscribers, IList<string> serviceList);
+    }
```
tabman83, line 29: See my comment below re. Cancellation token here

src/CaptainHook.DirectorService/ReaderServicesManager.cs
dawidkeshop: Ok, but what would be the real source of cancellation token in that case?

src/CaptainHook.DirectorService/DirectorService.cs
dawidkeshop: Unfortunately I can't:

> Within an async method, you can't use the await operator in the body of a synchronous function, inside the block of a lock statement, and in an unsafe context.

https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/await

src/CaptainHook.DirectorService/Events/RefreshSubscribersEvent.cs
dawidkeshop: Could you elaborate more on this? How it should look like?
I tried to use arrays in telemetry events, but these seems to not be serialized, so I've decided to use `string.Join`.

src/CaptainHook.DirectorService/DirectorService.cs
tabman83: Yeah, thanks

src/CaptainHook.DirectorService/Events/RefreshSubscribersEvent.cs
tabman83: Use serialization so it will still look like an array on AI side.
This is I envision it, but feel free to change:
```c#
new RefreshSubscriberEvent 
{
    AddedCount: result.Added.Count,
    RemovedCount: result.Removed.Count,
    ChangedCount: result.Changed.Count,
    AddedReaders: JsonConvert.SerializeObject(result.Added.Keys),
    RemovedReaders: JsonConvert.SerializeObject(result.Removed.Keys),
    ChangedReaders: JsonConvert.SerializeObject(result.Changed.Keys)
};
```

src/CaptainHook.DirectorService/ReaderServicesManager.cs
tabman83: I think we should save the cancellation token passed to RunAsync on a private variable and pass it there. This way we guarantee that SF has an opportunity to also cancel that operation when in progress during a SF event.

src/CaptainHook.DirectorService/DirectorService.cs
```diff
@@ -70,10 +74,23 @@ protected override async Task RunAsync(CancellationToken cancellationToken)
...
+                var serviceList = await _fabricClientWrapper.GetServiceUriListAsync();
+
+                await _readerServicesManager.RefreshReadersAsync(configuration, _subscriberConfigurations, serviceList);
+
+                _subscriberConfigurations = configuration.SubscriberConfigurations;
+                _webhookConfigurations = configuration.WebhookConfigurations;
+            }
+            finally
+            {
+                _semaphore.Release(1);
```
mszymeshop, line 92: Consider `Mutex` so that you don't have to manage the counter (you have a single resource after all) or just use the Release which defaults to one. I would personally go with Mutex, easier to address a single resource - especially if you look at initialization.

src/CaptainHook.DirectorService/ReaderServicesManager.cs
```diff
@@ -0,0 +1,140 @@
...
+        /// Compares newly read Configuration with list of currently deployed subscribers and based on that create new, delete old
+        /// and refresh (by pair of create and delete operation) existing readers.
+        /// </summary>
+        /// <param name="newConfiguration">Target Configuration to be deployed</param>
+        /// <param name="serviceList">List of currently deployed services names</param>
+        /// <param name="currentSubscribers">List of currently deployed subscribers</param>
+        /// <returns></returns>
+        public async Task RefreshReadersAsync(Configuration newConfiguration, IDictionary<string, SubscriberConfiguration> currentSubscribers, IList<string> serviceList)
+        {
+            var comparisonResult = new SubscriberConfigurationComparer().Compare(currentSubscribers, newConfiguration.SubscriberConfigurations);
```
mszymeshop, line 57: Can we inject Comparer? Easier unit tests?

src/CaptainHook.DirectorService/ReaderServicesManager.cs
```diff
@@ -0,0 +1,140 @@
...
+            }
+        }
+
+        private async Task DeleteAsync(IEnumerable<SubscriberConfiguration> subscribers, IList<string> serviceList, IEnumerable<WebhookConfig> webhooks)
+        {
+            foreach (var subscriber in subscribers)
+            {
+                var (_, oldNames) = FindServiceNames(subscriber, serviceList);
+
+                foreach (var oldName in oldNames.Where(n => n != null))
```
mszymeshop, line 86: Should the `FindServiceNames` do this filtering instead and not return `null` entries?

src/CaptainHook.DirectorService/ReaderServicesManager.cs
```diff
@@ -0,0 +1,140 @@
...
+                {
+                    await _fabricClientWrapper.DeleteServiceAsync(oldName, CancellationToken.None);
```
mszymeshop, line 88: I am thinking whether we should build in some resilience to this. Creation and Deletion of services might time out, we should consider retrying when this happens. What do you think?

src/CaptainHook.DirectorService/ReaderServicesManager.cs
```diff
@@ -0,0 +1,140 @@
...
+                    _bigBrother.Publish(new ServiceDeletedEvent(oldName));
+                }
+            }
+        }
+
+        private async Task RefreshAsync(IEnumerable<SubscriberConfiguration> subscribers, IList<string> serviceList, IEnumerable<WebhookConfig> webhooks, CancellationToken cancellationToken)
```
mszymeshop, line 94: It looks like Refresh == Delete + Create. I think there is a benefit in having a list of services to refresh, but under the hood, these should be broken down into deletes and creates respectively so that the codebase is clearer. Optionally, you might consider a Refresh method which reuses Delete and Create instead.

src/CaptainHook.DirectorService/ReaderServicesManager.cs
dawidkeshop: You're probably right but deletion logic will be significantly changed soon, so there's no point to fix this right now.
And I think we should discuss this within the team.

src/CaptainHook.DirectorService/ReaderServicesManager.cs
dawidkeshop: Yes, you right. This class was significantly changed yesterday to expose new operations and I didn't have enough time to refactor the implementation details.

src/CaptainHook.DirectorService/ReaderServicesManager.cs
dawidkeshop: Actually I see that's not even possible to have `null` here, so I'll remove the check.

src/CaptainHook.Common/Remoting/IReloadConfigurationResult.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace CaptainHook.Common.Remoting
+{
+    public interface IReloadConfigurationResult
```
chomick, line 3: Are we dropping this interface?

src/CaptainHook.DirectorService/DirectorService.cs
```diff
@@ -15,29 +16,33 @@ namespace CaptainHook.DirectorService
 {
     public class DirectorService : StatefulService, IDirectorServiceRemoting
     {
+        private static readonly SemaphoreSlim _semaphore = new SemaphoreSlim(1, 1);
```
chomick, line 19: I wonder if we need to use sempahore or should simple lock be enough?

src/CaptainHook.DirectorService/DirectorService.cs
```diff
@@ -60,8 +65,7 @@ protected override async Task RunAsync(CancellationToken cancellationToken)
                     await _fabricClientWrapper.CreateServiceAsync(description, cancellationToken);
                 }
 
-                var manager = new SubscriptionManager(_fabricClientWrapper, serviceList, _webhookConfigurations);
-                await manager.CreateAsync(_subscriberConfigurations, cancellationToken);
+                await _readerServicesManager.CreateReadersAsync(_subscriberConfigurations.Values, serviceList, _webhookConfigurations, cancellationToken);
```
chomick, line 68: A tiny thing - I would probably reorder the params to be more logical - configuration (subs & webhooks) could be together followed by list of services (or vice versa).
But ok to update it later.

src/CaptainHook.DirectorService/DirectorService.cs
chomick: I have seen comment below, fine to keep it.

src/CaptainHook.DirectorService/DirectorService.cs
```diff
@@ -70,10 +74,23 @@ protected override async Task RunAsync(CancellationToken cancellationToken)
             }
         }
 
-        public Task ReloadConfigurationForEventAsync(string eventName)
+        public async Task ReloadConfigurationAsync()
         {
-            var configuration = Configuration.Load();
-            return Task.CompletedTask;
+            await _semaphore.WaitAsync();
```
chomick, line 79: I think we need to change it slightly.
If you have one massive block here, any subsequent calls via API will be blocking and you will not have a way to communicate anything back.
I would prefer to check if we are in process of refreshing and if yes, return immediately with some flag.

{
  if (isRefreshing)
    return ; // refreshing, nothing to do (maybe return some info about that state or even progress?)

  lock (refreshingSync) { 
    if (isRefreshing)
      return ; // we are actually refreshing, so have to leave

    isRefreshing = true;
  }

  try {
    // not certain how much of critical section we actually need here?
    // once you have the refreshing flag check above, the rest can safely be executed in normal code
  }
  finally {
    lock (refreshingSync) { isRefreshing = false; }
  }

}

src/CaptainHook.DirectorService/DirectorService.cs
dawidkeshop: I choose this order to indicate that webhooks will be removed (we already have tech debt for that).

src/CaptainHook.DirectorService/ReaderServicesManager.cs
chomick: agree with Nino.

src/CaptainHook.DirectorService/ReaderServicesManager.cs
chomick: Agree - let's tackle that with the second story

src/CaptainHook.DirectorService/DirectorService.cs
chomick: ok

src/CaptainHook.DirectorService/DirectorService.cs
```diff
@@ -65,20 +78,49 @@ protected override async Task RunAsync(CancellationToken cancellationToken)
...
+                }
+            }
         }
 
-        public Task ReloadConfigurationForEventAsync(string eventName)
+        public async Task ReloadConfigurationAsync()
         {
-            var configuration = Configuration.Load();
-            return Task.CompletedTask;
+            lock (_refreshSync)
```
chomick, line 99: You can also check the flag before the lock to cater for a case when we are actually still refreshing so we do not have to acquire the lock to check it really. 
But it is an optimisation you can add later.

src/CaptainHook.Api/Controllers/RefreshConfigController.cs
```diff
@@ -32,18 +32,17 @@ public RefreshConfigController(IBigBrother bigBrother)
         /// <summary>
         /// Refreshes configuration for the given event
         /// </summary>
-        /// <param name="request">Request with details to refresh configuration</param>
         /// <returns>If the event name is valid, returns its configuration. If invalid, returns BadRequest</returns>
         [HttpPost]
         [ProducesResponseType(StatusCodes.Status200OK)]
         [ProducesResponseType(StatusCodes.Status404NotFound)]
```
mszymeshop, line 38: 404 no longer valid?

src/CaptainHook.DirectorService/ReaderServicesManager.cs
mszymeshop: Marked as resolved and no changes made.

PR #87, opened 2020-06-02
#69917 implement API response

tabman83:


src/CaptainHook.Api/Controllers/RefreshConfigController.cs
```diff
@@ -38,14 +38,14 @@ public RefreshConfigController(IBigBrother bigBrother)
...
+        public async Task<IActionResult> RefreshConfigForEvent()
         {
             try
             {
                 var directorServiceClient = ServiceProxy.Create<IDirectorServiceRemoting>(new Uri(ServiceNaming.DirectorServiceFullName));
-                await directorServiceClient.ReloadConfigurationForEventAsync(request.EventName);
+                var result = await directorServiceClient.ReloadConfigurationForEventAsync();
 
-                return Ok();
+                return Ok(result);
```
chomick, line 48: The contract for API should be separate from service contract IMO.

src/CaptainHook.Common/Remoting/IReloadConfigurationResult.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace CaptainHook.Common.Remoting
+{
+    public interface IReloadConfigurationResult
```
chomick, line 3: Any reason we need interface here?
This is just data object, isn't it?

src/CaptainHook.Api/Controllers/RefreshConfigController.cs
```diff
@@ -35,17 +35,24 @@ public RefreshConfigController(IBigBrother bigBrother)
...
 
-                return Ok();
+                var apiResult = new RefreshConfigResultDto
+                {
+                    Added = operationResult.Added,
+                    Removed = operationResult.Removed,
+                    Changed = operationResult.Changed
+                };
+
+                return Ok(apiResult);
```
dawidkeshop, line 55: I'm just wondering if OK is a proper status code to return in case of some long-term process. Shouldn't it be rather Accepted (201)?

tabman83: You are right, this will be changed in light of changes to the rest process.

src/CaptainHook.Api/Controllers/RefreshConfigController.cs
```diff
@@ -35,17 +35,24 @@ public RefreshConfigController(IBigBrother bigBrother)
         /// <param name="request">Request with details to refresh configuration</param>
         /// <returns>If the event name is valid, returns its configuration. If invalid, returns BadRequest</returns>
         [HttpPost]
-        [ProducesResponseType(StatusCodes.Status200OK)]
+        [ProducesResponseType(typeof(RefreshConfigResultDto), StatusCodes.Status200OK)]
         [ProducesResponseType(StatusCodes.Status404NotFound)]
         [ProducesResponseType(StatusCodes.Status400BadRequest)]
-        public async Task<IActionResult> RefreshConfigForEvent([FromBody]RefreshConfigRequest request)
+        public async Task<IActionResult> RefreshConfigForEvent()
```
mszymeshop: If it is a general refresh, consider changing the name so that it does not say `ForEvent`

src/CaptainHook.Api/Controllers/RefreshConfigController.cs
```diff
@@ -38,14 +38,14 @@ public RefreshConfigController(IBigBrother bigBrother)
         [ProducesResponseType(StatusCodes.Status200OK)]
```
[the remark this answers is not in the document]
tabman83: I am not gonna have a dto response in this pr as the base pr does not return the number of added/removed/changed subscribers for now.

src/CaptainHook.Api/Controllers/RefreshConfigController.cs
tabman83: Not needed on the basis of the change above

src/CaptainHook.Common/Remoting/IReloadConfigurationResult.cs
tabman83: Changed as well

src/CaptainHook.Api/Controllers/RefreshConfigController.cs
mszymeshop: I guess this suggestion makes sense, but at the same time isn't the current implementation blocking? Should we change this under the hood? Also, do we know what kind of wait times we should expect for this operation?

src/CaptainHook.Common/Remoting/IDirectorServiceRemoting.cs
```diff
@@ -5,6 +5,6 @@ namespace CaptainHook.Common.Remoting
 {
     public interface IDirectorServiceRemoting: IService
     {
-        Task ReloadConfigurationForEventAsync(string eventName);
+        Task<IReloadConfigurationResult> ReloadConfigurationForEventAsync();
```
mszymeshop: `ForEvent` present in the name

PR #88, opened 2020-06-04
Fix issues with old reader (a/b) and partitions for event reader

tabman83:


src/CaptainHook.DirectorService/IFabricClientWrapper.cs
```diff
@@ -18,12 +19,14 @@ public class ServiceCreationDescription
         public string ServiceName { get; }
         public string ServiceTypeName { get; }
         public byte[] InitializationData { get; }
+        public PartitionSchemeDescription PartitionScheme { get; }
 
-        public ServiceCreationDescription(string serviceName, string serviceTypeName, byte[] initializationData = null)
+        public ServiceCreationDescription(string serviceName, string serviceTypeName, PartitionSchemeDescription partitionScheme, byte[] initializationData = null)
```
chomick, line 24: How do we use that? Did you mean to use it below?

PR #89, opened 2020-06-04
#70680 Add reader service versioning logic

tabman83:


src/CaptainHook.DirectorService/DateTimeProvider.cs
```diff
@@ -0,0 +1,9 @@
+﻿using System;
+
+namespace CaptainHook.DirectorService
+{
+    public class DateTimeProvider : IDateTimeProvider
+    {
+        public DateTimeOffset UtcNow => DateTimeOffset.UtcNow;
```
chomick, line 7: why offset, not just datetime?

src/CaptainHook.DirectorService/IReaderServicesManager.cs
```diff
@@ -26,5 +26,7 @@ public interface IReaderServicesManager
         /// <param name="currentSubscribers">List of currently deployed subscribers</param>
         /// <returns></returns>
         Task RefreshReadersAsync(Configuration newConfiguration, IDictionary<string, SubscriberConfiguration> currentSubscribers, IList<string> serviceList, CancellationToken cancellationToken);
+
+        (string newName, IEnumerable<string> oldNames) FindServiceNames(SubscriberConfiguration subscriber, IList<string> serviceList);
```
chomick, line 30: i know this was the internal contract before but as you expose it as pyblic now, should we make it more explicit ie wrap the response in a class?

src/CaptainHook.DirectorService/ReaderServicesManager.cs
```diff
@@ -97,19 +100,22 @@ private async Task DeleteReaderServicesAsync(IEnumerable<string> oldNames, Cance
...
+
             var readerServiceNameUri = ServiceNaming.EventReaderServiceFullUri(subscriber.EventType, subscriber.SubscriberName, subscriber.DLQMode != null);
 
-            var names = new[] { readerServiceNameUri, $"{readerServiceNameUri}-a", $"{readerServiceNameUri}-b" };
+            var possibleOldNames = new[] { readerServiceNameUri, $"{readerServiceNameUri}-a", $"{readerServiceNameUri}-b" };
 
-            var oldNames = serviceList.Intersect(names);
+            var oldNames = new List<string>();
+            oldNames.AddRange(serviceList.Intersect(possibleOldNames));
+            oldNames.AddRange(serviceList.Where(x => Regex.IsMatch(x, @"-\d{10}$")));
```
chomick, line 116: can you simply look for names starting with reader uri? this should be unique enough and faster then regexp

src/CaptainHook.DirectorService/ReaderServicesManager.cs
```diff
@@ -97,19 +100,22 @@ private async Task DeleteReaderServicesAsync(IEnumerable<string> oldNames, Cance
             }
         }
 
-        private static (string newName, IEnumerable<string> oldNames) FindServiceNames(SubscriberConfiguration subscriber, IList<string> serviceList)
+        public (string newName, IEnumerable<string> oldNames) FindServiceNames(SubscriberConfiguration subscriber, IList<string> serviceList)
         {
+            var ticksPerSecond = 10_000_000;
+
+            // generates a new id every minute
+            var newSuffix = (_dateTimeProvider.UtcNow.Ticks / (60 * ticksPerSecond)).ToString();
```
chomick, line 108: is 1 minute window small enough? are we saying that we will never update config for event more than once in a min?

i would not link it to time window.
if you want a number, i suggest we simply use sequential ints but fill the gaps, ie if you have 1, 2, 5 as current names, pick 3 as new name.
if you prefer to have something that indicates the latest version, you can either use utcnow (at least with seconds but maybe even millisecond?) or just ignore gaps aboce and alway pick max+1 as new name.

src/Tests/CaptainHook.Tests/Director/ReaderServicesManagerTests.cs
```diff
@@ -0,0 +1,102 @@
...
+            };
+            _readerServicesManager = new ReaderServicesManager(fabricClientWrapper, bigBrother, _dateTimeProvider);
+        }
+
+        [Fact, IsLayer0]
+        public void CanGenerateServiceName()
+        {
+            var serviceList = new string[] { };
+
+            var time = new DateTimeOffset(637268886191109359, TimeSpan.Zero);
```
chomick, line 38: you could move it to a func with some params so you dont have to repeat this part in every test.

src/Tests/CaptainHook.Tests/Director/ReaderServicesManagerTests.cs
```diff
@@ -0,0 +1,102 @@
...
+                newName2.Should().Be("fabric:/CaptainHook/EventReader.customer.product.core.events.productrefreshevent-captain-hook-1062114811");
+            }
+        }
+
+        [Fact, IsLayer0]
+        public void CanFindOldServiceNames()
+        {
+            var subscriber = new SubscriberConfiguration()
+            {
+                EventType = "customer.product.core.events.productrefreshevent",
```
chomick, line 72: you could have picked a shorter name for testing :)

src/CaptainHook.DirectorService/IReaderServicesManager.cs
dawidkeshop: This method should not be exposed here, it should be moved to a new class.
It's an implementation detail made public only to allow unit testing. If we think that this algorithm is so complicated that deserves for own tests suite, it deserves for own class also.
And it's not an services management action, it's different level of abstraction than creating or refreshing readers. It won't be called from DirectorService which uses ReaderServiceManager, so DirectorService should not have access to it.

dawidkeshop: This method should not be exposed here, it should be moved to a new class.
It's an implementation detail made public only to allow unit testing. If we think that this algorithm is so complicated that deserves for own tests suite, it deserves for own class also.
And it's not an services management action, it's different level of abstraction than creating or refreshing readers. It won't be called from DirectorService which uses ReaderServiceManager, so DirectorService should not have access to it.

src/CaptainHook.DirectorService/DateTimeProvider.cs
tabman83: DateTimeOffset is not affected by leap seconds, as being the instantaneous time.

src/Tests/CaptainHook.Tests/Director/ReaderServicesManagerTests.cs
tabman83: Moved to a constant

src/Tests/CaptainHook.Tests/Director/ReaderServicesManagerTests.cs
tabman83: Done

src/CaptainHook.DirectorService/ReaderServicesManager.cs
tabman83: Changed as discussed

src/CaptainHook.DirectorService/ReaderServicesManager.cs
tabman83: Changed as discussed - using regex and subscriber name to match all situations

src/CaptainHook.DirectorService/IReaderServicesManager.cs
tabman83: Moved to separate class as discussed

src/Tests/CaptainHook.Tests/Director/ReaderServiceNameGeneratorTests.cs
```diff
@@ -0,0 +1,97 @@
...
+using FluentAssertions.Execution;
+using Moq;
+using System;
+using Xunit;
+
+namespace CaptainHook.Tests.Director
+{
+    public class ReaderServiceNameGeneratorTests
+    {
+        private IReaderServiceNameGenerator _readerServiceNameGenerator;
```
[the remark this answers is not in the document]
tabman83: good point changing it

src/CaptainHook.DirectorService/Infrastructure/ReaderServiceNameGenerator.cs
```diff
@@ -0,0 +1,36 @@
...
+using CaptainHook.Common.Configuration;
+using CaptainHook.DirectorService.Infrastructure.Interfaces;
+using System.Collections;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text.RegularExpressions;
+
+namespace CaptainHook.DirectorService.Infrastructure
+{
+    public class ReaderServiceNameGenerator: IReaderServiceNameGenerator
```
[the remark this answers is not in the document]
chomick: I guess for testability?

src/CaptainHook.DirectorService/Infrastructure/ReaderServiceNameGenerator.cs
```diff
@@ -0,0 +1,36 @@
...
+    {
+        private IDateTimeProvider _dateTimeProvider;        
+
+        public ReaderServiceNameGenerator(IDateTimeProvider dateTimeProvider)
+        {
+            _dateTimeProvider = dateTimeProvider;
+        }
+
+        public string GenerateNewName(SubscriberConfiguration subscriber)
```
chomick, line 20: For clean contract, you should not need entire subscriber as it is unclear what needs to be provided.
It will later help with the code refactoring as well when we start changing the models.
I would introduce a data object that has all the bits that you require here and nothing more :)

tabman83: I can change that

src/Tests/CaptainHook.Tests/Director/ReaderServiceNameGeneratorTests.cs
```diff
@@ -0,0 +1,97 @@
...
+            {
+                newName1.Should().Be(FullEventName(suffix: GetMillisecondsAsString(time1)));
+                newName2.Should().Be(FullEventName(suffix: GetMillisecondsAsString(time2)));
+            }
+        }
+
+        [Fact, IsLayer0]
+        public void CanFindOldServiceNames()
+        {
+            var serviceList = new[]
```
chomick, line 74: I would like to see some cases that try to break the regexp, maybe some funny stuff after the numbers etc - just to ensure we filter the stuff properly

PR #90, opened 2020-06-05
#69916 - add async reload and notify in api response

mszymeshop:
### What
Changes to config reload and API response:
- The logic for double-if has been changed so that both ifs result in the same operation and yield start of reload
- When reload in progress, we issue a new event
- 202 (Accepted) status code is issued when a reload has been started, 409 (Conflict) is issued when another reload is in progress already.


src/CaptainHook.Api/Controllers/RefreshConfigController.cs
```diff
@@ -29,24 +27,30 @@ public RefreshConfigController(IBigBrother bigBrother)
...
-                var directorServiceClient = ServiceProxy.Create<IDirectorServiceRemoting>(new Uri(ServiceNaming.DirectorServiceFullName));
-                await directorServiceClient.ReloadConfigurationAsync();
+                var directorServiceUri = new Uri(ServiceNaming.DirectorServiceFullName);
+                var directorServiceClient = ServiceProxy.Create<IDirectorServiceRemoting>(directorServiceUri);
+                var requestReloadConfigurationResult = await directorServiceClient.RequestReloadConfigurationAsync();
 
-                return Ok();
+                return requestReloadConfigurationResult == RequestReloadConfigurationResult.ReloadStarted
+                    ? (IActionResult) Accepted()
+                    : NoContent();
```
dawidkeshop, line 51: Wouldn't be better to return 409 (Conflict) here? It indicates that the request could not be completed due to a conflict with the current state of the resource. I think this is what is happening in our case.

204 indicates that everything is OK, just no data need to be returned, but in this situation it wasn't OK because we actually refuse to process the request.

src/CaptainHook.DirectorService/DirectorService.cs
```diff
@@ -94,19 +97,27 @@ protected override async Task RunAsync(CancellationToken cancellationToken)
...
+                    }
                 }
             }
 
+            _bigBrother.Publish(new RefreshConfigRequestedWhenAnotherInProgressEvent());
+            return Task.FromResult(RequestReloadConfigurationResult.ReloadInProgress);
+        }
+
+        private async void ExecuteConfigRefresh(object state)
+        {
```
tabman83, line 120: We could have a timed telemetry event here to report completion and duration of config refresh

src/CaptainHook.Api/Controllers/RefreshConfigController.cs
mszymeshop: Indeed we should be saying, please come back later and retry your request. I'll change to 409.

src/CaptainHook.DirectorService/DirectorService.cs
mszymeshop: I'll add it

PR #92, opened 2020-06-08
69915 unit tests for ReaderServicesManager

dawidkeshop:


src/Tests/CaptainHook.Tests/Builders/SubscriberConfigurationBuilder.cs
```diff
@@ -53,7 +53,7 @@ public SubscriberConfigurationBuilder WithOidcAuthentication()
             return this;
         }
 
-        public SubscriberConfigurationBuilder WithCallback(string uri)
+        public SubscriberConfigurationBuilder WithCallback(string uri = "https://calback.eshopworld.com")
```
tabman83, line 56: Typo

src/Tests/CaptainHook.Tests/Director/ReaderServicesManagerTests.cs
```diff
@@ -0,0 +1,227 @@
...
+                "fabric:/CaptainHook/EventReader.testevent-subscriber1-10000000000000",
+                "fabric:/CaptainHook/EventReader.testevent.completed-captain-hook-10000000000000");
+
+            VerifyFabricClientDeleteCalls();
+
+            VerifyServiceCreatedEventPublished("fabric:/CaptainHook/EventReader.testevent-captain-hook-10000000000000",
+               "fabric:/CaptainHook/EventReader.testevent-subscriber1-10000000000000",
+               "fabric:/CaptainHook/EventReader.testevent.completed-captain-hook-10000000000000");
+
+            VerifyServiceDeletedEventPublished();
```
tabman83, line 70: You would probably need an AssertionScope() here

src/Tests/CaptainHook.Tests/Builders/SubscriberConfigurationBuilder.cs
dawidkeshop: Good catch, thanks!

src/Tests/CaptainHook.Tests/Director/ReaderServicesManagerTests.cs
dawidkeshop: Great idea, thanks!

src/Tests/CaptainHook.Tests/Director/ReaderServicesManagerTests.cs
```diff
@@ -0,0 +1,237 @@
...
+        {
+            _dateTimeProviderMock.SetupGet(x => x.UtcNow)
+                .Returns(DateTimeOffset.MinValue.AddMilliseconds(10000000000000));
+            var generator = new ReaderServiceNameGenerator(_dateTimeProviderMock.Object);
+            return new ReaderServicesManager(_fabricClientMock.Object, _bigBrotherMock.Object, _dateTimeProviderMock.Object, generator);
+        }
+
+        private void ProgressTime(int milliseconds)
+        {
+            _dateTimeProviderMock.SetupGet(x => x.UtcNow).Returns(DateTimeOffset.MinValue.AddMilliseconds(10000000000000 + milliseconds));
```
chomick, line 49: You could simply 'get' the current one and set the expectation with increased value - no need to hardcode the base then

src/Tests/CaptainHook.Tests/Director/ReaderServicesManagerTests.cs
```diff
@@ -0,0 +1,237 @@
...
+
+        private readonly Mock<IBigBrother> _bigBrotherMock = new Mock<IBigBrother>();
+        private readonly Mock<IFabricClientWrapper> _fabricClientMock = new Mock<IFabricClientWrapper>();
+        private readonly Mock<IDateTimeProvider> _dateTimeProviderMock = new Mock<IDateTimeProvider>();
+
+        private ReaderServicesManager CreateReaderServiceManager()
+        {
+            _dateTimeProviderMock.SetupGet(x => x.UtcNow)
+                .Returns(DateTimeOffset.MinValue.AddMilliseconds(10000000000000));
+            var generator = new ReaderServiceNameGenerator(_dateTimeProviderMock.Object);
```
chomick, line 43: We could mock that as well maybe?

src/CaptainHook.DirectorService/Infrastructure/ReaderServiceNameGenerator.cs
```diff
@@ -26,7 +26,7 @@ public string GenerateNewName(SubscriberNaming naming)
             return $"{readerServiceNameUri}-{newSuffix}";
         }
 
-        public IList<string> FindOldNames(SubscriberNaming naming, IList<string> serviceList)
+        public IList<string> FindOldNames(SubscriberNaming naming, IEnumerable<string> serviceList)
```
chomick, line 29: Do we have unit tests for that class and its two methods?

dawidkeshop: Yes we do, both methods was well covered as part of Nino's task. 
This change is only result of refactoring, we never need any IList features for that list, so I decided to change it everywhere to IEnumerable.

src/Tests/CaptainHook.Tests/Director/ReaderServicesManagerTests.cs
dawidkeshop: You're right, thanks!

PR #93, opened 2020-06-09
Remove conditional compilation for auth mode

tabman83:
### What 
Removes a conditional compilation constant that allows to bypass authentication when in debug mode. The reason why this was originally brought in is because this is the same approach used in Bullfrog and it seemed reasonable for consistency.

### Why
The pipeline is currently building and releasing in debug mode so authentication won't be enabled. We clearly don't want this.


tabman83:
Closing as not needed.

PR #96, opened 2020-06-12
#71417 remove feature flag and ensure cosmos db communication works

tabman83:


tabman83:
I have updated the load step to use the CH config section in order to reflect the KV setup.

src/CaptainHook.DirectorService/Program.cs
```diff
@@ -97,18 +93,6 @@ private static async Task Main()
             }
         }
 
-        /// <summary>
-        /// Configures Feature Flags based on input configuration and register the instance with the container.
-        /// </summary>
-        private static FeatureFlagsConfiguration ConfigureFeatureFlags(IConfiguration config, ContainerBuilder builder)
-        {
-            var featureFlags = new FeatureFlagsConfiguration();
```
mszymeshop, line 105: `FeatureFlagsConfiguration` class can be removed now that it is no longer used.

tabman83: It has been in fact removed, you are probably looking at an old commit.

src/CaptainHook.DirectorService/DirectorService.cs
```diff
@@ -28,6 +29,7 @@ public class DirectorService : StatefulService, IDirectorServiceRemoting
         private readonly IBigBrother _bigBrother;
         private readonly IFabricClientWrapper _fabricClientWrapper;
         private readonly IReaderServicesManager _readerServicesManager;
+        private readonly ICosmosDbRepository _cosmosDbRepository;
```
chomick, line 32: Why do we need that in Director?

chomick: I get we wanted to test it intially but we will not reference that repo here, right?

tabman83: yeah needs to be removed

PR #97, opened 2020-06-14
#71419 business domain data models

tabman83:
Introduce business domain data models for events, subscribers, etc.


src/CaptainHook.Domain/Models/Endpoint.cs
```diff
@@ -0,0 +1,10 @@
+﻿namespace CaptainHook.Domain.Models
+{
+    public class Endpoint
+    {
+        public string Uri { get; set; }
+        public string Selector { get; set; }
+        public string Authentication { get; set; }
```
mszymeshop, line 7: Is it a string for temporary purposes only?

src/CaptainHook.Domain/Models/Event.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System.Collections.Generic;
+
+namespace CaptainHook.Domain.Models
+{
+    public class Event        
+    {
+        public string Name { get; set; }
+        public string Type { get; set; }
```
mszymeshop, line 8: Should this be an enum of enumerated values?

src/CaptainHook.Domain/Models/Subscriber.cs
```diff
@@ -0,0 +1,10 @@
+﻿namespace CaptainHook.Domain.Models
+{
+    public class Subscriber
+    {
+        public string Name { get; set; }
+        public Webhook Webhooks { get; set; }
```
dawidkeshop, line 6: Singular name of the type and plural name of property looks unusual. Wouldn't be better to have a `public Webhook Webhook` as well as `public Webhook Callback`?

src/CaptainHook.Domain/Models/Subscriber.cs
```diff
@@ -0,0 +1,10 @@
+﻿namespace CaptainHook.Domain.Models
+{
+    public class Subscriber
```
mszymeshop, line 3: If this is for domain models, have you thought about partitioning? Looks like you want to use the Event as root but maybe it makes sense to make Subscriber the root? Let me know if you have thought about this.

src/CaptainHook.Domain/Models/Subscriber.cs
chomick: What about "WebhooksDefinition"? Or at least "WebhooksModel"?

src/CaptainHook.Domain/Models/Event.cs
tabman83: this is the event type name at this stage. @chomick suggested to leave the internal/external type def for later.

src/CaptainHook.Domain/Models/Subscriber.cs
tabman83: what we discussed with @chomick is that we'll have the endpoint as root and it'll be clear with the next PR which implements the repository and db models

src/CaptainHook.Domain/Models/Endpoint.cs
tabman83: for now, yes

chomick, line 7: I think you may need a separate model for authentication.

Also, as the models would also have some business logic in those classes, those properties may actually be readonly and will be modified via certain methods to capture any business logic errors etc.

I am ok if we keep them as normal POCO for now but the intention is to have the as proper models.Maybe we should start introducing that functionality gradually for things that we actually use?

src/CaptainHook.Domain/Models/Subscriber.cs
tabman83: The plural is definitely an overlook. It was meant to be Webhook Webhook. I don't like adding Model as it sounds as a repetition of the namespace but I am open to take it on board if you feel strong about it.

src/CaptainHook.Domain/Models/Endpoint.cs
```diff
@@ -0,0 +1,10 @@
+﻿namespace CaptainHook.Domain.Models
+{
+    public class Endpoint
```
chomick, line 3: You may have your models referencing their parents as well so we can navigate i the code through those objects.

src/CaptainHook.Domain/Models/Endpoint.cs
dawidkeshop: I think at this stage it's hard to predict what business logic is going to be there, so instead of blind guessing it's better to keep it simple as for now and improve when we need.

src/CaptainHook.Domain/Models/Authentication.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System.Collections;
+
+namespace CaptainHook.Domain.Models
+{
+    public class Authentication
+    {
+        public string ClientId { get; set; }
+        public string Secret { get; set; }
```
mszymeshop, line 8: We are not going to have the secret, right?

src/CaptainHook.Domain/Models/Event.cs
mszymeshop: in that case you don't need both Name and Type, as they are the same thing when it comes to events.

src/CaptainHook.Domain/Models/Authentication.cs
tabman83: I just want to have something in place for now and work out the details later.

PR #98, opened 2020-06-15
#71416 Introduce Cosmos DB repository and models

tabman83:


src/CaptainHook.Domain/Models/Subscriber.cs
```diff
@@ -4,8 +4,8 @@ public class Subscriber
     {
         public string Name { get; set; }
         public Event Event { get; set; }
-        public Webhooks Webhooks { get; set; }
-        public Webhooks Callbacks { get; set; }
-        public Webhooks Dlq { get; set; }
+        public Webhooks Webhooks { get; set; } = new Webhooks();
```
chomick, line 7: I still has a feeling we should make all props read-only, have a ctor's and start introducing methods to manipulate the data as we need it.
For now, the only place you need to be able to populate the data in there is in the repo.

src/CaptainHook.Domain/Models/Webhooks.cs
```diff
@@ -13,7 +12,7 @@ public class Webhooks
 
         public IList<Endpoint> Endpoints => _endpoints;
```
chomick, line 13: Any reason this is not just an enumerable?

src/CaptainHook.Domain/Models/Webhooks.cs
```diff
@@ -13,7 +12,7 @@ public class Webhooks
...
 
-        public void Add(Endpoint endpoint)
+        public void AddEndpoint(Endpoint endpoint)
```
chomick, line 15: I suggest we skip the methods for now, until we have needs for those - we will be doing unit test etc for them so we want to keep it to min., as needed.

src/CaptainHook.Repository/EndpointRepository.cs
```diff
@@ -0,0 +1,157 @@
...
+using System.Threading.Tasks;
+using DbEndpoint = CaptainHook.Repository.Models.Endpoint;
+using DomainEndpoint = CaptainHook.Domain.Models.Endpoint;
+using DomainAuthentication = CaptainHook.Domain.Models.Authentication;
+using CaptainHook.Repository.QueryBuilders;
+using System.Linq;
+
+namespace CaptainHook.Repository
+{
+    public class EndpointRepository : IEndpointRepository
```
chomick, line 15: Maybe 'endpoint' is not the best name for the repo as you cover most/all models?

src/CaptainHook.Repository/IEndpointRepository.cs
```diff
@@ -0,0 +1,8 @@
+﻿using System;
+
+namespace CaptainHook.Repository
+{
+    public class IEndpointRepository
```
chomick, line 5: When you populate the interface, please do the proper comments.
Also, the repository interface has to be defined in the Domain layer.

src/CaptainHook.Repository/EndpointRepository.cs
```diff
@@ -0,0 +1,157 @@
...
+
+            if (string.IsNullOrWhiteSpace(subscriberName))
+            {
+                throw new ArgumentNullException(nameof(subscriberName));
+            }
+
+            return GetSubscriberInternalAsync(eventName, subscriberName);
+        }
+
+        public Task<IEnumerable<Subscriber>> GetSubscribersAsync(string eventName)
```
chomick, line 47: Small thing - usually, the plural form is too difficult to distinguish from singular -maybe consider naming it more like GetSubscribersList, GetEventSubscribers, etc

src/CaptainHook.Repository/EndpointRepository.cs
```diff
@@ -0,0 +1,157 @@
...
+
+        public EndpointRepository(ICosmosDbRepository cosmosDbRepository, IEndpointQueryBuilder endpointQueryBuilder)
+        {
+            _cosmosDbRepository = cosmosDbRepository ?? throw new ArgumentNullException(nameof(cosmosDbRepository));
+            _endpointQueryBuilder = endpointQueryBuilder ?? throw new ArgumentNullException(nameof(endpointQueryBuilder));
+
+            _cosmosDbRepository.UseCollection(CollectionName);
+        }
+
+        public Task<Subscriber> GetSubscriberAsync(string eventName, string subscriberName)
```
chomick, line 32: Are you going to load entire data for the subscriber here (all webhooks, etc)?

src/CaptainHook.Repository/EndpointRepository.cs
```diff
@@ -0,0 +1,157 @@
...
+                        subscriber.Webhooks.Endpoints.Add(domainEndpoint);
+                        break;
+                    case WebhookType.Callback:
+                        subscriber.Webhooks.Endpoints.Add(domainEndpoint);
+                        break;
+                    case WebhookType.Dlq:
+                        subscriber.Webhooks.Endpoints.Add(domainEndpoint);
+                        break;
+                    default:
+                        throw new ArgumentException("Invalid webhook type for endpoint");
```
chomick, line 117: Think if we want to throw exceptions or simply return less complete records.
If data is invalid, there is nothing user/caller can do about it. They can't react to this exception either.

Also, that exception would be part of contract definition - if you plan to pass any exceptions beyond the repo, they should be part of the domain (but I don't believe this case qualifies)

src/CaptainHook.Repository/Models/Endpoint.cs
```diff
@@ -0,0 +1,27 @@
+﻿using Newtonsoft.Json;
+using System;
+
+namespace CaptainHook.Repository.Models
+{
+    public class Endpoint
+    {
+        public const string Type = "Endpoint";
+        public static string GetPartitionKey(string eventName, string subscriberName) => $"{eventName}-{subscriberName}";
+        public static string GetDocumentId() => Guid.NewGuid().ToString();
```
chomick, line 10: Should you not construct the Id from some data, i.e. event + subscriber + endpoint-selector (or 'default')
You will need to be able to find the endpoint document based on those three things so it makes sense to use them as the key.
I presume subscriber does not contain the event name.

src/CaptainHook.Repository/Models/Endpoint.cs
```diff
@@ -0,0 +1,27 @@
...
+
+        [JsonProperty("id")]
+        public string Id => GetDocumentId();
+
+        [JsonProperty("pk")]
+        public string Pk => GetPartitionKey(EventName, SubscriberName);
+
+        public string Uri { get; set; }
```
chomick, line 18: Do we want to keep the naming convention with fields starting with lower case in DB?

src/CaptainHook.Repository/Models/Endpoint.cs
```diff
@@ -0,0 +1,27 @@
...
+        public string Selector { get; set; }
+        public Authentication Authentication { get; set; }
+        public string HttpVerb { get; set; }
+        public string WebhookSelector { get; set; }
+        public string SubscriberName { get; set; }
+        public WebhookType WebhookType { get; set; }
```
chomick, line 24: I would ensure that gets serialised as a string

src/CaptainHook.Repository/QueryBuilders/EndpointQueryBuilder.cs
```diff
@@ -0,0 +1,22 @@
+﻿using Eshopworld.Data.CosmosDb;
+using Microsoft.Azure.Cosmos;
+
+namespace CaptainHook.Repository.QueryBuilders
+{
+    public class EndpointQueryBuilder: IEndpointQueryBuilder
+    {
+        public CosmosQuery BuildSelectSubscriberEndpoints(string eventName, string subscriberName)
+        {
+            var partitionKey = $"{eventName}-{subscriberName}";
```
chomick, line 10: You should use the partition key static method to construct it - it should be defined only in one place and in it is the model itself?

src/CaptainHook.Repository/QueryBuilders/EndpointQueryBuilder.cs
```diff
@@ -0,0 +1,22 @@
...
+            var query = new QueryDefinition("select * from c");
```
chomick, line 11: Don't you need to restrict the query?

src/CaptainHook.Repository/QueryBuilders/EndpointQueryBuilder.cs
```diff
@@ -0,0 +1,22 @@
...
+            return new CosmosQuery(query, partitionKey);
+        }
+
+        public CosmosQuery BuildSelectSubscribersEndpoints(string eventName)
+        {
+            var query = new QueryDefinition("select * from c where name = @name")
```
chomick, line 17: 'name' does not exist?

src/CaptainHook.Repository/QueryBuilders/IEndpointQueryBuilder.cs
```diff
@@ -0,0 +1,11 @@
+﻿using Eshopworld.Data.CosmosDb;
+
+namespace CaptainHook.Repository.QueryBuilders
+{
+    public interface IEndpointQueryBuilder
+    {
+        CosmosQuery BuildSelectSubscriberEndpoints(string eventName, string subscriberName);
```
chomick, line 7: Please add description what the methods are doing, e.g.
Creates SQL query to retrieve all endpoints for a given subscriber name and event name

src/CaptainHook.Repository/QueryBuilders/EndpointQueryBuilder.cs
chomick: You are querying endpoint document type here so you should restrict your query to the type as well.

src/CaptainHook.Repository/EndpointRepository.cs
dawidkeshop: If I understand correctly our domain, one event can have multiple subscribers, and subscribers can have webhooks, callbacks and dlqs (along with transformations and authentication details). This makes me think that `Event` is the root, so repository should allow to retrieve and store `Events`.

src/CaptainHook.Repository/EndpointRepository.cs
dawidkeshop: We can potentially return some object containing properly loaded Subscribers along with some data about what wasn't read properly. Or just log such errors somewhere.

src/CaptainHook.Repository/EndpointRepository.cs
tabman83: Yes this is supposed to be used by the api later

src/CaptainHook.Domain/Models/AuthenticationModel.cs
```diff
@@ -0,0 +1,44 @@
+﻿using System.Collections;
+
+namespace CaptainHook.Domain.Models
```
dawidkeshop, line 3: I think it's quite obvious that the `Domain` project contains the Domain Model, it can't contain the real domain anyway. So I would rather name this folder/namespace as `Entities` because looks that it contains entities.

Also I don't think that 'Model` suffixes for every entity provides any value here.

src/CaptainHook.Repository/EventRepository.cs
```diff
@@ -0,0 +1,168 @@
+﻿using CaptainHook.Domain.Models;
+using CaptainHook.Repository.Models;
+using Eshopworld.Data.CosmosDb;
+using System;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+using CaptainHook.Repository.QueryBuilders;
+using System.Linq;
+
+namespace CaptainHook.Repository
```
dawidkeshop, line 10: I think the name `Repository` for whole project is a bit misleading. I would rather follow common practices and name it `Persistence` or `Storage`.

src/CaptainHook.Domain/Models/SubscriberModel.cs
```diff
@@ -0,0 +1,48 @@
+﻿namespace CaptainHook.Domain.Models
+{
+    /// <summary>
+    /// Subscriber model
+    /// </summary>
+    public class SubscriberModel
```
dawidkeshop, line 6: Hmm, didn't we agree with @chomick  to use the suffix `Entity` and namespace `Entities` for all classes in this folder?

src/Tests/CaptainHook.Storage.Cosmos.Tests/SubscriberRepositoryTests.cs
```diff
@@ -0,0 +1,147 @@
...
+using Xunit;
+
+namespace CaptainHook.Repository.Tests
+{
+    public class SubscriberRepositoryTests
+    {
+        private readonly Mock<ICosmosDbRepository> cosmosDbRepositoryMock;
+        private readonly Mock<ISubscriberQueryBuilder> queryBuilderMock;
+
+        private readonly SubscriberRepository repository;
```
dawidkeshop, line 23: I think we should follow the `_` prefix conventions for fields also here.

src/Tests/CaptainHook.Storage.Cosmos.Tests/SubscriberRepositoryTests.cs
```diff
@@ -0,0 +1,147 @@
...
+using FluentAssertions.Execution;
+using Moq;
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection.Metadata;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace CaptainHook.Repository.Tests
```
dawidkeshop, line 16: Namespace doesn't correspond to the project name.

src/Tests/CaptainHook.Storage.Cosmos.Tests/SubscriberRepositoryTests.cs
```diff
@@ -0,0 +1,147 @@
...
+            cosmosDbRepositoryMock = new Mock<ICosmosDbRepository>();
+            queryBuilderMock = new Mock<ISubscriberQueryBuilder>();
+
+            repository = new SubscriberRepository(cosmosDbRepositoryMock.Object, queryBuilderMock.Object);
+        }
+
+        [Fact, IsUnit]
+        public async Task EmptyEventNameThrowsException()
+        {
+            // Arrange
```
dawidkeshop, line 36: Was this `//Arrange //Act //Assert` convention recently introduced to the project? I don't remember using this couple of weeks ago.

src/CaptainHook.Domain/Models/SubscriberModel.cs
tabman83: I honestly don't remember this specifically, but ok to change it!

src/Tests/CaptainHook.Storage.Cosmos.Tests/SubscriberRepositoryTests.cs
tabman83: Not specifically I've seen that as a generally adopted convention for testing which is also used in some esw places.

tabman83:
I have addressed @dawidkeshop comments with the latest commit.

src/Tests/CaptainHook.Storage.Cosmos.Tests/SubscriberRepositoryTests.cs
```diff
@@ -0,0 +1,147 @@
...
+            var expectedEndpointEntity = new EndpointEntity(sampleDocument.Uri, expectedAuthenticationEntity, sampleDocument.HttpVerb, sampleDocument.EndpointSelector);
+            expectedSubscriberEntity.AddWebhookEndpoint(expectedEndpointEntity);
+
+            // Act
+            var result = await repository.GetSubscribersListAsync(eventName);
+
+            // Assert
+            using (new AssertionScope())
+            {
+                result.Should().HaveCount(1);
```
mszymeshop, line 139: nit: I think you can simplify this into a single assertion

src/CaptainHook.Domain/Entities/WebhooksEntity.cs
```diff
@@ -0,0 +1,36 @@
...
+        /// Webhook endpoints
+        /// </summary>
+        public IEnumerable<EndpointEntity> Endpoints => _endpoints;
+
+        public WebhooksEntity(string selectionRule): this(selectionRule, null) { }
+
+        public WebhooksEntity(string selectionRule, IEnumerable<EndpointEntity> endpoints)
+        {
+            SelectionRule = selectionRule;
+            _endpoints = endpoints?.ToList() ?? new List<EndpointEntity>();
```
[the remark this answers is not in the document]
tabman83: Will change.

src/CaptainHook.Storage.Cosmos/SubscriberRepository.cs
```diff
@@ -0,0 +1,99 @@
...
+        {
+            if (string.IsNullOrWhiteSpace(eventName))
+            {
+                throw new ArgumentNullException(nameof(eventName));
+            }
+
+            return GetSubscribersListInternalAsync(eventName);
+        }
+
+        #region Private methods
```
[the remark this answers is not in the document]
tabman83: I just wanted to tidy up a little bit since we are having internal versions of each method for input validation

src/CaptainHook.Storage.Cosmos/SubscriberRepository.cs
```diff
@@ -0,0 +1,99 @@
...
+
+        private AuthenticationEntity Map(AuthenticationData authentication)
+        {
+            var secretStore = new SecretStoreEntity(authentication.KeyVaultName, authentication.SecretName);
+            return new AuthenticationEntity(authentication.ClientId, secretStore, authentication.Uri, authentication.Type, authentication.Scopes);
+        }
+
+        private SubscriberEntity Map(IEnumerable<EndpointDocument> endpoints)
+        {
+            var subscriberName = endpoints.First().SubscriberName;
```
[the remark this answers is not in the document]
tabman83: By the way our endpoint documents are defined, yes. That map method won't even be called if no endpoints are returned by the query on line 55.

src/Tests/CaptainHook.Storage.Cosmos.Tests/SubscriberRepositoryTests.cs
```diff
@@ -0,0 +1,147 @@
...
+        public SubscriberRepositoryTests()
+        {
+            _cosmosDbRepositoryMock = new Mock<ICosmosDbRepository>();
+            _queryBuilderMock = new Mock<ISubscriberQueryBuilder>();
+
+            repository = new SubscriberRepository(_cosmosDbRepositoryMock.Object, _queryBuilderMock.Object);
+        }
+
+        [Fact, IsUnit]
+        public async Task EmptyEventNameThrowsException()
```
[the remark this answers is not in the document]
tabman83: Totally makes sense thanks!

src/Tests/CaptainHook.Storage.Cosmos.Tests/SubscriberRepositoryTests.cs
tabman83: Done thanks!

src/CaptainHook.DirectorService/Infrastructure/ConfigurationMerger.cs
```diff
@@ -54,39 +54,39 @@ private SubscriberConfiguration MapBasicSubscriberData(Subscriber cosmosModel)
...
-            return new WebhookConfig()
-            {
-                Name = cosmosModel.Name,
-                Uri = endpoint.Uri,
-                HttpVerb = endpoint.HttpVerb,
-                AuthenticationConfig = MapAuthentication(endpoint.Authentication),
-                EventType = cosmosModel.Event.Type,
-            };
-        }
+        //    return new WebhookConfig()
```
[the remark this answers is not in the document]
tabman83: Not dealing with callbacks for now

src/CaptainHook.Domain/Entities/SecretStoreEntity.cs
```diff
@@ -0,0 +1,26 @@
...
+        /// Name of the keyvault which holds the secret
+        /// </summary>
+        public string KeyVaultName { get; }
+
+        /// <summary>
+        /// Name of the secret key which holds the actual secret
+        /// </summary>
+        public string SecretName { get; }
+        
+        public SecretStoreEntity(): this(null, null) { }
```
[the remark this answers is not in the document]
tabman83: I think it's not needed at this stage, so removing.

PR #99, opened 2020-06-16
71421 merge subscriber models from KV and Cosmos

dawidkeshop:


src/CaptainHook.DirectorService/Infrastructure/ConfigurationMerger.cs
```diff
@@ -0,0 +1,96 @@
...
+using CaptainHook.Common.Configuration;
+using CaptainHook.Domain.Models;
+
+namespace CaptainHook.DirectorService.Infrastructure
+{
+    public class ConfigurationMerger
+    {
+        public IEnumerable<SubscriberConfiguration> Merge(IEnumerable<SubscriberConfiguration> kvModels, IEnumerable<Subscriber> cosmosModels)
+        {
+            var onlyInKv = kvModels.Where(k => !cosmosModels.Any(c => k.Name == c.Event.Name && k.SubscriberName == c.Name));
```
chomick, line 15: We have to ensure that this criteria for identifying both entries is valid and consistent.
In particular, I am not sure how we are going to name/identify subscribers in Cosmos.

src/CaptainHook.DirectorService/Infrastructure/ConfigurationMerger.cs
```diff
@@ -0,0 +1,96 @@
...
+
+        private SubscriberConfiguration MapBasicSubscriber(Subscriber cosmosModel)
+        {
+            var config = new SubscriberConfiguration
+            {
+                Name = cosmosModel.Event.Name,
+                SubscriberName = cosmosModel.Name,
+                Uri = cosmosModel.Webhooks.Endpoints.First().Uri,
+                AuthenticationConfig = MapAuthentication(cosmosModel.Webhooks.Endpoints.FirstOrDefault()?.Authentication),
+                Callback = MapCallback(cosmosModel),
```
chomick, line 44: We don't have to support 'callbacks' just yet.

src/CaptainHook.DirectorService/Infrastructure/ConfigurationMerger.cs
```diff
@@ -0,0 +1,96 @@
...
+            return result;
+        }
+
+        private IEnumerable<SubscriberConfiguration> MapSubscriber(Subscriber cosmosModel)
+        {
+            var result = new List<SubscriberConfiguration>();
+            var config = MapBasicSubscriber(cosmosModel);
+            result.Add(config);
+
+            var dlq = cosmosModel?.Dlq?.Endpoints.FirstOrDefault();
```
chomick, line 27: We don't have to support DLQ's just yet

src/CaptainHook.DirectorService/Infrastructure/ConfigurationMerger.cs
```diff
@@ -0,0 +1,96 @@
...
+        private AuthenticationConfig MapAuthentication(Authentication cosmosAuthentication)
+        {
+            if (cosmosAuthentication == null)
+                return null;
+
+            return new OidcAuthenticationConfig
+            {
+                Type = AuthenticationType.OIDC,
+                ClientId = cosmosAuthentication.ClientId,
+                ClientSecret = cosmosAuthentication.Secret,
```
chomick, line 90: Is the secret going to be populated before we get here (on cosmos side)?

src/CaptainHook.DirectorService/Infrastructure/ConfigurationMerger.cs
mszymeshop: and when we do we might have multiple endpoints there as well

src/CaptainHook.DirectorService/Infrastructure/ConfigurationMerger.cs
```diff
@@ -0,0 +1,96 @@
...
+            var dlq = cosmosModel?.Dlq?.Endpoints.FirstOrDefault();
+            if (dlq != null)
+            {
+                result.Add(MapDlq(cosmosModel));
+            }
+
+            return result;
+        }
+
+        private SubscriberConfiguration MapBasicSubscriber(Subscriber cosmosModel)
```
mszymeshop, line 45: What is a `BasicSubscriber`?

src/CaptainHook.DirectorService/Infrastructure/ConfigurationMerger.cs
```diff
@@ -0,0 +1,96 @@
...
+        {
+            var config = new SubscriberConfiguration
+            {
+                Name = cosmosModel.Event.Name,
+                SubscriberName = cosmosModel.Name,
+                Uri = cosmosModel.Webhooks.Endpoints.First().Uri,
```
mszymeshop, line 51: What if there is no Endpoint?

src/Tests/CaptainHook.Tests/Director/ConfigurationMergerTests.cs
```diff
@@ -0,0 +1,167 @@
...
+                    .WithWebhook("https://cosmos.eshopworld.com/testevent/", "POST")
+                    .WithDlq("https://cosmos.eshopworld.com/dlq/", "PUT")
+                    .Create(),
+            };
+
+            var result = new ConfigurationMerger().Merge(new List<SubscriberConfiguration>(), cosmosSubscribers);
+
+            using (new AssertionScope())
+            {
+                result.Should().HaveCount(2);
```
mszymeshop, line 160: The way I like to approach these kinds of cases is to create the expected collection and compare the result against it. If you want to focus on specific properties only you can still do this with the config. Just a suggestion though, your call.

src/CaptainHook.DirectorService/Infrastructure/ConfigurationMerger.cs
```diff
@@ -0,0 +1,96 @@
...
+namespace CaptainHook.DirectorService.Infrastructure
+{
+    public class ConfigurationMerger
+    {
+        public IEnumerable<SubscriberConfiguration> Merge(IEnumerable<SubscriberConfiguration> kvModels, IEnumerable<Subscriber> cosmosModels)
+        {
+            var onlyInKv = kvModels.Where(k => !cosmosModels.Any(c => k.Name == c.Event.Name && k.SubscriberName == c.Name));
+            var fromCosmos = cosmosModels.SelectMany(MapSubscriber);
+            var result = onlyInKv.Union(fromCosmos);
+            return result;
```
mszymeshop, line 27: Personally I would materialise this collection as read-only here and return it like that as it is going to be heavily used from now on. This way you're potentially risking multiple materialisations in the client code.

src/CaptainHook.DirectorService/Infrastructure/ConfigurationMerger.cs
```diff
@@ -0,0 +1,96 @@
...
+        }
+
+        private IEnumerable<SubscriberConfiguration> MapSubscriber(Subscriber cosmosModel)
+        {
+            var result = new List<SubscriberConfiguration>();
+            var config = MapBasicSubscriber(cosmosModel);
+            result.Add(config);
```
mszymeshop, line 34: consider yield return for simplified code

src/CaptainHook.DirectorService/Infrastructure/ConfigurationMerger.cs
```diff
@@ -0,0 +1,96 @@
...
+            }
+
+            return result;
+        }
+
+        private SubscriberConfiguration MapBasicSubscriber(Subscriber cosmosModel)
+        {
+            var config = new SubscriberConfiguration
+            {
+                Name = cosmosModel.Event.Name,
```
mszymeshop, line 49: Should we invest in more validation? What if cosmosModel is null, what if Event is empty, what if Webhooks are not defined etc?

src/CaptainHook.DirectorService/Infrastructure/ConfigurationMerger.cs
```diff
@@ -0,0 +1,96 @@
...
+                DLQMode = SubscriberDlqMode.WebHookMode,
+            };
+        }
+
+        private AuthenticationConfig MapAuthentication(Authentication cosmosAuthentication)
+        {
+            if (cosmosAuthentication == null)
+                return null;
+
+            return new OidcAuthenticationConfig
```
mszymeshop, line 95: Oidc might be the only one supported now, but should we check if types match first?

src/Tests/CaptainHook.Tests/Builders/SubscriberBuilder.cs
```diff
@@ -0,0 +1,63 @@
+﻿using CaptainHook.Domain.Models;
+
+namespace CaptainHook.Tests.Builders
+{
+    internal class SubscriberBuilder
```
mszymeshop, line 5: :100:

src/CaptainHook.DirectorService/Infrastructure/ConfigurationMerger.cs
dawidkeshop: And in case of multiple endpoints we will map to `SubscriberConifiguration.WebhookRequestRules[0].Routes` ?

src/CaptainHook.DirectorService/Infrastructure/ConfigurationMerger.cs
dawidkeshop: Approach confirmed with the team, comparison improved to ignore casing.

src/CaptainHook.DirectorService/Infrastructure/ConfigurationMerger.cs
dawidkeshop: It's not a basic subscriber, it's more like its basic data. I'll rename the method. :)

src/CaptainHook.DirectorService/Infrastructure/ConfigurationMerger.cs
dawidkeshop: Is this valid scenario? I thought that this is obligatory to have at least one endpoint here.

src/Tests/CaptainHook.Tests/Director/ConfigurationMergerTests.cs
dawidkeshop: This test is skipped as for now, I think better to improve it when become required.

src/CaptainHook.DirectorService/Infrastructure/ConfigurationMerger.cs
dawidkeshop: Good idea, thanks!

src/CaptainHook.DirectorService/Infrastructure/ConfigurationMerger.cs
dawidkeshop: Good idea, thanks!

src/CaptainHook.DirectorService/Infrastructure/ConfigurationMerger.cs
dawidkeshop: I think we shouldn't create an invalid domain model in a first place. As for me it's the repository responsibility to read and build model properly.

If we really want additional checks at that level, it brings up new question - how to deal in case of invalid data? Exceptions, errors, ignore, log, etc.?

src/CaptainHook.DirectorService/Infrastructure/ConfigurationMerger.cs
dawidkeshop: OK, but how I can infer the type based on `Authenctication` class?

src/CaptainHook.DirectorService/Infrastructure/ConfigurationMerger.cs
chomick: Agree that we should not need to repeat the validation logic here. Happy to keep it as is

src/CaptainHook.DirectorService/Infrastructure/ConfigurationMerger.cs
chomick: I think we should be fine - I would maybe get the endpoint first rather then fetch it twice but we probably do not expect it to be null.
I would add try-catch for all possible mapping issues outside somewhere and log it, skipping the subscriber in such case

src/CaptainHook.DirectorService/Infrastructure/ConfigurationMerger.cs
chomick: Not sure if we have definition/clarity on the auth-models so happy to assume only OAuth for now

src/CaptainHook.DirectorService/Infrastructure/ConfigurationMerger.cs
chomick: @tabman83 - will let it pass for now but we need to ensure this is fetched from KV at some stage

tabman83: I have changed it in the other PR already

PR #104, opened 2020-06-19
DTOs defined

dawidkeshop:


src/CaptainHook.Api/Dtos/AuthenticationDto.cs
```diff
@@ -0,0 +1,13 @@
+﻿using System.Collections.Generic;
```
chomick, line 1: I would suggest to have the DTOs in a separate 'contract' project but can be updated in subsequent PR, if agreed to move

src/CaptainHook.Api/Dtos/AuthenticationDto.cs
```diff
@@ -0,0 +1,13 @@
...
+
+namespace CaptainHook.Api.Dtos
+{
+    public class AuthenticationDto
```
chomick, line 5: I was thinking - what if we simply exposed the existing KV models (temporarily) so we can see all the data properly?
The problem with original approach is that we need to map from KV to new DTOs and we will lose lots of info which possibly is very useful to have, even if format is temporary and there is no clear separation between API contract and internal models for now.

Just a thought.
The ticket did allow for bits to be skipped but I wonder if it makes sense to do it that way, considering where we are now we the knowledge/work/plans.

Alternatively, maybe we can expose this GET +additional GET (or section within subscriber) where KV settings are out there in raw (minus secrets)?

The benefit would be that we can see the actual settings that are really used by CH.

src/CaptainHook.Api/Dtos/AuthenticationDto.cs
dawidkeshop: OK, makes sense.

src/CaptainHook.Api/Dtos/AuthenticationDto.cs
dawidkeshop: We can do anything but what are the actual requirements there? If we need another secret API endpoints for some test purposes, we should define the story in a different way.

PR #106, opened 2020-06-22
71422 extend config load for cosmos

tabman83:


src/CaptainHook.DirectorService/DirectorService.cs
```diff
@@ -43,17 +44,25 @@ public class DirectorService : StatefulService, IDirectorServiceRemoting
...
             _readerServicesManager = readerServicesManager;
         }
 
+        private async Task LoadConfigurationAsync()
+        {
+            var result = await _subscriberConfigurationLoader.LoadAsync();
+
+            _webhookConfigurations = result.Item1;            
+            _subscriberConfigurations = result.Item2.ToDictionary(x => SubscriberConfiguration.Key(x.EventType, x.SubscriberName));
+
```
mszymeshop, line 62: unnecessary blank lines

src/CaptainHook.DirectorService/DirectorService.cs
```diff
@@ -43,17 +44,25 @@ public class DirectorService : StatefulService, IDirectorServiceRemoting
...
+            _subscriberConfigurationLoader = subscriberConfigurationLoader;
             _readerServicesManager = readerServicesManager;
         }
 
+        private async Task LoadConfigurationAsync()
+        {
+            var result = await _subscriberConfigurationLoader.LoadAsync();
+
+            _webhookConfigurations = result.Item1;            
+            _subscriberConfigurations = result.Item2.ToDictionary(x => SubscriberConfiguration.Key(x.EventType, x.SubscriberName));
```
mszymeshop, line 61: you can use tuple language support to have named values or create a compound object to store this. Item1 and Item2 do not say much about the content.

src/CaptainHook.DirectorService/Infrastructure/SubscriberConfigurationLoader.cs
```diff
@@ -0,0 +1,32 @@
...
+    {
+        private readonly ISubscriberRepository _subscriberRepository;
+
+        public SubscriberConfigurationLoader(ISubscriberRepository subscriberRepository)
+        {
+            _subscriberRepository = subscriberRepository;
+        }
+
+        public async Task<(IList<WebhookConfig>, IList<SubscriberConfiguration>)> LoadAsync()
+        {
```
mszymeshop, line 18: looks nice but why blank lines after every code line?

src/CaptainHook.DirectorService/Infrastructure/SubscriberConfigurationLoader.cs
```diff
@@ -0,0 +1,32 @@
...
+            var configurationMerger = new ConfigurationMerger();
```
mszymeshop, line 19: would unit tests be easier if we had `ConfigurationMerger` injected instead?

src/CaptainHook.DirectorService/DirectorService.cs
```diff
@@ -43,17 +44,25 @@ public class DirectorService : StatefulService, IDirectorServiceRemoting
...
-            _webhookConfigurations = webhookConfigurations;
+            _subscriberConfigurationLoader = subscriberConfigurationLoader;
             _readerServicesManager = readerServicesManager;
         }
 
+        private async Task LoadConfigurationAsync()
+        {
+            var result = await _subscriberConfigurationLoader.LoadAsync();
+
+            _webhookConfigurations = result.Item1;
```
dawidkeshop, line 60: `Item1` and `Item2` doesn't seem very descriptive.

src/CaptainHook.Storage.Cosmos/SubscriberRepository.cs
```diff
@@ -37,6 +37,16 @@ public SubscriberRepository(ICosmosDbRepository cosmosDbRepository, ISubscriberQ
             _cosmosDbRepository.UseCollection(CollectionName);
         }
 
+        public async Task<IEnumerable<SubscriberEntity>> GetAllSubscribersAsync()
+        {
+            var query = _endpointQueryBuilder.BuildSelectAllSubscribersEndpoints();
+            var endpoints = await _cosmosDbRepository.QueryAsync<EndpointDocument>(query);
+
+            return endpoints
+                .GroupBy(x => x.SubscriberName)
```
mszymeshop, line 46: Why is `GroupBy` needed here?

src/CaptainHook.DirectorService/DirectorService.cs
dawidkeshop, line 62: I think we can remove empty lines at the end of method.

src/CaptainHook.DirectorService/Program.cs
```diff
@@ -71,6 +72,10 @@ private static async Task Main()
                 builder.RegisterType<ReaderServicesManager>()
                     .As<IReaderServicesManager>()
                     .SingleInstance();
+
+                builder
+                    .RegisterType<ConfigurationMerger>()
```
dawidkeshop, line 77: This registration is not needed.

src/CaptainHook.DirectorService/DirectorService.cs
chomick: Or simply use correct type, especially for an interface/contract.

src/CaptainHook.DirectorService/Infrastructure/SubscriberConfigurationLoader.cs
tabman83: There is not much to be tested here since all the magic is happening inside the configurationmerger itself. I believe this is the reason why it does not have an interface in the first place?

src/CaptainHook.DirectorService/Infrastructure/SubscriberConfigurationLoader.cs
tabman83: Addressed this with latest commit.

src/CaptainHook.DirectorService/DirectorService.cs
tabman83: Addressed this with latest commit.

src/CaptainHook.DirectorService/DirectorService.cs
tabman83: Addressed this with latest commit.

src/CaptainHook.DirectorService/DirectorService.cs
tabman83: Addressed this with latest commit.

src/CaptainHook.DirectorService/Infrastructure/SubscriberConfigurationLoader.cs
mszymeshop: You seem to have `ConfigurationMerger` registered with IoC though. So something is not adding up.

src/CaptainHook.DirectorService/Program.cs
tabman83: Addressed with latest commit

src/CaptainHook.DirectorService/Infrastructure/SubscriberConfigurationLoader.cs
```diff
@@ -0,0 +1,32 @@
...
+
+            var configuration = Configuration.Load();
+
+            var subscribersFromKV = configuration.SubscriberConfigurations;
+
+            var subscribersFromCosmos = await _subscriberRepository.GetAllSubscribersAsync();
+
+            var merged = configurationMerger.Merge(subscribersFromKV.Values, subscribersFromCosmos);
+
+            return (configuration.WebhookConfigurations, merged);
```
chomick, line 25: Why do we return webhooks configuration unmerged?

src/CaptainHook.DirectorService/Infrastructure/SubscriberConfigurationLoader.cs
dawidkeshop: More dependencies makes unit tests harder, because test setup becomes more complicated. Also exposing implementation details tightly couples tests with the implementation which makes safe refactoring not possible.

src/CaptainHook.DirectorService/Infrastructure/SubscriberConfigurationLoader.cs
```diff
@@ -0,0 +1,32 @@
+﻿using CaptainHook.Common.Configuration;
+using CaptainHook.Domain.Repositories;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+
+namespace CaptainHook.DirectorService.Infrastructure
+{
+    public class SubscriberConfigurationLoader
```
chomick, line 8: Are you going to add unit tests for this class?

src/CaptainHook.Storage.Cosmos/SubscriberRepository.cs
tabman83: Because I am retrieving endpoints and mapping each one of them to the subscriber they belong

src/CaptainHook.DirectorService/Infrastructure/SubscriberConfigurationLoader.cs
dawidkeshop: Because these doesn't have to be merged, we need only fresh values to be passed to Handlers.

This technical debt confuses every time, maybe we should finally fix it? :)

src/CaptainHook.DirectorService/Infrastructure/SubscriberConfigurationLoader.cs
chomick: Agree

src/CaptainHook.DirectorService/Infrastructure/SubscriberConfigurationLoader.cs
```diff
@@ -0,0 +1,28 @@
...
+    {
+        private readonly ISubscriberRepository _subscriberRepository;
+
+        public SubscriberConfigurationLoader(ISubscriberRepository subscriberRepository)
+        {
+            _subscriberRepository = subscriberRepository;
+        }
+
+        public async Task<(IList<WebhookConfig>, IList<SubscriberConfiguration>)> LoadAsync()
```
chomick, line 17: I think you should define a structure (e.g. LoadResult) that encapsulates those two fields

src/CaptainHook.DirectorService/Infrastructure/SubscriberConfigurationLoader.cs
tabman83: This is not doing much apart from loading&merging for which we already have tests

src/CaptainHook.DirectorService/Infrastructure/SubscriberConfigurationLoader.cs
tabman83: I did not want to overcomplicate it at this stage since we are later gonna change it to only have subscribers

src/CaptainHook.DirectorService/Infrastructure/SubscriberConfigurationLoader.cs
tabman83: ConfigurationMerger in IoC was a mistake and has been removed

src/CaptainHook.DirectorService/DirectorService.cs
tabman83: Done already

src/CaptainHook.DirectorService/Infrastructure/SubscriberConfigurationLoader.cs
chomick: To verify you passed the correct data in-out and that you called correct methods with the correct params etc.
Unit test here would cover the functionality you have implemented here - I think we would need a strong reason NOT to add the tests and I can't see any?

src/CaptainHook.DirectorService/Infrastructure/SubscriberConfigurationLoader.cs
chomick: in that case I'm ok with that :)

src/CaptainHook.DirectorService/Infrastructure/SubscriberConfigurationLoader.cs
chomick: I am happy to look at the tests as a separate PR

src/CaptainHook.DirectorService/Infrastructure/SubscriberConfigurationLoader.cs
chomick: I think we should have an interface for the merger and use the interface only as a dependency.
Again, by default, we should use that approach unless there is a good reason NOT to do so.

tabman83: I will change that.

src/CaptainHook.DirectorService/Infrastructure/SubscriberConfigurationLoader.cs
```diff
@@ -0,0 +1,28 @@
+﻿using CaptainHook.Common.Configuration;
+using CaptainHook.Domain.Repositories;
+using System.Collections.Generic;
+using System.Threading.Tasks;
+
+namespace CaptainHook.DirectorService.Infrastructure
+{
+    public class SubscriberConfigurationLoader
```
chomick, line 8: I would also define an interface for this class and use it instead in the Director Service

src/CaptainHook.DirectorService/Infrastructure/ConfigurationMerger.cs
```diff
@@ -8,24 +8,23 @@
...
+    public class ConfigurationMerger : IConfigurationMerger
     {
         /// <summary>
         /// Merges subscribers loaded from Cosmos and from KeyVault. If particular subscriber is defined in both sources, the Cosmos version overrides the KeyVault version.
         /// </summary>
         /// <param name="subscribersFromKeyVault">Subscriber definitions loaded from KeyVault</param>
         /// <param name="subscribersFromCosmos">Subscriber models retrieved from Cosmos</param>
         /// <returns>List of all subscribers converted to KeyVault structure</returns>
-        public ReadOnlyCollection<SubscriberConfiguration> Merge(IEnumerable<SubscriberConfiguration> subscribersFromKeyVault, IEnumerable<SubscriberEntity> subscribersFromCosmos)
+        public IEnumerable<SubscriberConfiguration> Merge(IEnumerable<SubscriberConfiguration> subscribersFromKeyVault, IEnumerable<SubscriberEntity> subscribersFromCosmos)
```
dawidkeshop, line 19: Why to change this? If I remember correctly, in PR it was requested to not return `IEnumerable` here.

tabman83: We have IEnumerable all over and I thought it would make more sense for consistency. I don't honestly remember the conversation anymore, but happy to change it back if it makes sense.

src/CaptainHook.DirectorService/Infrastructure/Interfaces/IConfigurationMerger.cs
```diff
@@ -0,0 +1,11 @@
+﻿using CaptainHook.Common.Configuration;
+using CaptainHook.Domain.Entities;
+using System.Collections.Generic;
+
+namespace CaptainHook.DirectorService.Infrastructure.Interfaces
+{
+    public interface IConfigurationMerger
```
dawidkeshop, line 7: What's the actual value of having an interface for an internal utility class with only one implementation which is owned by our team?

tabman83: @chomick suggested it here https://github.com/eShopWorld/captain-hook/pull/106#discussion_r443568320

src/CaptainHook.DirectorService/Infrastructure/ConfigurationMerger.cs
dawidkeshop: I think @mszymeshop had some doubts about multiple materialization: https://github.com/eShopWorld/captain-hook/pull/99#discussion_r441417377

src/Tests/CaptainHook.Tests/Director/SubscriberConfigurationLoaderTests.cs
```diff
@@ -0,0 +1,36 @@
...
+            _configurationMergerMock = new Mock<IConfigurationMerger>();
+
+            _subscriberConfigurationLoader = new SubscriberConfigurationLoader(_subscriberRepositoryMock.Object, _configurationMergerMock.Object);
+        }
+
+        [Fact, IsLayer0]
+        public async Task MergeIsInvoked()
+        {
+            var result = await _subscriberConfigurationLoader.LoadAsync();
+            _configurationMergerMock.Verify(x => x.Merge(It.IsAny<IEnumerable<SubscriberConfiguration>>(), It.IsAny<IEnumerable<SubscriberEntity>>()));
```
mszymeshop, line 33: I don't think it brings any value (it is interaction-based and of high maintenance), but if you changed it be state-based you could verify what is the first value returned from, and the second (the tuple).

src/CaptainHook.DirectorService/Infrastructure/ConfigurationMerger.cs
tabman83: Missed that! Many thanks, will revert!

PR #107, opened 2020-06-22
71428 get subscribers endpoint

dawidkeshop:


src/CaptainHook.Api/Controllers/RetrieveSubscribersController.cs
```diff
@@ -0,0 +1,87 @@
...
+                return Ok(subscribers);
+            }
+            catch (Exception exception)
+            {
+                _bigBrother.Publish(exception);
+                return BadRequest();
+            }
+        }
+
+        private void HideCredentials(ICollection<SubscriberConfiguration> subscribers)
```
chomick, line 56: You probably can use IEnumerable<> here

src/CaptainHook.Api/Controllers/RetrieveSubscribersController.cs
```diff
@@ -0,0 +1,87 @@
...
+                }
+            }
+        }
+
+        private void HideWebhookCredentials(WebhookConfig configuration)
+        {
+            if (configuration == null)
+                return;
+
+            switch (configuration.AuthenticationConfig)
```
chomick, line 76: I would mask both fields regardless of the type set

src/CaptainHook.Api/Controllers/RetrieveSubscribersController.cs
```diff
@@ -0,0 +1,87 @@
...
+            _bigBrother = bigBrother;
+        }
+
+        /// <summary>
+        /// Retrieve all subscribers from current configuration
+        /// </summary>
+        /// <response code="200">Subscribers retrieved properly</response>
+        /// <response code="401">Unauthorized request</response>
+        [HttpGet]
+        [ProducesResponseType(StatusCodes.Status200OK)]
```
chomick, line 36: As this is temp endpoint, I am ok not to expose the actual type returned here.

src/CaptainHook.DirectorService/DirectorService.cs
```diff
@@ -119,6 +119,11 @@ public Task<RequestReloadConfigurationResult> RequestReloadConfigurationAsync()
             return Task.FromResult(RequestReloadConfigurationResult.ReloadInProgress);
         }
 
+        public Task<IDictionary<string, SubscriberConfiguration>> GetAllSubscribers()
```
tabman83, line 122: Shall we use the -async suffix here?

src/CaptainHook.Api/Controllers/RetrieveSubscribersController.cs
dawidkeshop: Right

src/CaptainHook.Api/Controllers/RetrieveSubscribersController.cs
dawidkeshop: The names of fields depends on the type, I'm masking ClientSecret for OidcAuthentication and Password in case of BasicAuthentication. Should I mask something else?

src/CaptainHook.DirectorService/DirectorService.cs
dawidkeshop: OK

src/CaptainHook.Api/Controllers/RetrieveSubscribersController.cs
tabman83: @dawidkeshop I think Darek means it's ok not to specify the return type is a list of subscribers so you should be keeping the 200 OK hint for swagger

dawidkeshop: Ah, OK now I understand.

chomick: yes - that was just a comment not suggestion to change - normally I would expect the type to be specified along with the 200 code here but if for some reason we do not want to do it, it is ok in this case IMO.

src/CaptainHook.Api/Controllers/RetrieveSubscribersController.cs
chomick: I would mask them always, if they are set.
If for some reason the type is misconfigured and you have some KV entries for the other element, it is possible (?) that the secret may leak - just as an extra protection I would simply apply the masking for both fields as long as they are set.

dawidkeshop: We have a class hierarchy here, `AuthenticationConfig` is the base class, and it has two children: `BasicAuthenticationConfig` and `OidcAuthenticationConfig`. Each of these two contains different fields, so I have to cast to the actual type first and mask fields which exists in the actual type.

PR #108, opened 2020-06-22
#71454 - get secret value from key vault

mszymeshop:
### What
Introduce Secret Client in order to read cosmos db secret names and translate into actual secrets

### Why
We do not want to store secrets outside of Key Vault

### Disclaimer
Code has been unit tested but hasn't been manually tested yet as it needs to be merged with https://github.com/eShopWorld/captain-hook/pull/106 first.

### Proof of life
image


src/CaptainHook.Common/Configuration/KeyVault/ISecretManager.cs
```diff
@@ -0,0 +1,10 @@
+﻿using System.Threading.Tasks;
+using JetBrains.Annotations;
+
+namespace CaptainHook.Common.Configuration.KeyVault
+{
+    public interface ISecretManager
```
chomick, line 6: Maybe ISecretProvider? :)

src/CaptainHook.Common/Configuration/KeyVault/KeyVaultModule.cs
```diff
@@ -0,0 +1,21 @@
...
+{
+    public class KeyVaultModule: Module
+    {
+        protected override void Load(ContainerBuilder builder)
+        {
+            base.Load(builder);
+
+            builder.Register(context => new SecretClient(
+                new Uri(Environment.GetEnvironmentVariable(ConfigurationSettings.KeyVaultUriEnvVariable)),
+                new DefaultAzureCredential()))
```
chomick, line 16: How does it work? To run it locally, is it me who needs to have the permissions to the KV?
If running on env's, is the app running under some credentials/system user?

src/CaptainHook.Common/Configuration/KeyVault/KeyVaultSecretManager.cs
```diff
@@ -0,0 +1,45 @@
...
+            {
+                throw new ArgumentException("Secret name cannot be null or empty.", nameof(secretName));
+            }
+
+            try
+            {
+                var response = await _secretClient.GetSecretAsync(secretName);
+                return response?.Value?.Value;
+            }
+            catch (RequestFailedException ex)
```
chomick, line 33: Can we get KV throttling as well?

src/CaptainHook.DirectorService/Infrastructure/ConfigurationMerger.cs
```diff
@@ -91,11 +99,13 @@ private AuthenticationConfig MapAuthentication(AuthenticationEntity cosmosAuthen
             if (cosmosAuthentication == null)
                 return null;
 
+            var secretValue = _secretManager.GetSecretValueAsync(cosmosAuthentication.SecretStore.SecretName).Result;
```
chomick, line 102: Will we add caching later in the manager/provider?

chomick, line 102: Should we not call if the reference is not provided?

src/CaptainHook.EventReaderService/ServiceBusNamespaceExtensions.cs
```diff
@@ -79,8 +78,8 @@ public static async Task<ITopic> SetupTopic(string azureSubscriptionId, string s
                                    .WithLogLevel(HttpLoggingDelegatingHandler.Level.Basic)
                                    .WithCredentials(new AzureCredentials(tokenCredentials, tokenCredentials, string.Empty, AzureEnvironment.AzureGlobalCloud))
                                    .Build();
-
-            var sbNamespace = Azure.Authenticate(client, string.Empty)
+            
+            var sbNamespace = Microsoft.Azure.Management.Fluent.Azure.Authenticate(client, string.Empty)
```
chomick, line 82: Did you get some conflicts? Why no 'using'?

src/CaptainHook.DirectorService/Infrastructure/ConfigurationMerger.cs
mszymeshop: tbh I haven't touched on caching at all. Do we have this as part of another story or should it be implemented right now?

dawidkeshop, line 102: Is calling `.Result` a good approach? Maybe would be better to change surrounding method to be `async`?

mszymeshop: Async is definitely coming. My goal here was to validate if the algorithm and secret client are working before I change the entire `ConfigurationMerger` class.

src/CaptainHook.Common/Configuration/KeyVault/ISecretManager.cs
mszymeshop: done

src/CaptainHook.Common/Configuration/KeyVault/KeyVaultModule.cs
mszymeshop: I have now changed this a bit to reflect the same mechanism we use to load all KV for config. Please take a look. The differences are there because of v3 used in old code and v4 used in new. v4 has benefits of better abstraction and throttling support.

src/CaptainHook.Common/Configuration/KeyVault/KeyVaultSecretManager.cs
mszymeshop: Exponential approach to throttling is also supported now

src/CaptainHook.DirectorService/Infrastructure/ConfigurationMerger.cs
mszymeshop: I have updated this logic a bit to call `secretProvider` only when `name` is available

src/CaptainHook.EventReaderService/ServiceBusNamespaceExtensions.cs
mszymeshop: Ambiguous reference anyway even with using. The reason is I have added additional SDK libraries which make this call ambiguous.

src/CaptainHook.DirectorService/Infrastructure/ConfigurationMerger.cs
mszymeshop: Will close, but let me know if this should be addressed

mszymeshop: async done

src/CaptainHook.Common/Configuration/KeyVault/KeyVaultModule.cs
```diff
@@ -0,0 +1,50 @@
...
+                    Delay = TimeSpan.FromSeconds(2.0),
+                    MaxDelay = TimeSpan.FromSeconds(16.0),
+                    MaxRetries = 3,
+                    Mode = RetryMode.Exponential
+                }
+            };
+
+            builder.Register(context => new SecretClient(
+                new Uri(Environment.GetEnvironmentVariable(ConfigurationSettings.KeyVaultUriEnvVariable)),
+                new AzureServiceTokenCredential(),
```
chomick, line 31: Have we tested this code in CI (off branch) to see if that works?

src/CaptainHook.Common/Configuration/KeyVault/ISecretProvider.cs
```diff
@@ -0,0 +1,10 @@
+﻿using System.Threading.Tasks;
+using JetBrains.Annotations;
+
+namespace CaptainHook.Common.Configuration.KeyVault
+{
+    public interface ISecretProvider
+    {
+        Task<string> GetSecretValueAsync([NotNull] string secretName);
```
tabman83, line 8: Consider accepting a cancellation token as when we will be loading many secrets at once we might want to be able to support cancellation all the way down here from RunAsync()

src/CaptainHook.DirectorService/Infrastructure/ConfigurationMerger.cs
```diff
@@ -2,36 +2,58 @@
...
+        }
+
         /// <summary>
         /// Merges subscribers loaded from Cosmos and from KeyVault. If particular subscriber is defined in both sources, the Cosmos version overrides the KeyVault version.
         /// </summary>
         /// <param name="subscribersFromKeyVault">Subscriber definitions loaded from KeyVault</param>
         /// <param name="subscribersFromCosmos">Subscriber models retrieved from Cosmos</param>
         /// <returns>List of all subscribers converted to KeyVault structure</returns>
-        public ReadOnlyCollection<SubscriberConfiguration> Merge(IEnumerable<SubscriberConfiguration> subscribersFromKeyVault, IEnumerable<SubscriberEntity> subscribersFromCosmos)
+        public async Task<ReadOnlyCollection<SubscriberConfiguration>> MergeAsync(
```
tabman83, line 29: This may accept CancellationToken, see comment above

src/CaptainHook.Common/Configuration/KeyVault/KeyVaultModule.cs
```diff
@@ -0,0 +1,50 @@
...
+using System.Threading.Tasks;
+using Autofac;
+using Azure.Core;
+using Azure.Security.KeyVault.Secrets;
+using Microsoft.Azure.Services.AppAuthentication;
+using TokenCredential = Azure.Core.TokenCredential;
+
+namespace CaptainHook.Common.Configuration.KeyVault
+{
+    public class KeyVaultModule: Module
```
tabman83, line 12: I feel this name being a bit too much generic for something tied to secret loading. That's because we may need a proper keyvault module later to abstract main configuration loading

mszymeshop: I think the name is ok mostly because it creates an instance of the SecretClient, what's more it is ok to extend it in the future with further KV-specific registrations.

src/CaptainHook.Common/Configuration/KeyVault/ISecretProvider.cs
mszymeshop: Configuration Reload does not - currently - support Cancellation Token at all. It would be a significant change to introduce it. In other words, the way KeyVault secret loading works at the moment -> it loads everything. The new mechanism does the same with the change that a significantly smaller number of calls will be actually made -> only for provided ref auths. What's more we can invest in cache.

src/CaptainHook.DirectorService/Infrastructure/ConfigurationMerger.cs
mszymeshop: See above.

src/CaptainHook.Common/Configuration/KeyVault/KeyVaultModule.cs
mszymeshop: This has been tested and communication with KV is correct and working.

mszymeshop:
@tabman83 Nino, I am going to go ahead with this. If you want us to come back to the discussion regarding Cancellation Token please let me know in Slack.

PR #110, opened 2020-06-23
Use camel case for all the endpoint document properties

tabman83:


src/CaptainHook.Storage.Cosmos/Models/EndpointDocument.cs
```diff
@@ -10,7 +10,6 @@ internal class EndpointDocument
         public const string Type = "Endpoint";
         public static string GetPartitionKey(string eventName) => $"{Type}_{eventName}";
         
-        //change
         public static string GetDocumentId(string eventName, string subscriberName, string selector) => $"{eventName}_{subscriberName}_{selector}";
```
chomick, line 13: you may want to consider how to handle the 'default' selector (i.e. when there is no selection rules defined for a subscriber) but it is fine for now, I guess (or maybe remove the 'selector' for now as it cannot be used yet?)

tabman83: Probably makes more sense to remove it for now until we have better understanding of how we are gonna define and use the selector itself. Updating.

chomick:
@tabman83 what is going to happen with this PR? Are we going to merge it or was it merged to another PR and this should be abandond?

PR #111, opened 2020-06-23
71440 integration tests

dawidkeshop:
Integration tests are there, but there are some issues to fix:
1. Authorization for API integration tests doesn't work properly, the `token` for tests isn't valid. 
1. `IsDev` as well as `IsIntegration` attribute is not recognized properly on CI, so API integration tests are not executed as part of release.


src/CaptainHook.Api/Startup.cs
```diff
@@ -137,6 +137,11 @@ public void ConfigureServices(IServiceCollection services)
                     // See https://eshopworld.visualstudio.com/evo-core/_git/security-services-telemetry?path=%2FREADME.md&_a=preview
                     // x.AddJwtBearerEventsTelemetry(bb); 
                 });
+                services.AddAuthorization(options =>
```
chomick, line 140: Not necessarily part of this PR but the authentication requires only this setup:
                services
                    .AddAuthentication (JwtBearerDefaults.AuthenticationScheme)
                    .AddIdentityServerAuthentication (x =>
                    {
                        x.ApiName = serviceConfigurationOptions.Value.ApiName;
                        x.Authority = serviceConfigurationOptions.Value.Authority;
                    });

For the scope checks alone, looks ok.

src/Tests/CaptainHook.Api.Tests/GetAllSubscribersTests.cs
```diff
@@ -0,0 +1,53 @@
...
+namespace CaptainHook.Api.Tests
+{
+    [Collection(ApiClientCollection.TestFixtureName)]
+    public class GetAllSubscribersTests : IDisposable
+    {
+        private readonly ICaptainHookClient _authenticatedClient;
+        private readonly ICaptainHookClient _unauthenticatedClient;
+
+
+        public GetAllSubscribersTests(ApiClientFixture testFixture)
```
chomick, line 21: The way the fixture is defined in master right now, this will work only locally.
The fixture should be using STS url from the settings files.

src/CaptainHook.Api/Startup.cs
dawidkeshop: I think such change should be discussed with author of original code, maybe there was a reason for extending this configuration.

src/Tests/CaptainHook.Api.Tests/GetAllSubscribersTests.cs
dawidkeshop: These tests run only locally anyway, we should play another story/stories to make tests working in pipelines, and then fix CI configuration and tests authentication.

tabman83: STS _is_ currently defined in the settings files

PR #115, opened 2020-06-24
#73562 - ensure Merge uses EventType and SubscriberName for matching

mszymeshop:
### What
This PR enables MERGE and RELOAD which were both broken for config changes. *Please ensure to question every line, this PR is ⚠️ crucial*

### Testing
Locally, the following has been tested:
- config in KV
- config in KV and Cosmos independent
- config in KV and Cosmos and KV changed and reloaded
- config in KV and Cosmos and Cosmos changed and reloaded
- config in KV and Cosmos override
- config in KV and Cosmos override changed and reloaded

I am going to perform more tests while this PR is being reviewed.

### DIFF
--- a/src/CaptainHook.DirectorService/Infrastructure/ReaderServicesManager.cs
+++ b/src/CaptainHook.DirectorService/Infrastructure/ReaderServicesManager.cs
@@ -108,11 +108,10 @@ namespace CaptainHook.DirectorService.Infrastructure
             }
         }
 
-        private static byte[] BuildInitializationData(SubscriberConfiguration subscriber, IEnumerable<WebhookConfig> webhooks)
+        private static byte[] BuildInitializationData(SubscriberConfiguration subscriber, IEnumerable<WebhookConfig> _)
         {
-            var webhookConfig = webhooks.SingleOrDefault(x => x.Name == subscriber.Name);
             return EventReaderInitData
-                .FromSubscriberConfiguration(subscriber, webhookConfig)
+                .FromSubscriberConfiguration(subscriber, subscriber)
                 .ToByteArray();
         }
     }
--- a/src/CaptainHook.Storage.Cosmos/Models/EndpointDocument.cs
+++ b/src/CaptainHook.Storage.Cosmos/Models/EndpointDocument.cs
@@ -77,11 +77,5 @@ namespace CaptainHook.Storage.Cosmos.Models
         /// </summary>
         [JsonProperty("eventName")]
         public string EventName { get; set; }
-
-        /// <summary>
-        /// Identifies the subscriber by event and name.
-        /// </summary>
-        [JsonIgnore]
-        public string SubscriberId => $"{EventName}-{SubscriberName}";
     }
 }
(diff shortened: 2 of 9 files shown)


src/CaptainHook.DirectorService/Infrastructure/ReaderServicesManager.cs
```diff
@@ -108,11 +108,10 @@ private async Task DeleteReaderServicesAsync(IEnumerable<string> oldNames, Cance
             }
         }
 
-        private static byte[] BuildInitializationData(SubscriberConfiguration subscriber, IEnumerable<WebhookConfig> webhooks)
+        private static byte[] BuildInitializationData(SubscriberConfiguration subscriber, IEnumerable<WebhookConfig> _)
```
dawidkeshop, line 111: Can we just remove the second parameter it's not used?

src/CaptainHook.Storage.Cosmos/Models/EndpointDocument.cs
```diff
@@ -81,7 +81,7 @@ internal class EndpointDocument
         /// <summary>
         /// Identifies the subscriber by event and name.
         /// </summary>
-        [JsonProperty("subscriberId")]
-        public string SubscriberId => $"{EventName}_{SubscriberName}";
+        [JsonIgnore]
+        public string SubscriberId => $"{EventName}-{SubscriberName}";
```
dawidkeshop, line 85: Do we need Id to be stored? Wouldn't be enough to create it on Entity level.

src/CaptainHook.DirectorService/Infrastructure/ReaderServicesManager.cs
mszymeshop: Agreed, but I focused on making sure this PR is as readable as possible without the introduction of additional changes. I believe we already have a tech debt story to address the webconfig aspect.

src/CaptainHook.Storage.Cosmos/Models/EndpointDocument.cs
mszymeshop: It is not going to be stored as it is a computed property with JsonIgnore. Having said that, I agree that it shouldn't be here really as it is not part of the Document. We need to reiterate how we read the data though as at the moment the grouping is done by SubscriberId and, again, it made the most sense to include it here for the time being.

src/CaptainHook.DirectorService/Infrastructure/ReaderServicesManager.cs
dawidkeshop: OK, makes sense

src/CaptainHook.Storage.Cosmos/Models/EndpointDocument.cs
dawidkeshop: OK, makes sense.

mszymeshop:
@dawidkeshop I have updated the PR so that `SubscriberId` is no longer required

PR #116, opened 2020-06-24
73487 tests for passwords clearing logic

dawidkeshop:


src/CaptainHook.Api/Controllers/SubscribersController.cs
```diff
@@ -42,7 +39,7 @@ public async Task<IActionResult> GetAll()
                 var directorServiceClient = ServiceProxy.Create<IDirectorServiceRemoting>(directorServiceUri);
                 var subscribers = await directorServiceClient.GetAllSubscribersAsync();
 
-                HideCredentials(subscribers.Values);
+                AuthenticationConfigSanitizer.Sanitize(subscribers.Values);
```
chomick, line 42: You went so far, you could have done it as a dependency interface rather than static method :)
But ok for this PR

PR #117, opened 2020-06-26
Configuration data removed from ServiceCreatedEvent

dawidkeshop:


mszymeshop:
Another option would be to sanitise the config. Are we sure we want to remove it entirely?

PR #118, opened 2020-06-26
Honor cancellation token

tabman83:
When deploying CH, all readers are recreated and old ones removed.
Any new reader puts additional strain on the service bus what may result in SB activating throttling. This is not confirmed whether adding readers alone is enough to cause those issues as some deployments work fine but at other times, it does not. There still can be another underlying issue that is resulting in the behavior below.
When service bus throttling starts, Director Service fails to delete old reader services. The old reader services fail to get deleted as the operation time outs. There is an assumption this happens due to internal back-off in service bus SDK when throttling takes place.

Do not let deleting reader service to timeout when service bus is throttling requests - cancellation token should be passed around deeper (incl. service bus interactions) and that should result in cancellation of the re-connect or any other backed-off operation on service bus. That in turn should result in delete to complete immediately.


src/CaptainHook.EventReaderService/EventReaderService.cs
```diff
@@ -273,32 +273,30 @@ protected override async Task RunAsync(CancellationToken cancellationToken)
...
-                            ).ExecuteAsync(SetupServiceBus);
-                    }
-                    catch (Exception exception)
-                    {
-                        BigBrother.Write(exception);
+                            ).ExecuteAsync(ct => SetupServiceBusAsync(ct), cancellationToken);
                     }
                 }
 
+                cancellationToken.ThrowIfCancellationRequested();
```
chomick, line 291: why would we throw the exception instead of letting the loop to finish?

src/CaptainHook.EventReaderService/EventReaderService.cs
```diff
@@ -273,32 +273,30 @@ protected override async Task RunAsync(CancellationToken cancellationToken)
...
+            }
+            catch (OperationCanceledException)
```
chomick, line 293: i would add a log after the loop to say we are leaving if that was the intention here

src/CaptainHook.EventReaderService/EventReaderService.cs
```diff
@@ -273,32 +273,30 @@ protected override async Task RunAsync(CancellationToken cancellationToken)
...
+            {
                 _bigBrother.Publish(new Common.Telemetry.CancellationRequestedEvent { FabricId = $"{Context.ServiceName}:{Context.ReplicaId}" });
             }
-            catch (Exception e)
+            catch (Exception exception)
             {
-                BigBrother.Write(e.ToExceptionEvent());
+                BigBrother.Write(exception);
```
chomick, line 299: looks like old code - can we use the bb reference rather than static method?

src/CaptainHook.EventReaderService/EventReaderService.cs
tabman83: If I don't explicitly throw here it means I have to 
- log the cancellation because we finished the loop
- catch the OperationCancelledEx anyway to log the cancellation again (this is thrown by the inner Service Bus methods which not get passed the cancellation token)
So I would be logging the same thing twice as I come from two different paths.
I chose instead to throw explicitly so I catch and log the cancellation regardless of where it is originated from.

src/CaptainHook.EventReaderService/EventReaderService.cs
tabman83: That's the reason for the throw: we bump out to the operationcanceledexception catch block and we log the cancellation regardless of where it originated from

src/CaptainHook.EventReaderService/EventReaderService.cs
tabman83: Yep, changed.

src/CaptainHook.EventReaderService/EventReaderService.cs
```diff
@@ -271,34 +271,32 @@ protected override async Task RunAsync(CancellationToken cancellationToken)
                     }
                     catch (ServiceBusCommunicationException sbCommunicationException)
                     {
-                        BigBrother.Write(sbCommunicationException);
+                        _bigBrother.Publish(sbCommunicationException.ToExceptionEvent());
```
mszymeshop, line 274: :100:

PR #122, opened 2020-07-03
Set MinReplicaSetSize to 2 to facilitate cluster upgrades

tabman83:
Set MinReplicaSetSize = 2 to facilitate cluster upgrades as asked by DevOps.


mszymeshop:
@tabman83 what is the plan regarding # of replicas in environments? PREP and PROD will remain at 3 replicas right?

tabman83:
Ihor is asking for this change in CI/TEST/SAND only as when the cluster goes to 2 nodes on core scale set our services will be in quorum loss state (stateful service quorum is minreplicasetsize)

PR #123, opened 2020-07-03
72985 regenerate only changed readers

dawidkeshop:
**A bit of explanation:**

1. Reader service name will now contain subscriber configuration hash. This implies that we can rely only on names to decide if subscriber configuration for particular reader has been changed, no need to compare "previous" and "current" KV configurations.
1. The goal is to make readers creation during Director startup more intelligent, to only create and delete readers which configurations has been changed. Effectively it means that this operation is logically the same as reload configuration by an API call. So now we can have only one method which performs this logic.
1. Previous implementation has a bug. Names of services to delete were generated for every subscriber configuration loaded from KV. This means that if a subscriber had removed from KV, we wouldn't have deleted the related reader. Now we will remove all readers except these which want to keep.
1. Since webhooks are not used to build Reader initialization data, then I removed all usages of it in code I've touched.


src/CaptainHook.DirectorService/Infrastructure/ReaderServicesManager.cs
```diff
@@ -66,27 +41,41 @@ public async Task CreateReadersAsync(IEnumerable<SubscriberConfiguration> subscr
...
+        public async Task RefreshReadersAsync(IEnumerable<SubscriberConfiguration> newSubscribers, IEnumerable<string> deployedServicesNames, CancellationToken cancellationToken)
         {
-            var comparisonResult = new SubscriberConfigurationComparer().Compare(currentSubscribers, newSubscribers);
-            _bigBrother.Publish(new RefreshSubscribersEvent(comparisonResult));
+            // Describe the situation
+            var desiredReaders = newSubscribers.Select(c => new DesiredReaderDefinition(c)).ToList();
+            var existingReaders = deployedServicesNames
+                .Where(s => s.StartsWith($"fabric:/{Constants.CaptainHookApplication.ApplicationName}/{ServiceNaming.EventReaderServiceShortName}"))
+                .Select(s => new ExistingReaderDefinition(s)).ToList();
+
```
tabman83, line 53: Just to err on the side of caution I would make the StartsWith string comparison case insensitive.

src/CaptainHook.DirectorService/Infrastructure/ReaderServicesManager.cs
```diff
@@ -108,11 +97,41 @@ private async Task DeleteReaderServicesAsync(IEnumerable<string> oldNames, Cance
...
+            public DesiredReaderDefinition(SubscriberConfiguration subscriberConfig)
+            {
+                SubscriberConfig = subscriberConfig;
+                ServiceName = ServiceNaming.EventReaderServiceFullUri(subscriberConfig.EventType, subscriberConfig.SubscriberName, subscriberConfig.DLQMode.HasValue);
+                ServiceNameWithSuffix = $"{ServiceName}-{GetEncodedHash(subscriberConfig)}";
+            }
+
+            private static string GetEncodedHash(SubscriberConfiguration configuration)
+            {
+                var md5 = new MD5CryptoServiceProvider();
```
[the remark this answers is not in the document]
dawidkeshop: Good catch! I didn't notice that earlier, thank you!

src/CaptainHook.DirectorService/Infrastructure/ReaderServicesManager.cs
dawidkeshop: Right

src/CaptainHook.DirectorService/Infrastructure/ReaderServicesManager.cs
```diff
@@ -108,11 +97,41 @@ private async Task DeleteReaderServicesAsync(IEnumerable<string> oldNames, Cance
             }
         }
 
-        private static byte[] BuildInitializationData(SubscriberConfiguration subscriber, IEnumerable<WebhookConfig> _)
+        private class ExistingReaderDefinition
+        {
+            private static readonly Regex RemoveSuffixRegex = new Regex("(|-a|-b|-\\d{14}|-[a-zA-Z0-9]{22})$", RegexOptions.Compiled);
```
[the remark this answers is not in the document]
dawidkeshop: In theory you're right, but nothing stops anybody from deploying old version with `a/b` naming again, then the newest one, and it will result in a mess with readers.
I think keeping this code is just safer.

src/CaptainHook.DirectorService/Infrastructure/ReaderServicesManager.cs
```diff
@@ -66,27 +42,41 @@ public async Task CreateReadersAsync(IEnumerable<SubscriberConfiguration> subscr
...
         /// <returns></returns>
-        public async Task RefreshReadersAsync(IDictionary<string, SubscriberConfiguration> newSubscribers, IEnumerable<WebhookConfig> newWebhooks, IDictionary<string, SubscriberConfiguration> currentSubscribers, IEnumerable<string> deployedServicesNames, CancellationToken cancellationToken)
+        public async Task RefreshReadersAsync(IEnumerable<SubscriberConfiguration> newSubscribers, IEnumerable<string> deployedServicesNames, CancellationToken cancellationToken)
         {
-            var comparisonResult = new SubscriberConfigurationComparer().Compare(currentSubscribers, newSubscribers);
-            _bigBrother.Publish(new RefreshSubscribersEvent(comparisonResult));
+            // Describe the situation
+            var desiredReaders = newSubscribers.Select(c => new DesiredReaderDefinition(c)).ToList();
+            var existingReaders = deployedServicesNames
+                .Where(s => s.StartsWith($"fabric:/{Constants.CaptainHookApplication.ApplicationName}/{ServiceNaming.EventReaderServiceShortName}", StringComparison.InvariantCultureIgnoreCase))
```
mszymeshop, line 52: nit: invariant culture not needed here, consider ordinalignorecase instead

src/CaptainHook.DirectorService/Infrastructure/ReaderServicesManager.cs
```diff
@@ -66,27 +42,41 @@ public async Task CreateReadersAsync(IEnumerable<SubscriberConfiguration> subscr
...
+                .Select(s => new ExistingReaderDefinition(s)).ToList();
+
+            // Detect changes
+            var changed = desiredReaders.Where(d => existingReaders.Any(e => d.ServiceName == e.ServiceName && d.ServiceNameWithSuffix != e.ServiceNameWithSuffix)).ToList();
```
mszymeshop, line 56: nit: I think set operations would make the algorithm clearer. Fine to stay as it is though.

src/CaptainHook.DirectorService/Infrastructure/ReaderServicesManager.cs
```diff
@@ -66,27 +42,41 @@ public async Task CreateReadersAsync(IEnumerable<SubscriberConfiguration> subscr
...
-            var servicesToDelete = comparisonResult.Removed.Values.Union(comparisonResult.Changed.Values).SelectMany(s => _readerServiceNameGenerator.FindOldNames(s.ToSubscriberNaming(), deployedServicesNames));
+            // now we know the numbers, so we can publish event
+            _bigBrother.Publish(new RefreshSubscribersEvent(added.Select(s => s.ServiceName), deleted.Select(s => s.ServiceName), changed.Select(s => s.ServiceName)));
 
-            await CreateReaderServicesAsync(servicesToCreate, newWebhooks, cancellationToken);
-            await DeleteReaderServicesAsync(servicesToDelete, cancellationToken);
+            // prepare to actual work (put changed to added, delete everything except desired)
+            var servicesToCreate = added.Union(changed).ToDictionary(dsd => dsd.ServiceNameWithSuffix, kvp => kvp.SubscriberConfig);
+            var allServiceNamesToDelete = existingReaders.Select(e => e.ServiceNameWithSuffix).Except(desiredReaders.Select(d => d.ServiceNameWithSuffix));
```
mszymeshop, line 65: nit: usage of equality comparers on set operations is way clearer

src/CaptainHook.DirectorService/Infrastructure/ReaderServicesManager.cs
dawidkeshop: Ok, you're right

src/CaptainHook.DirectorService/Infrastructure/ReaderServicesManager.cs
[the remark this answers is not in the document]
tabman83: That makes perfect sense then!

src/CaptainHook.DirectorService/Infrastructure/ReaderServicesManager.cs
dawidkeshop: The thing is that I'm comparing items of different types, so it won't be possible to use standard set operations.

src/CaptainHook.DirectorService/Infrastructure/ReaderServicesManager.cs
dawidkeshop: Same thing here - different types compared.

dawidkeshop:
Will be merged as part of #129

dawidkeshop:
Will be merged as part of #129

PR #124, opened 2020-07-04
#72985 - added retry mechanism for reader deletion

chomick:
- Added better configuration loading mechanism - we load only selected KV entries for appsettings and we load ALL entries only in Director Service Run method. 
- Added retry mechanism for reader service deletion and optimised it to run in parallel


src/CaptainHook.DirectorService/DirectorService.cs
```diff
@@ -106,8 +106,8 @@ protected override async Task RunAsync(CancellationToken cancellationToken)
             }
             catch (Exception exception)
             {
-                _bigBrother.Publish(exception);
-                throw;
+                _bigBrother.Publish(exception.ToExceptionEvent ());
+                throw; // should this really be rethrown?
```
tabman83, line 110: I think it is important we re-throw it.

If we just complete RunAsync(), it will be called again on the same service object if this is demoted and promoted to primary again. While if we throw an exception the object will be destroyed and the health error reported. This gives us the option to start with a clean slate if the original exception has left the service in an inconsistent state.

Reference: https://docs.microsoft.com/en-us/azure/service-fabric/service-fabric-reliable-services-lifecycle#notes-on-the-service-lifecycle

src/CaptainHook.Common/Configuration/TempConfigLoader.cs
```diff
@@ -0,0 +1,70 @@
...
+
+namespace CaptainHook.Common.Configuration
+{
+    /// <summary>
+    /// This is a temporary solution while KV cannot be fetched fully (as it is too big)
+    /// Once EDA config is moved to DB, this can be removed and replaced with normal configuration initialisation
+    /// </summary>
+    public static class TempConfigLoader
+    {
+        private static SecretClient GetClient ()
```
mszymeshop, line 15: This isn't the coding style we use in the project :)

src/CaptainHook.EventHandlerActor/Handlers/EventHandlerFactory.cs
```diff
@@ -10,29 +9,23 @@ namespace CaptainHook.EventHandlerActor.Handlers
...
-        private readonly IIndex<string, WebhookConfig> _webHookConfig;
         private readonly IHttpClientFactory _httpClientFactory;
         private readonly IAuthenticationHandlerFactory _authenticationHandlerFactory;
         private readonly IRequestLogger _requestLogger;
         private readonly IRequestBuilder _requestBuilder;
 
         public EventHandlerFactory(
             IBigBrother bigBrother,
-            IIndex<string, SubscriberConfiguration> subscriberConfigurations,
-            IIndex<string, WebhookConfig> webHookConfig,
```
mszymeshop, line 23: I thought there was a separate story to handle this.

src/CaptainHook.Common/Configuration/TempConfigLoader.cs
```diff
@@ -0,0 +1,70 @@
...
+using CaptainHook.Common.Configuration.KeyVault;
+using Microsoft.Extensions.Configuration;
+
+namespace CaptainHook.Common.Configuration
+{
+    /// <summary>
+    /// This is a temporary solution while KV cannot be fetched fully (as it is too big)
+    /// Once EDA config is moved to DB, this can be removed and replaced with normal configuration initialisation
+    /// </summary>
+    public static class TempConfigLoader
```
dawidkeshop, line 13: I wouldn't put `Temp` inside a class name, because it doesn't tell what is the actual responsibility of this class. I would rather name this `ApplicationConfigLoader`. (My way of thinking is that it loads config values required for application to run, not related with subscribers.)
Another thing is, that usually things planned to be "temporary" lasts forever. :)

src/CaptainHook.Common/Configuration/TempConfigLoader.cs
```diff
@@ -0,0 +1,70 @@
...
+            return builder
+                .AddInMemoryCollection (entries)
+                .AddEnvironmentVariables ()
+                .Build ();
+        }
+
+        private static KeyValuePair<string, string> AddEntry (SecretClient client, string kvPath)
+        {
+            var items = kvPath.Split ("--");
+            var configKey = string.Join (":", items);
```
mszymeshop, line 59: would a simple replace be a better option here? Mostly because otherwise a bunch of smaller strings are created on split (and they are not needed for anything) and replace would be more efficient.

src/CaptainHook.DirectorService/Infrastructure/ReaderServicesManager.cs
```diff
@@ -99,12 +111,40 @@ private async Task CreateReaderServicesAsync(IDictionary<string, SubscriberConfi
 
         private async Task DeleteReaderServicesAsync(IEnumerable<string> oldNames, CancellationToken cancellationToken)
```
mszymeshop, line 112: Do you plan on covering this with unit tests?

src/CaptainHook.Common/Configuration/TempConfigLoader.cs
```diff
@@ -0,0 +1,70 @@
...
+    {
+        private static SecretClient GetClient ()
+        {
+            return new SecretClient (
+                new Uri (Environment.GetEnvironmentVariable (ConfigurationSettings.KeyVaultUriEnvVariable)),
+                new AzureServiceTokenCredential (),
+                new SecretClientOptions ());
+        }
+
+        public static IConfigurationRoot Load ()
```
dawidkeshop, line 23: As this is `public` method of this class, I would move it to the top. I think it's easier to read code if methods called from a method  (like `GetClient` here) are placed in file after this method.

src/CaptainHook.DirectorService/Events/ReaderServicesDeletionEvent.cs
```diff
@@ -0,0 +1,36 @@
...
+            DeletedNames = JsonConvert.SerializeObject (deletedNames.ToArray ());
+        }
+
+        public void SetFailed (IEnumerable<FailedReaderServiceDeletion> failed)
+        {
+            Failed = JsonConvert.SerializeObject (failed.ToArray ());
+        }
+    }
+
+    public readonly struct FailedReaderServiceDeletion
```
dawidkeshop, line 24: Why this is a `struct`?

src/CaptainHook.DirectorService/Events/ReaderServicesDeletionEvent.cs
```diff
@@ -0,0 +1,36 @@
...
+using Newtonsoft.Json;
+
+namespace CaptainHook.DirectorService.Events
+{
+    public class ReaderServicesDeletionEvent : TimedTelemetryEvent
+    {
+        public string DeletedNames { get; private set; }
+        public string Failed { get; private set; }
+
+        public void SetDeletedNames (IEnumerable<string> deletedNames)
```
dawidkeshop, line 13: Wouldn't be better to have a class with two constructors instead of two methods which manipulates the state? I think later usage would be simpler - one constructor could by used in the `try` block, another one in `catch` block.

src/CaptainHook.Common/Configuration/TempConfigLoader.cs
chomick: yes, initially i had different approach but later I settled on this and simple replace will do the job as well

src/CaptainHook.Common/Configuration/TempConfigLoader.cs
chomick: I did it on purpose as this class should be removed - I don't want anyone to invest any time/effort into this class and name will hopefully prevent people from doing it.
Also, it will be a reminder that this is wrong we should get rid of it ASAP

src/CaptainHook.Common/Configuration/TempConfigLoader.cs
chomick: yep, will do

src/CaptainHook.DirectorService/Events/ReaderServicesDeletionEvent.cs
chomick: No, because the instance has to be created first (it is timed event) and data is assigned after things are done - so can't be done via ctor.

src/CaptainHook.DirectorService/Events/ReaderServicesDeletionEvent.cs
chomick: To be more efficient

src/CaptainHook.DirectorService/Infrastructure/ReaderServicesManager.cs
chomick: I have another PR on top of this where I split this class into two:
- change detection
- execution of set of changes

I have covered them with unit tests

src/CaptainHook.EventHandlerActor/Handlers/EventHandlerFactory.cs
chomick: It was such a small change that decided to do it here and now - no risk imo

src/CaptainHook.Common/Configuration/TempConfigLoader.cs
chomick: As long as I am not changing existing formatting should be fine :)

src/CaptainHook.DirectorService/Infrastructure/ReaderServicesManager.cs
chomick: Actually, I do not have the exception cases for delete covered in unit tests - will add but this will have to be done in the split class - to much mess with testing in the current shape.
I did tested it manually though

chomick:
Closing the PR as the changes are now included in 
https://github.com/eShopWorld/captain-hook/pull/130

PR #125, opened 2020-07-05
Retry logic for create topic and setup service bus

tabman83:
This PR addresses three issue:

- ServiceBusNamespaceExtensions.cs, CreateTopicIfNotExists:
Use retry logic to query for the new topic after creation. As stated by the documentation the topic may not be readily available after the CreateAsync() method as it just queues the operation for execution. The implemented logic will retry with an exponential back-off until the topic is available.

- EventReaderService.cs@272
The catch block has been changed to a more generic ServiceBusException as that is the ancestor for the ServerBusyException which is thrown by the event reader main loop when it is being throttled in Service Bus.

- EventReaderService.cs@281
A specific event is being sent to AI when the swallowing catch block is reached because we want to be able to recognize the exception that might lead the event reader service to an inconsistent state.


mszymeshop:
@tabman83 were you able to observe improve the behaviour of the app or are those blind changes mostly?

tabman83:
For the change on CreateTopicIfNotExists:
the error happens mostly locally because we are creating all those topics ending in `-username`. Otherwise I would expect that to happen only when adding a new event in the configuration and we have had that kind of exception a while ago,if you recall it is something I showed you on AI.

For the change on EventReaderService.cs@272:
I was able to reproduce it locally and see the exception being swallowed in the reader while the director times out. That happens on CI mostly because of throttling from SB standard tier.

PR #127, opened 2020-07-06
Temporarily move KV test out of unit testing

tabman83:
This test needs a env variable configured with the KV to run. So excluding for now so we can satisfactorily run after CI deploy.

### DIFF
--- a/src/Tests/CaptainHook.Tests/Director/SubscriberConfigurationLoaderTests.cs
+++ b/src/Tests/CaptainHook.Tests/Director/SubscriberConfigurationLoaderTests.cs
@@ -26,7 +26,7 @@ namespace CaptainHook.Tests.Director
             _subscriberConfigurationLoader = new SubscriberConfigurationLoader(_subscriberRepositoryMock.Object, _configurationMergerMock.Object);
         }
 
-        [Fact, IsLayer0]
+        [Fact, IsDev]
         public async Task MergeIsInvoked()
         {
             var result = await _subscriberConfigurationLoader.LoadAsync();


src/Tests/CaptainHook.Tests/Director/SubscriberConfigurationLoaderTests.cs
```diff
@@ -26,7 +26,7 @@ public SubscriberConfigurationLoaderTests()
             _subscriberConfigurationLoader = new SubscriberConfigurationLoader(_subscriberRepositoryMock.Object, _configurationMergerMock.Object);
         }
 
-        [Fact, IsLayer0]
+        [Fact, IsDev]
```
chomick, line 29: If you need it to be backward compatible, leave the old attribute for now.

dawidkeshop, line 29: This test loads data from KV, are you sure that `IsDev` is a correct attribute in that case?

chomick: Sorry, actually I missed that - should there not be IsUnit attribute used instead?

tabman83: I just wanna exclude it from running on CI for now, but leave the option to run it locally on request.

chomick: Never used IsDev attribute before so can't comment on its purpose.

PR #132, opened 2020-07-07
68215 fix KV loading tests

dawidkeshop:
Simple change which allows to execute all CaptainHook.Tests tests locally. 
Fixed tests are marked with `IsDev`, so won't fail on environments.

### DIFF
--- a/src/Tests/CaptainHook.Tests/Configuration/ConfigLoaderTests.cs
+++ b/src/Tests/CaptainHook.Tests/Configuration/ConfigLoaderTests.cs
@@ -9,7 +9,7 @@ namespace CaptainHook.Tests.Configuration
         [IsDev]
         public void ConfigNotEmpty()
         {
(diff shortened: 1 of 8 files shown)


src/Tests/CaptainHook.Tests/Configuration/ConfigLoaderTests.cs
```diff
@@ -9,7 +9,7 @@ public class ConfigLoaderTests
         [IsDev]
         public void ConfigNotEmpty()
         {
-            var configuration = CaptainHook.Common.Configuration.Configuration.Load();
+            var configuration = CaptainHook.Common.Configuration.Configuration.Load("https://esw-tooling-ci-we.vault.azure.net/");
```
mszymeshop, line 12: Is this something we can move to config or at least reuse between both tests?

PR #135, opened 2020-07-10
73244 basic infrastructure for API validation 

dawidkeshop:
Domain layer initially contained validators and handlers, and has references to MediatR and CaptainHook.Contracts. I don't think this was a good design, so I extracted these to the Application project. 

On the other hand API still has reference to Domain, because Domain contains definition of errors and API maps these to HTTP codes and responses. I don't like this that way, maybe we should extract another project for errors only?


src/CaptainHook.Domain/Common/BusinessError.cs
```diff
@@ -0,0 +1,37 @@
...
+        protected ErrorBase(string message, List<Failure> failures = default)
+        {
+            Message = message;
+            Failures = failures;
+        }
+    }
+
+    public class Failure
+    {
+        public string Code { get; set; }
```
tabman83, line 33: What is the code gonna be for us?
Would it make sense to have a numeric-only code (thus an integer) ?
Where are we gonna take the error codes from?

Those are just question that we are going to need to answer, but they don't prevent me from accepting the idea.

src/CaptainHook.Api/Controllers/SubscribersController.cs
```diff
@@ -17,19 +26,20 @@ namespace CaptainHook.Api.Controllers
...
 
         /// <summary>
         /// ctor
         /// </summary>
         /// <param name="bigBrother"></param>
-        public SubscribersController(IBigBrother bigBrother)
+        public SubscribersController(IBigBrother bigBrother, IMediator mediator)
         {
             _bigBrother = bigBrother;
+            _mediator = mediator;
```
tabman83, line 42: Just to be on the safe side you can add a `?? throw new ArgumentNullException` here.

src/CaptainHook.Domain/Common/BusinessError.cs
dawidkeshop: I think it was a @chomick's idea to have these codes, if I got the idea correctly codes should give us an ability to easily understand what happened because these meant to be unique.

src/CaptainHook.Api/Controllers/SubscribersController.cs
```diff
@@ -56,5 +62,52 @@ public async Task<IActionResult> GetAll()
                 return BadRequest();
             }
         }
+
+        [HttpGet]
+        [Route("getbyevent")]
```
chomick, line 67: is that temporary API?
are we going to discard this PR afterwards?
Or are you going ot remove this endpoint before merge?

chomick: My point is about contract itself - it should be more like:
```[HttpGet ("event-name/{eventName}")]```

File: src/CaptainHook.Application/Handlers/Subscribers/GetSubscribersForEventQueryHandler.cs [hunk 4]
```diff
@@ -0,0 +1,35 @@
...
+        }
+
+        public async Task<EitherErrorOr<List<SubscriberDto>>> Handle(GetSubscribersForEventQuery query, CancellationToken cancellationToken)
+        {
+            if (query.Name == "error")
+            {
+                return new BusinessError("Error is not a valid name!");
+            }
+
+            var subscriberEntities = await _repository.GetSubscribersListAsync(query.Name);
```
chomick, line 30: This should return all data, incl. KV, not only DB
Are we doing too much proper-functionality code in this spike-PR?
Is this code going to stay or be removed?

src/CaptainHook.Domain/Repositories/ISubscriberRepository.cs
```diff
@@ -11,12 +12,12 @@ public interface ISubscriberRepository
         /// </summary>
         /// <param name="eventName">The event name</param>
         /// <returns></returns>
-        public Task<IEnumerable<SubscriberEntity>> GetSubscribersListAsync(string eventName);
+        public Task<EitherErrorOr<List<SubscriberEntity>>> GetSubscribersListAsync(string eventName);
```
chomick, line 15: Why not IEnum<> or at least IList<>?

src/CaptainHook.Domain/Results/EitherErrorOr.cs
```diff
@@ -0,0 +1,40 @@
+﻿using System;
+
+namespace CaptainHook.Domain.Results
+{
+    public class EitherErrorOr<TData>
```
chomick, line 5: Names with "Or", "And" etc are indication of a problem with the code usually :)
Can we name it more like "OperationResult<>" or "BusinessResult<>"?

src/CaptainHook.Domain/Errors/EntityNotFoundError.cs
```diff
@@ -0,0 +1,12 @@
+﻿using CaptainHook.Domain.Results;
+
+namespace CaptainHook.Domain.Errors
+{
+    public class EntityNotFoundError : ErrorBase
+    {
+        public EntityNotFoundError(string type, string key) : 
+            base("Entity not found", new Failure("NotFound", $"Can't find an entity of type '{type}' by '{key}'"))
```
chomick, line 8: I wonder if we should maintain the codes in a single place (static class with their definitions) so they can be easily searched and you also can ensure there is no duplicates.
You could also keep them in some dictionary: { "code" => "message template string" } but you would have to assume certain order of params there (if using %-notation). But at min I would have list of static/consts with the codes.

src/Tests/CaptainHook.Application.Tests/ValidatorPipelineBehaviorTests.cs
```diff
@@ -0,0 +1,77 @@
...
+            var thisAssembly = Assembly.GetExecutingAssembly();
+
+            var builder = new ContainerBuilder();
+            builder.RegisterMediatorInfrastructure(thisAssembly, thisAssembly);
+            var container = builder.Build();
+
+            _mediator = container.Resolve<IMediator>();
+        }
+
+        [Fact]
```
chomick, line 29: You also need "IsUnit" for it to be run on the pipelines

src/CaptainHook.Domain/Common/BusinessError.cs
chomick: The 'code' was meant to be used in the DTOs.
If we are using these business-results error classes (which map to a code), question is what we would use in our application logic. Would we ever check the results in business domain and have some decisions based on the type of validation issue?
Would we check the code or the type (class) of error?

dawidkeshop: I don't think we need a code to check what is the problem in the domain/application layers. Specific error class should be enough to cover that situation if it ever happen. 
So what I understand is to:
1. Create a DTO for Error with Code and Message. 
2. Map Error DTO from domain error, specific domain error class will be mapped to a specific Code.
Is this correct?

src/CaptainHook.Api/Controllers/SubscribersController.cs
dawidkeshop: Sure

src/CaptainHook.Api/Controllers/SubscribersController.cs
dawidkeshop: This is temporary, just to show that it works, and errors are returned in the HTTP response. A person who gets specific task should replace this with actual implementation. 
I think it's nothing wrong to merge it, nobody will use but it can help developer to jump into the design. What do you think?

src/CaptainHook.Application/Handlers/Subscribers/GetSubscribersForEventQueryHandler.cs
dawidkeshop: It's temporary, just to show that errors can be passed through all layers and returned to the API user.

src/CaptainHook.Domain/Results/EitherErrorOr.cs
dawidkeshop: For sure, if there is a `Or` or `And` in the name of a method something is wrong, not sure if it applies for types also.
In that case my idea was that this discriminated union can contain either error or some data, so the name is very descriptive as for me. Of course I can change it, I would prefer `OperationResult<>` in that case.

src/CaptainHook.Domain/Errors/EntityNotFoundError.cs
dawidkeshop: I agree, I think when we implement some actual functionality, we can have more clarity about what's the best way of keeping error codes.

src/Tests/CaptainHook.Application.Tests/ValidatorPipelineBehaviorTests.cs
dawidkeshop: Right, thanks!

src/CaptainHook.Domain/Repositories/ISubscriberRepository.cs
dawidkeshop: Fixed

src/CaptainHook.Domain/Common/BusinessError.cs
chomick: I see a value in having the code imposed by the specific error class as well :) It avoids having additional class -> code mapping which actually sits well as you did it now.
My question was more on how else we plan to use code - would we use it in our domain layers or it would simply represent value passed outside of the domain.

src/CaptainHook.Api/Controllers/SubscribersController.cs
chomick: fair enough - I just wanted to avoid strange endpoints in the contract - but not the end of the world if you remove it/update it on subsequent PR

src/CaptainHook.Application/Handlers/Subscribers/GetSubscribersForEventQueryHandler.cs
chomick: np

src/CaptainHook.Domain/Results/EitherErrorOr.cs
chomick: Sounds good with the operationresult<>

src/CaptainHook.Domain/Errors/EntityNotFoundError.cs
chomick: agreed

PR #139, opened 2020-07-13
Task 75634 extend source config with replace option

mszymeshop:
### What
Parser Location is extended to facilitate a collection of Replace entries

### Why
We want to be able to extend CH with Replace transformation without affecting the existing code. New class to handle Source is created.

### Examples
Json entry:

```json
      {
        "Source": {
          "Replace": {
            "ordercode": "$.OrderCode",
            "selector": "$.TenantCode"
          }
        },
        "Destination": {
          "RuleAction": "Route"
        },
        "Routes": [
          {...
```
Powershell
```ps1
setConfig 'event--26--webhookconfig--webhookrequestrules--2--source--replace--selector' '$.TenantCode' $KeyVault
setConfig 'event--26--webhookconfig--webhookrequestrules--2--source--replace--ordercode' '$.OrderCode' $KeyVault
```

### DIFF
--- a/src/CaptainHook.Cli/Common/ShouldSerializeContractResolver.cs
+++ b/src/CaptainHook.Cli/Common/ShouldSerializeContractResolver.cs
@@ -1,4 +1,5 @@
 ﻿using CaptainHook.Common.Authentication;
+using CaptainHook.Common.Configuration;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Serialization;
 using System;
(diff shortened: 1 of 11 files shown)


src/CaptainHook.Cli/Common/ShouldSerializeContractResolver.cs
```diff
@@ -20,7 +21,9 @@ protected override JsonProperty CreateProperty(MemberInfo member, MemberSerializ
             JsonProperty property = base.CreateProperty(member, memberSerialization);
 
             // don't serialize empty collections
-            if (property.PropertyType != typeof(string) && property.PropertyType.GetInterface(nameof(IEnumerable)) != null)
+            if (property.PropertyType != typeof(string)
```
dawidkeshop, line 24: I think would be good to move this long condition to a separate method with a name telling us what kind of types we are looking for.

mszymeshop: @dawidkeshop I have changed the code slightly to simplify the understanding of the codebase. I hope this makes more sense now. Let me know!

dawidkeshop: Ok, now it looks better for me.

PR #140, opened 2020-07-13
#75635 - add request builder for Route and Replace

mszymeshop:
### What
Introduces a backbone to implement the logic for request builder where RouteAndReplace mechanism is going to be used. Logic itself is yet to be implemented.

### Classes
1. DefaultRequestBuilder (renamed from RequestBuilder) is the previous implementation of the builder
2. RequestBuilderDispatcher decides which builder to use
3. RouteAndReplaceRequestBuilder is the backbone for the new request builder for Route and Replace. Yet to be implemented fully.

### DIFF
--- a/src/CaptainHook.EventHandlerActor/Handlers/Requests/DefaultRequestBuilder.cs
+++ b/src/CaptainHook.EventHandlerActor/Handlers/Requests/DefaultRequestBuilder.cs
@@ -10,13 +10,13 @@ using Eshopworld.Platform.Messages;
 using Newtonsoft.Json;
 using Newtonsoft.Json.Linq;
 
-namespace CaptainHook.EventHandlerActor.Handlers
+namespace CaptainHook.EventHandlerActor.Handlers.Requests
 {
-    public class RequestBuilder : IRequestBuilder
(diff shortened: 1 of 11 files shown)


src/CaptainHook.EventHandlerActor/Handlers/Requests/DefaultRequestBuilder.cs
```diff
@@ -10,13 +10,13 @@
...
 {
-    public class RequestBuilder : IRequestBuilder
+    public class DefaultRequestBuilder : IRequestBuilder
     {
         private readonly IBigBrother bb;
 
-        public RequestBuilder(IBigBrother _bb)
+        public DefaultRequestBuilder(IBigBrother _bb)
         {
             bb = _bb;
```
dawidkeshop, line 21: I think the convention is to start the name of a field with `_` prefix, not a constructor parameter.

mszymeshop: I am just renaming the constructor here. But will change this for you :)

PR #142, opened 2020-07-14
Task 75635 route and replace logic implementation

mszymeshop:
### What
Implement the logic for Route and Replace and test with unit tests

### Disclaimer
Some comments in the code (especially around exceptions and validation) are going to be addressed in a separate task which is going to address the validation aspect.

### DIFF
--- a/src/CaptainHook.EventHandlerActor/Handlers/Requests/RouteAndReplaceRequestBuilder.cs
+++ b/src/CaptainHook.EventHandlerActor/Handlers/Requests/RouteAndReplaceRequestBuilder.cs
@@ -61,7 +61,8 @@ namespace CaptainHook.EventHandlerActor.Handlers.Requests
                     var valueFromPayload = RetrieveValueFromPayload(value);
                     if (!string.IsNullOrEmpty(valueFromPayload))
                     {
-                        yield return new KeyValuePair<string, string>(key, valueFromPayload);
+                        var escapedValue = Uri.EscapeDataString(valueFromPayload);
+                        yield return new KeyValuePair<string, string>(key, escapedValue);
                     }
                 }
             }
--- a/src/CaptainHook.EventHandlerActor/Handlers/Requests/RouteAndReplaceRequestBuilder.cs
+++ b/src/CaptainHook.EventHandlerActor/Handlers/Requests/RouteAndReplaceRequestBuilder.cs
@@ -1,5 +1,7 @@
 ﻿using System;
+using System.Collections.Generic;
 using System.Linq;
+using CaptainHook.Common;
 using CaptainHook.Common.Configuration;
 using Eshopworld.Core;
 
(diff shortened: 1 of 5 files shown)


src/Tests/CaptainHook.Tests/Services/Actors/Requests/RouteAndReplaceRequestBuilderTests.cs
```diff
@@ -0,0 +1,161 @@
...
+                                }
+                            }
+                        }
+                    },
+                    "{\"OrderCode\":\"9744b831-df2c-4d59-9d9d-691f4121f73a\", \"BrandType\":\"Brand2\", \"TenantCode\":\"tenant1\"}",
+                    "https://blah.blah.tenant1.eshopworld.com/webhook/9744b831-df2c-4d59-9d9d-691f4121f73a"
+                },
+                new object[]
+                {
+                    new WebhookConfig
```
chomick, line 81: I think we have builders for this data, tests looks better if we can use them

src/CaptainHook.EventHandlerActor/Handlers/Requests/RouteAndReplaceRequestBuilder.cs
```diff
@@ -1,25 +1,86 @@
...
+
         public RouteAndReplaceRequestBuilder(IBigBrother bigBrother): base(bigBrother)
         {
         }
 
         protected override Uri BuildUriFromExistingConfig(WebhookConfig config, string payload)
         {
-            var routeAndReplaceRule = config.WebhookRequestRules.FirstOrDefault(r => r.Destination.RuleAction == RuleAction.RouteAndReplace);
-            if (routeAndReplaceRule != null)
+            var routeAndReplaceRule = config.WebhookRequestRules.First(r => r.Destination.RuleAction == RuleAction.RouteAndReplace);
```
chomick, line 22: Are we expecting exception here at times? Should we take it safer and log an error with more details before we throw an exception?

src/CaptainHook.EventHandlerActor/Handlers/Requests/RouteAndReplaceRequestBuilder.cs
```diff
@@ -1,25 +1,86 @@
...
+
+            var selectorSource = routeAndReplaceRule.Source.Replace[SelectorKeyName];
```
chomick, line 24: I get that we to route & replace here - eventually, this will have to be replaced by simple 'replace' as URL selection will be something independent, right?

src/CaptainHook.EventHandlerActor/Handlers/Requests/RouteAndReplaceRequestBuilder.cs
```diff
@@ -1,25 +1,86 @@
...
+            return new Dictionary<string, string>(RetrieveReplacements());
+        }
+
+        private string GetUri(WebhookRequestRule rule, string selector)
+        {
+            var route = rule.Routes.FirstOrDefault(r => r.Selector.Equals(selector, StringComparison.OrdinalIgnoreCase));
+            if (route == null)
+            {
+                var defaultRoute =
+                    rule.Routes.First(r => r.Selector.Equals(DefaultFallbackSelectorKey, StringComparison.OrdinalIgnoreCase));
```
chomick, line 79: What if it is not there? Do we prefer to log something cleaner (e.g. what value were we looking, for what event, etc) before we bail-out?

src/Tests/CaptainHook.Tests/Services/Actors/Requests/RouteAndReplaceRequestBuilderTests.cs
dawidkeshop: @chomick:  we have builders for `SubscriberConfiguration` not for `WebhookConfig`. I'm going to add some as part of my task.

mszymeshop: Good plan, as soon as we have this (maybe it is something for @dawidkeshop or we can join efforts) we can change this code as well.

src/CaptainHook.EventHandlerActor/Handlers/Requests/RouteAndReplaceRequestBuilder.cs
mszymeshop: we will not, I assume this will be be pre-checked by validation (it is also something that is verified in `RequestBuilderDispatcher`)

src/CaptainHook.EventHandlerActor/Handlers/Requests/RouteAndReplaceRequestBuilder.cs
mszymeshop: URL selection is here as well, as we agreed to handle specific selectors first and fallback to * if none is found.

src/CaptainHook.EventHandlerActor/Handlers/Requests/RouteAndReplaceRequestBuilder.cs
mszymeshop: This code assumes validation has been done as a precondition for request builder. There will be a separate PR for this.

mszymeshop:
@chomick I am going to merge this since it has been accepted. If you want to discuss some of the answers to your remarks, please reach out.

PR #148, opened 2020-07-17
Added validation rule and test for at least one default selector

dawidkeshop:


src/CaptainHook.EventHandlerActor/Validation/WebhookRequestRuleForRouteAndReplaceValidator.cs
```diff
@@ -13,13 +13,24 @@ public WebhookRequestRuleForRouteAndReplaceValidator()
...
+            RuleFor(x => x.Routes).Must(ContainValidSelectors);
             RuleForEach(x => x.Routes).SetValidator((x, y) => new RouteValidator(x));
         }
 
-        private bool HaveUniqueSelectors(List<WebhookConfigRoute> routes)
+        private bool ContainValidSelectors(List<WebhookConfigRoute> routes)
         {
-            return routes?.Select(x => x.Selector).Distinct().Count() == routes?.Count;
+            var selectors = routes?.Select(r => r.Selector).ToList();
+            return ContainExactlyOneDefaultSelector(selectors) && AllSelectorsMustBeUnique(selectors);
```
chomick, line 23: Do we always have to have default selector?

mszymeshop: With Route And Replace - yes. With Route no. That is the assumption we have got in the logic code. The goal was not to create an algorithm which duplicates Route.

PR #150, opened 2020-07-18
Use DefaultAzureCredential for Azure auth for KV loading

tabman83:
### What
With the new Azure SDK, let's switch to DefaultAzureCredential for Azure authentication instead of the custom AzureServiceTokenCredential implementation.

### Why
The DefaultAzureCredential class provides a default TokenCredential authentication flow for applications that will be deployed to Azure. The following credential types if enabled will be tried, in order:
- EnvironmentCredential
- ManagedIdentityCredential
- SharedTokenCacheCredential
- InteractiveBrowserCredential

### Reference
- https://docs.microsoft.com/en-us/dotnet/api/azure.identity.defaultazurecredential?view=azure-dotnet
- https://www.rahulpnath.com/blog/defaultazurecredential_from_azure_sdk/

### Notes
In your local environment, DefaultAzureCredential uses the shared token credential from the IDE. In the case of Visual Studio, please make sure the account to use is properly configured under Options -> Azure Service Authentication.

### DIFF
--- a/src/CaptainHook.Common/Configuration/KeyVault/KeyVaultModule.cs
+++ b/src/CaptainHook.Common/Configuration/KeyVault/KeyVaultModule.cs
@@ -3,26 +3,13 @@ using System.Threading;
 using System.Threading.Tasks;
 using Autofac;
 using Azure.Core;
+using Azure.Identity;
 using Azure.Security.KeyVault.Secrets;
 using Microsoft.Azure.Services.AppAuthentication;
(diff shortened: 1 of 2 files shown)


mszymeshop:
@tabman83 are you sure this works locally and in CI? I remember trying the same class some weeks ago and at the time it did not work for me (having said that I might have misconfigured things).

chomick:
@tabman83 - is this PR still needed?

tabman83:
Closing for now as it is not needed.

PR #152, opened 2020-07-21
#76794 - remove placeholders for http client caching

mszymeshop:
### What
Placeholder in the host broke the algorithm. Its presence in other parts is accepted, however. This is a blunt fix as I wasn't able to achieve a design I originally wanted without changing the implementation extensively.

### DIFF
--- a/src/CaptainHook.EventHandlerActor/Handlers/HttpClientFactory.cs
+++ b/src/CaptainHook.EventHandlerActor/Handlers/HttpClientFactory.cs
@@ -38,7 +38,8 @@ namespace CaptainHook.EventHandlerActor.Handlers
         {
             if (config == null) throw new ArgumentNullException(nameof(config));
 
-            var uri = new Uri(config.Uri);
+            var uriString = config.Uri?.Replace("{", string.Empty).Replace("}", string.Empty);
+            var uri = new Uri(uriString);
 
             if (_httpClients.TryGetValue(uri.Host.ToLowerInvariant(), out var httpClient))
             {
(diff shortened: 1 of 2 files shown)


src/CaptainHook.EventHandlerActor/Handlers/HttpClientFactory.cs
```diff
@@ -38,7 +38,8 @@ public HttpClient Get(WebhookConfig config)
         {
             if (config == null) throw new ArgumentNullException(nameof(config));
 
-            var uri = new Uri(config.Uri);
+            var uriString = config.Uri?.Replace("{", string.Empty).Replace("}", string.Empty);
```
chomick, line 41: Surely this is not correct? The URL in question will not be a valid URL if we do not replace the variable but just remove the brackets?

mszymeshop: discussed offline, this is just to have a key for collection of HTTP clients

chomick: Even though you use it only for the key, you will end up with a single http client that is destined for different hosts afterwards. Is that going to be a problem? What's the reason we index the clients by the host?

mszymeshop: The way I see it is to keep as few HTTP clients as possible at the same time. Also, the timeout settings is per entry so they might differ between entries. Both will have the same impact when we use replace functionality. If we wanted to do this for individual hosts I would need to change more code (due to backwards-compatibility we want to keep).

I think that we want to minimise the number of HTTP clients which share the same setup. This means those that come from the template/replace uri with timeout could be reused for all incoming tenants (even between subscribers as long as the hosts match), which sounds reasonable to me.

tabman83: I am approving it for now as I don't see a quick way for a clean fix but I suggest we create a tech debt to track it.

chomick: This safe for now as sharing host for a single http client instance is safe.

PR #153, opened 2020-07-23
77046 API endpoint for adding subscriber

dawidkeshop:


src/CaptainHook.Api/Controllers/EventsController.cs
```diff
@@ -0,0 +1,43 @@
...
+        }
+
+        [HttpPut]
+        [Route("/{eventName}/subscriber/{subscriberName}")]
+        [ProducesResponseType(StatusCodes.Status200OK)]
+        [ProducesResponseType(StatusCodes.Status400BadRequest)]
+        [ProducesResponseType(StatusCodes.Status401Unauthorized)]
+        [ProducesResponseType(StatusCodes.Status403Forbidden)]
+        [ProducesResponseType(StatusCodes.Status409Conflict)]
+        public async Task<IActionResult> PutSubscriber([FromRoute] string eventName, [FromRoute] string subscriberName, [FromBody] SubscriberDto dto)
```
chomick, line 35: The story is not about subscriber - it is about adding a single webhook to a subscriber - we just expect the subscriber to be added in this case.

src/CaptainHook.Api/Controllers/EventsController.cs
```diff
@@ -0,0 +1,43 @@
...
+        private readonly IBigBrother _bigBrother;
+
+        public EventsController(IMediator mediator, IBigBrother bigBrother)
+        {
+            _mediator = mediator ?? throw new ArgumentNullException(nameof(mediator));
+            _bigBrother = bigBrother ?? throw new ArgumentNullException(nameof(bigBrother));
+        }
+
+        [HttpPut]
+        [Route("/{eventName}/subscriber/{subscriberName}")]
```
chomick, line 29: This is the endpoint I was expecting for the story:
PUT /event/{name-of-event}/subscriber/{name-of-subscriber}/webhooks/endpoint/{selector}

where 'selector' is optional and could be excluded for now as we do not support it anyway yet.

src/CaptainHook.Contract/WebhooksDto.cs
```diff
@@ -5,7 +5,6 @@ namespace CaptainHook.Contract
     public class WebhooksDto
     {
         public string SelectionRule { get; set; }
```
chomick, line 7: This is not needed/defined yet

src/CaptainHook.Contract/SubscriberDto.cs
```diff
@@ -3,9 +3,6 @@
     public class SubscriberDto
     {
         public string Name { get; set; }
```
chomick, line 5: That's a duplicate from the URL - should not be part of this DTO

src/CaptainHook.Contract/EndpointDto.cs
```diff
@@ -3,11 +3,8 @@
     public class EndpointDto
     {
         public string Uri { get; set; }
-
         public string HttpVerb { get; set; }
-
         public string Selector { get; set; }
```
chomick, line 7: We do not support it yet - could be removed for now?

src/CaptainHook.Contract/AuthenticationDto.cs
```diff
@@ -5,9 +5,8 @@ namespace CaptainHook.Contract
     public class AuthenticationDto
     {
         public string Type { get; set; }
-
-        public string Ref { get; set; }
```
chomick, line 9: And how will you pass the secret reference? Maybe we just need to name it properly but it has to be part of this definition.

chomick: So to confirm - the secret reference here should point to a name in our KV

src/CaptainHook.Application/Requests/Subscribers/UpsertSubscriberRequest.cs
```diff
@@ -0,0 +1,21 @@
...
+using CaptainHook.Domain.Results;
+using MediatR;
+
+namespace CaptainHook.Application.Requests.Subscribers
+{
+    public class UpsertSubscriberRequest : IRequest<OperationResult<Guid>>
+    {
+        public string EventName { get; set; }
+        public string SubscriberName { get; set; }
+        public SubscriberDto SubscriberDto { get; set; }
```
chomick, line 12: So we are using DTOs in this layer, right?

src/CaptainHook.Api/Controllers/EventsController.cs
```diff
@@ -0,0 +1,43 @@
...
+        [ProducesResponseType(StatusCodes.Status200OK)]
```
chomick, line 30: Can you specify the response type for swagger to render for each code, please?

src/CaptainHook.Contract/AuthenticationDto.cs
dawidkeshop: As discussed with @mszymeshop it should not be a reference to a KV directly, it should be "CH authentication reference".  Since this is not supported yet, I've updated the code with new DTO which points to KV and secret name.

src/CaptainHook.Application/Requests/Subscribers/UpsertSubscriberRequest.cs
dawidkeshop: Yes, Application layer knows DTOs, validates these and maps to Domain models if needed.

src/CaptainHook.Api/Controllers/EventsController.cs
dawidkeshop: I've added but as for now I used errors from Domain layer. It's not a valid design, we should have a `ErrorDto` class in Contracts and map from business errors to this ErrorDto at application layer. I'll implement this improvement as part of my current task which is about Application layer.

chomick: We do not return any details for OK response, do we? Maybe in that case we should return no-content one?

src/CaptainHook.Api/Helpers/FakeAuthenticationMiddleware.cs
```diff
@@ -44,7 +44,8 @@ public async Task Invoke(HttpContext context)
 
                 if (claims.Count == 0)
                 {
-                    claims.Add(new Claim("scope", "tooling.eda.api.all"));
+                    claims.Add(new Claim("scope", Constants.AuthorisationScopes.ApiAllAccess));
+                    claims.Add(new Claim("scope", Constants.AuthorisationScopes.ApiDefineSubscribers));
                 }
 
                 var identity = new ClaimsIdentity(new GenericIdentity("FakeUser", "FakeAuth"), claims);
```
chomick, line 51: Why do we need this fake user thing?

src/CaptainHook.Api/Controllers/EventsController.cs
chomick: Even better, maybe we should return the full subscriber in the response?

src/CaptainHook.Application/Handlers/Subscribers/GetSubscribersForEventQueryHandler.cs
```diff
@@ -28,7 +28,7 @@ public async Task<OperationResult<List<SubscriberDto>>> Handle(GetSubscribersFor
             }
 
             var subscriberEntities = await _repository.GetSubscribersListAsync(query.Name);
```
chomick, line 30: I know this is temp code (?) and I suspect we will have the single-subscriber retrieval?

src/CaptainHook.Contract/AuthenticationDto.cs
chomick: It's fine but I think we will not be accessing other KVs for now - happy to keep it as is, though.

src/CaptainHook.Application/Handlers/Subscribers/GetSubscribersForEventQueryHandler.cs
dawidkeshop: This was introduced as sample code in the spike done last sprint, I'll remove it now.

src/CaptainHook.Api/Helpers/FakeAuthenticationMiddleware.cs
dawidkeshop: It's a matter of convenience. It allows us to use API on local system without token.

src/CaptainHook.Application/Handlers/Subscribers/GetSubscribersForEventQueryHandler.cs
chomick: Ah, sorry, I can see this is for a different handler.

src/CaptainHook.Api/Controllers/EventsController.cs
```diff
@@ -0,0 +1,44 @@
...
+        private readonly IBigBrother _bigBrother;
+
+        public EventsController(IMediator mediator, IBigBrother bigBrother)
+        {
+            _mediator = mediator ?? throw new ArgumentNullException(nameof(mediator));
+            _bigBrother = bigBrother ?? throw new ArgumentNullException(nameof(bigBrother));
+        }
+
+        [HttpPut]
+        [Route("/{eventName}/subscriber/{subscriberName}/webhooks/endpoint/")]
```
chomick, line 30: I think this may make it a root and skip 'api/event' string from the URL?

dawidkeshop: You're right, fixed.

src/CaptainHook.Api/Controllers/EventsController.cs
dawidkeshop: Since it's an upsert operation, ideally we should distinct two possible effects and return:
1. 201 (created) when we create a new webhook
2. 204 (no content) when we update a existing webhook

And use a `Location` response header in both cases to return an URI which an API consumer can use to query about resource, URI like: `api/events/[eventName]/subscriber/{subscriberName}`.

Of course, we can use also return a Subscriber data in the response (in that case I would still go with 201 for newly-created item and 200 in case of update) if we think that it gives profit to the consumer. Would that data be required immediately in the most cases? If it's not going to be used, maybe better not to do that and save some bandwidth?

PR #154, opened 2020-07-23
#77383 - include http verb for route and replace

mszymeshop:
### What
Route and Replace incorrectly used HTTP verb for its requests. The current PR changes that by using the HTTP verb defined for the particular endpoint.

### DIFF
--- a/src/CaptainHook.EventHandlerActor/Handlers/Requests/DefaultRequestBuilder.cs
+++ b/src/CaptainHook.EventHandlerActor/Handlers/Requests/DefaultRequestBuilder.cs
@@ -179,7 +179,7 @@ namespace CaptainHook.EventHandlerActor.Handlers.Requests
 
         /// <inheritdoc />
 
-        public HttpMethod SelectHttpMethod(WebhookConfig webhookConfig, string payload)
+        public virtual HttpMethod SelectHttpMethod(WebhookConfig webhookConfig, string payload)
         {
             if (webhookConfig == null) throw new ArgumentNullException(nameof(webhookConfig));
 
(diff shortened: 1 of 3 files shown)


src/CaptainHook.EventHandlerActor/Handlers/Requests/DefaultRequestBuilder.cs
```diff
@@ -179,7 +179,7 @@ public string BuildPayload(WebhookConfig config, string sourcePayload, IDictiona
 
         /// <inheritdoc />
 
-        public HttpMethod SelectHttpMethod(WebhookConfig webhookConfig, string payload)
+        public virtual HttpMethod SelectHttpMethod(WebhookConfig webhookConfig, string payload)
```
chomick, line 182: Have you checked for all other usages of RuleAction.Route and if it should consider the new the as well?

src/CaptainHook.EventHandlerActor/Handlers/Requests/RouteAndReplaceRequestBuilder.cs
```diff
@@ -55,6 +56,11 @@ protected override WebhookConfig SelectWebhookConfigCore(WebhookConfig webhookCo
             return routeConfig;
         }
 
+        public override HttpMethod SelectHttpMethod(WebhookConfig webhookConfig, string payload) =>
+            SelectWebhookConfig(webhookConfig, payload).HttpMethod;
```
chomick, line 60: Should we not simply replace the implementation of the SelectHttpMethod in base class to use virtual SelectWebhookConfig?

mszymeshop: Unfortunately, they differ slightly + there is code duplication in DefaultRequestBuilder which requires additional checks and tests. Nothing major, but the ultimate goal was not to touch DefaultRequestBuilder too much for its core role in the product. Something we can definitely improve and refactor in the future but I didn't feel I have time to take care of this at the moment.

src/CaptainHook.EventHandlerActor/Handlers/Requests/DefaultRequestBuilder.cs
mszymeshop: As far as IRequestBuilder is concerned, I believe I have yes. 

> if it should consider the new the as well
Could you say more about this?

PR #155, opened 2020-07-24
77048 upsert Webhook validation

dawidkeshop:
This is not the whole application layer, it's only input validation and tests for it. 

I've also created new project `CaptainHook.TestsInfrastructure` which contains builders, assertion helpers, reusable test data attributes and so on. All of that can be reusable in API layer tests as well as old tests (director, readers, actors).


src/CaptainHook.Application/Validators/Common/HttpVerbValidator.cs
```diff
@@ -0,0 +1,21 @@
+﻿using System.Linq;
+using FluentValidation.Validators;
+
+namespace CaptainHook.Application.Validators.Common
+{
+    public class HttpVerbValidator : PropertyValidator
+    {
+        private readonly string[] _validVerbs = { "POST", "PUT", "GET" };
```
tabman83, line 8: static maybe?

src/CaptainHook.Application/Validators/Common/HttpVerbValidator.cs
```diff
@@ -0,0 +1,21 @@
...
+
+        public HttpVerbValidator()
+            : base("{PropertyName} must be valid HTTP verb.")
+        {
+        }
+
+        protected override bool IsValid(PropertyValidatorContext context)
+        {
+            var httpVerb = context.PropertyValue as string;
+            return _validVerbs.Contains(httpVerb);
```
tabman83, line 18: will consider case insensitive compare

src/CaptainHook.Application/Validators/Common/UriValidator.cs
```diff
@@ -0,0 +1,19 @@
...
+    {
+        public UriValidator()
+            : base("{PropertyName} must be valid URI.")
+        {
+        }
+
+        protected override bool IsValid(PropertyValidatorContext context)
+        {
+            var rawValue = context.PropertyValue as string;
+            return Uri.TryCreate(rawValue, UriKind.Absolute, out Uri _);
```
tabman83, line 16: how does this play out with RouteAndReplace where the Uri is not a valid one due to tokens

src/CaptainHook.Application/Validators/Dtos/AuthenticationDtoValidator.cs
```diff
@@ -0,0 +1,31 @@
...
+
+namespace CaptainHook.Application.Validators.Dtos
+{
+    public class AuthenticationDtoValidator : AbstractValidator<AuthenticationDto>
+    {
+        public AuthenticationDtoValidator()
+        {
+            CascadeMode = CascadeMode.StopOnFirstFailure;
+
+            RuleFor(x => x.Type).NotEmpty().Equal("OIDC");
```
tabman83, line 13: Case insensitive compare?

src/CaptainHook.Application/Validators/UpsertWebhookRequestValidator.cs
```diff
@@ -0,0 +1,19 @@
...
+using CaptainHook.Application.Validators.Dtos;
+using FluentValidation;
+
+namespace CaptainHook.Application.Validators
+{
+    public class UpsertWebhookRequestValidator : AbstractValidator<UpsertWebhookRequest>
+    {
+        public UpsertWebhookRequestValidator()
+        {
+            CascadeMode = CascadeMode.StopOnFirstFailure;
```
tabman83, line 11: This StopOnFirstFailure. Does it mean that we return as soon as we encounter the first error? We may want to collect all errors and report back to caller otherwise it's gonna be a trial and error approach for him

src/CaptainHook.Application/Validators/Common/UriValidator.cs
chomick: It will not play :) that was the issue Michal had.
For now, as we do not support URI transforms in this story, we can leave it.
But in the next story, we will need additional validations and maybe different approach to validate the URI itself?
There will be a few checks we will need to do, in particular:
- are all vars defined
- is selector included in vars (temporarily we need to enforce it)
- etc

src/CaptainHook.Application/Validators/Dtos/AuthenticationDtoValidator.cs
```diff
@@ -0,0 +1,31 @@
...
+            RuleFor(x => x.ClientId).NotEmpty();
+            RuleFor(x => x.Uri).NotEmpty().SetValidator(new UriValidator());
```
chomick, line 15: That's neat!

src/CaptainHook.Application/Validators/Dtos/AuthenticationDtoValidator.cs
```diff
@@ -0,0 +1,31 @@
...
+            RuleFor(x => x.ClientSecret).NotNull()
+                .SetValidator(new ClientSecretDtoValidator());
+        }
+
+        private class ClientSecretDtoValidator : AbstractValidator<ClientSecretDto>
+        {
+            public ClientSecretDtoValidator()
+            {
+                RuleFor(x => x.Name).NotEmpty();
+                RuleFor(x => x.Vault).NotEmpty();
```
chomick, line 27: As I suggested before, not sure what we are planning to pass here? The KV URL?
I suggest we just pass the secret name for now and the vault itself is the only one CH is using - we may not need ability to support multiple vaults - when we know we have to, then we can define it properly.

src/CaptainHook.Application/Validators/Dtos/EndpointDtoValidator.cs
```diff
@@ -0,0 +1,21 @@
...
+    public class EndpointDtoValidator : AbstractValidator<EndpointDto>
+    {
+        public EndpointDtoValidator()
+        {
+            CascadeMode = CascadeMode.StopOnFirstFailure;
+
+            RuleFor(x => x.HttpVerb).NotEmpty()
+                .SetValidator(new HttpVerbValidator());
+            RuleFor(x => x.Uri).NotEmpty()
+                .SetValidator(new UriValidator());
```
chomick, line 16: Yes, this will change for next story.

I really like the validators approach - I must say was bit more sceptic but this is really nice way of handling it!

src/CaptainHook.Application/Validators/UpsertWebhookRequestValidator.cs
chomick: I am not sure - if some data is invalid, other validations may be false-positives.. I think we are ok to return on first error tbh

src/Tests/CaptainHook.Application.Tests/RequestValidators/UpsertSubscriberRequestValidatorTests.cs
```diff
@@ -0,0 +1,111 @@
...
+            var result = _validator.Validate(request);
+
+            result.IsValid.Should().BeTrue();
+        }
+
+        [Theory, IsUnit]
+        [InlineData("POST")]
+        [InlineData("GET")]
+        [InlineData("PUT")]
+        public void When_HttpVerb_is_valid_HTTP_verb_then_then_no_failures_should_be_returned(string httpVerb)
```
chomick, line 31: Do you have a case for other incorrect string input here?

src/CaptainHook.Application/Validators/Common/HttpVerbValidator.cs
dawidkeshop: Of course!

src/Tests/CaptainHook.Application.Tests/RequestValidators/UpsertSubscriberRequestValidatorTests.cs
tabman83: Why not test also lowercase, mixed case. Just in case.

src/CaptainHook.Application/Validators/Common/HttpVerbValidator.cs
dawidkeshop: Sure

src/CaptainHook.Application/Validators/Common/UriValidator.cs
dawidkeshop: @chomick: we already have the logic which you described in `SourceParserLocationValidator` in EventHandlerActor project, so we should extract and reuse it.

src/CaptainHook.Application/Validators/Dtos/AuthenticationDtoValidator.cs
dawidkeshop: Sure

src/CaptainHook.Application/Validators/UpsertWebhookRequestValidator.cs
dawidkeshop: It stops processing when it encounters first error for particular property. For example if we have such configuration:
```
RuleFor(x => x.Uri).NotEmpty().SetValidator(new UriValidator());
```
then if `Uri` passed in request would be an empty string, then `UriValidatior` won't be called at all. And it makes sense, because if string is null or empty, it's for sure not a valid uri, so no need to check that.

From response point of view we will have only one error for one field. But if multiple fields in request will be invalid, we will see error for every of these.

src/CaptainHook.Application/Validators/Dtos/AuthenticationDtoValidator.cs
dawidkeshop: Not KV uri, the KV name as described by Michal in wiki. I think we all should discuss that, because I see that you both have different visions of how it should work.

src/Tests/CaptainHook.Application.Tests/RequestValidators/UpsertSubscriberRequestValidatorTests.cs
dawidkeshop: Yes, there is a test: 
```
[Theory, IsUnit]
[InlineData("duck")]
public void When_HttpVerb_is_not_a_valid_HTTP_verb_then_validation_should_fail(string httpVerb)
```

dawidkeshop: Yes, there is a test: 
```[Theory, IsUnit]
[InlineData("duck")]
public void When_HttpVerb_is_not_a_valid_HTTP_verb_then_validation_should_fail(string httpVerb)```

Comment (dawidkeshop) [reply src/Tests/CaptainHook.Application.Tests/RequestValidators/UpsertSubscriberRequestValidatorTests.cs]: @tabman83: I've just added lowercase tests, I had to add lowercase comparison first.

File: src/CaptainHook.Application/Validators/Common/HttpVerbValidator.cs [hunk 10]
```diff
@@ -0,0 +1,21 @@
...
+using FluentValidation.Validators;
+
+namespace CaptainHook.Application.Validators.Common
+{
+    public class HttpVerbValidator : PropertyValidator
+    {
+        private static readonly string[] _validVerbs = { "POST", "PUT", "GET" };
+
+        public HttpVerbValidator()
+            : base("{PropertyName} must be valid HTTP verb.")
```
mszymeshop, line 11: I am worried here that the error message is not entirely clear. It has to be a valid HTTP verb within the range we allow. Maybe worth mentioning the ones we allow?

src/CaptainHook.Application/Validators/Dtos/AuthenticationDtoValidator.cs
mszymeshop: Vault is the name of the Key Vault, Name is the name of the Secret. It can be addressed with a single KV Id, but I think having these two separate actually simplifies things.

src/Tests/CaptainHook.Application.Tests/RequestValidators/AuthenticationDtoValidatorTests.cs
```diff
@@ -0,0 +1,149 @@
...
+using Xunit;
+
+namespace CaptainHook.Application.Tests.RequestValidators
+{
+    public class AuthenticationDtoValidatorTests
+    {
+        private readonly AuthenticationDtoValidator _validator = new AuthenticationDtoValidator();
+
+        [Fact, IsUnit]
+        public void When_request_is_valid_then_no_failures_should_be_returned()
```
mszymeshop, line 18: The only worry I have is the format of the test names. It is the 3rd or 4th we are introducing. Can we agree as a team where we want to go with this?

A quick search gave me this breakdown: https://dzone.com/articles/7-popular-unit-test-naming. Should we do a quick team wide session to discuss this?

src/Tests/CaptainHook.TestsInfrastructure/Builders/AuthenticationDtoBuilder.cs
```diff
@@ -0,0 +1,17 @@
...
+namespace CaptainHook.TestsInfrastructure.Builders
+{
+    public class AuthenticationDtoBuilder : SimpleBuilder<AuthenticationDto>
+    {
+        public AuthenticationDtoBuilder()
+        {
+            With(x => x.Type, "OIDC");
+            With(x => x.ClientId, "clientId");
+            With(x => x.Uri, "https://security-api.com/token");
+            With(x => x.Scopes, new List<string>(new[] { "test.scope.api" }));
```
mszymeshop, line 13: Do you need a new List from new Array here?

src/Tests/CaptainHook.TestsInfrastructure/TestsData/EmptyStrings.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System.Collections;
+using System.Collections.Generic;
+
+namespace CaptainHook.TestsInfrastructure.TestsData
+{
+    public class EmptyStrings : IEnumerable<object[]>
```
mszymeshop, line 6: 💯

src/Tests/CaptainHook.TestsInfrastructure/TestsData/InvalidUris.cs
```diff
@@ -0,0 +1,17 @@
...
+
+namespace CaptainHook.TestsInfrastructure.TestsData
+{
+    public class InvalidUris : IEnumerable<object[]>
+    {
+        public IEnumerator<object[]> GetEnumerator()
+        {
+            yield return new object[] { "not-a-uri" };
+            yield return new object[] { "https:/invalid.com" };
+            yield return new object[] { "https://.com" };
```
mszymeshop, line 12: I am just thinking, should we add a url with template in the host? This one is invalid as well at the moment. (at least as an Uri).

src/CaptainHook.Application/Validators/Common/HttpVerbValidator.cs
dawidkeshop: Sure

src/Tests/CaptainHook.TestsInfrastructure/Builders/AuthenticationDtoBuilder.cs
dawidkeshop: It wasn't me, it was Resharper! 
Fixed

src/Tests/CaptainHook.TestsInfrastructure/TestsData/InvalidUris.cs
dawidkeshop: Sure

src/CaptainHook.Application/Validators/Common/UriValidator.cs
chomick: True- this is not part of this story, though

src/CaptainHook.Application/Validators/Dtos/AuthenticationDtoValidator.cs
chomick: Do we want to have the separation? I know that we discussed originally that approach but for now we do not support multiple KV's and it will add extra complexity that we may not need right now. Maybe we will never have to access more than one KV.
I suggest we only pass the key name and assume the only KV we currently recognise.

src/Tests/CaptainHook.Application.Tests/RequestValidators/AuthenticationDtoValidatorTests.cs
dawidkeshop: Regarding first three conventions - I think starting all tests with `test` just makes no sense (is it some kind of Hungarian notation?) as well as including test method name. In one test suite we should test no more than one method, in other case we will break the single responsibility principle. Also, especially in cases of Handlers and Validators, we are forced to test particular method, so we even can't choose the name of the method to test.

Basically I'm using the 6th approach except I'm using `then` instead of `expect` and I'm writing it in a different way because:
1. Having words separated with underscores is just MuchMoreReadableThanAVeryLongTextWrittenInPascalCase.
1. Having an English sentence is more readable also for non-programmers, tired people, people before first two coffees in the morning, etc.
1. Because words are separated with underscores I don't need to use Pascal case, so I can use capital letters only when I want to. And I do that when I'm referring to properties, entities and other important things. So now I can see immediately what is test about, I see the test subject at the first sight, without necessity to search test name and parse all words one by one. Eg. `BuildUri_ExecutesRouteAndReplace_WhenRouteAndReplaceConfig` - does it mean that `BuildUri` executes something called `route` and then do replacement? Or just executes something called `RouteAndReplace`? I know the answer, but person who see this for the first time probably will have a doubts. My way allows to avoid having doubts.

It's of course just my _objective opinion_, if team thinks differently I would be happy to rename everything. Just let's decide for one convention, because as for me lack of convention is much bigger problem than the format itself.

src/Tests/CaptainHook.TestsInfrastructure/TestsData/InvalidUris.cs
tabman83: Try a relative Uri as well

src/Tests/CaptainHook.TestsInfrastructure/Builders/AuthenticationDtoBuilder.cs
tabman83: Ah, that useful tool that you guys wanted me to install!

src/Tests/CaptainHook.TestsInfrastructure/TestsData/InvalidUris.cs
dawidkeshop: Sure

src/CaptainHook.Application/Validators/Dtos/AuthenticationDtoValidator.cs
dawidkeshop: @tabman83: here we validate incoming data to give fast feedback to an API consumer if something is missing in the request. I think KV accessibility will be validated on a different level, probably during mapping from entity to old model.

PR #156, opened 2020-07-29
Task 77370 add appsettings to tests

nikhilgirraj:
This PR addresses adding AppSettings to the CaptainHook.Tests project. The project now has appsettings.json files for CI, Dev, Prep and Test. 
The KeyVaults do not need any additional values for this code to execute. 
Also, this code can be optimized to run faster once the package from https://github.com/eShopWorld/devops/pull/31 is available.


src/Tests/CaptainHook.Tests/Web/FlowTests/E2EFlowTestsFixture.cs
```diff
@@ -41,36 +38,15 @@ public E2EFlowTestsFixture()
...
-            var sbConnString = "";
-            var subId = "";
-            PeterPanUrlBase = "";
-            StsClientId = "";
-            StsClientSecret = "";
-#endif
-
-            Bb = BigBrother.CreateDefault(instrKey, instrKey);
-            Bb.PublishEventsToTopics(new Messenger(sbConnString, subId));
+            var config = new TestsConfig(); // loads from different KVs for Development and CI environment
```
tabman83, line 43: Use ConfigurationBinder.Bind instead,
https://docs.microsoft.com/en-us/dotnet/api/microsoft.extensions.configuration.configurationbinder.bind?view=dotnet-plat-ext-3.1

src/Tests/CaptainHook.Tests/appsettings.Development.json
```diff
@@ -0,0 +1,6 @@
+{
+  "KEYVAULT_URL": "https://esw-tooling-ci-we-nk.vault.azure.net/",
```
tabman83, line 2: I know it's your KV but you probably should use the CI one here, and then each dev can overwrite with his own one. Your KV may not be always available or contain what is needed

src/Tests/CaptainHook.Tests/Web/FlowTests/E2EFlowTestsFixture.cs
nikhilgirraj: I did not know about `ConfigurationBinder`. I feel it might be better to load variables using `ConfigurationBinder` inside the `TestsConfig` class instead? What do you say?

src/Tests/CaptainHook.Tests/appsettings.Development.json
nikhilgirraj: Makes sense. I'll change it.

src/Tests/CaptainHook.Tests/Web/FlowTests/E2EFlowTestsFixture.cs
chomick: they have to match the data structures - I think your names and structure does not match KV entries - but if you can make them in sync, that's better way for sure

src/Tests/CaptainHook.Tests/Web/FlowTests/E2EFlowTestsFixture.cs
```diff
@@ -41,36 +38,15 @@ public E2EFlowTestsFixture()
...
-
-            Bb = BigBrother.CreateDefault(instrKey, instrKey);
-            Bb.PublishEventsToTopics(new Messenger(sbConnString, subId));
+            var config = ConfigurationLoader.GetTestsConfig(); // loads from different KVs for Development and CI environment
+            this.PeterPanUrlBase = config.PeterPanBaseUrl;
+            this.StsClientId = config.StsClientId;
+            this.StsClientSecret = config.ApiSecret;
+
+            Bb = BigBrother.CreateDefault(config.InstrumentationKey, config.InstrumentationKey);
+            Bb.PublishEventsToTopics(new Messenger(config.ServiceBusConnectionString, config.AzureSubscriptionId));
```
tabman83, line 49: You are only loading config once, so I suggest you drop the static class, and perform the loading straight into the fixture constructor. Then your config POCO will hold the config which you can use straight into the methods who need it (BigBrother). So you can drop also the config properties on the fixture class. If you need config elsewhere, then you can make the whole config POCO a class property.

nikhilgirraj: I like the idea of using the POCO instead of the variables. I should have done that sooner. However, the ConfigurationLoader class seems like a good segregation of concerns to me. This keeps it open to any other tests that might need to use the values later. Do you agree?

tabman83: The DevOps package is already a config loading abstraction. I believe that having a wrapper around that is just gonna be another black box.

src/Tests/CaptainHook.Tests/Web/FlowTests/E2EFlowTestsFixture.cs
```diff
@@ -29,9 +26,7 @@ namespace CaptainHook.Tests.Web.FlowTests
     public class E2EFlowTestsFixture
     {
         public IBigBrother Bb;
-        public string PeterPanUrlBase { get; set; }
-        public string StsClientId { get; set; }
-        public string StsClientSecret { get; set; }
+        public TestsConfig TestsConfig { get; set; }
```
tabman83, line 29: Do we need this to be public?

src/Tests/CaptainHook.Tests/Web/FlowTests/E2EFlowTestsFixture.cs
nikhilgirraj: I see your point.

src/Tests/CaptainHook.Tests/Web/FlowTests/E2EFlowTestsFixture.cs
nikhilgirraj: Not yet. I simply kept the access level of the previous members.

nikhilgirraj: Changed it to private anyway.

src/Tests/CaptainHook.Tests/Web/FlowTests/E2EFlowTestsFixture.cs
```diff
@@ -29,9 +28,7 @@ namespace CaptainHook.Tests.Web.FlowTests
     public class E2EFlowTestsFixture
     {
         public IBigBrother Bb;
-        public string PeterPanUrlBase { get; set; }
-        public string StsClientId { get; set; }
-        public string StsClientSecret { get; set; }
+        private TestsConfig TestsConfig { get; set; }
```
tabman83, line 31: - No need for a property here, this can be a field.
- For private members please follow the convention, as seen below: `_camelCase`
- If you feel you can move the config loading inside the constructor (and you can probably do that as I don't foresee any other place that can invoke config loading) then you can make this field readonly.

nikhilgirraj: Done!

PR #157, opened 2020-07-30
77825 handler implementation

dawidkeshop:
Upsert WebHook Handler implementation and tests


src/CaptainHook.DirectorService/DirectorService.cs
```diff
@@ -22,7 +25,7 @@
 namespace CaptainHook.DirectorService
 {
     [UsedImplicitly]
-    public class DirectorService : StatefulService, IDirectorServiceRemoting
+    public class DirectorService : StatefulService, IDirectorServiceRemoting, IDirectorServiceGateway
```
tabman83, line 28: What is this gonna be?

src/CaptainHook.Domain/Errors/UnhandledExceptionError.cs
```diff
@@ -0,0 +1,16 @@
+﻿using System;
+using CaptainHook.Domain.Results;
+
+namespace CaptainHook.Domain.Errors
+{
+    public class UnhandledExceptionError : ErrorBase
+    {
+        private readonly Exception _exception;
+
+        public UnhandledExceptionError(string message, Exception exception)
```
tabman83, line 10: What's the use case for this?

src/Tests/CaptainHook.Application.Tests/Handlers/UpsertWebhookRequestHandlerTests.cs
```diff
@@ -0,0 +1,231 @@
...
+            _directorServiceMock.Setup(r => r.CreateReader(It.Is<SubscriberEntity>(rci => MatchReaderChangeInfo(rci, request))))
+                .ReturnsAsync(false);
+
+            var result = await Handler.Handle(request, CancellationToken.None);
+
+            result.IsError.Should().BeTrue();
+            result.Error.Should().BeOfType<BusinessError>();
+            _repositoryMock.Verify(x => x.GetSubscriberAsync(It.IsAny<SubscriberId>()), Times.Once);
+            _directorServiceMock.Verify(x => x.CreateReader(It.IsAny<SubscriberEntity>()), Times.Never);
+            _repositoryMock.Verify(x => x.SaveSubscriberAsync(It.IsAny<SubscriberEntity>()), Times.Never);
```
tabman83, line 75: Can use an assertion scope

src/Tests/CaptainHook.Application.Tests/Handlers/UpsertWebhookRequestHandlerTests.cs
```diff
@@ -0,0 +1,231 @@
...
```
tabman83, line 96: Can use an assertion scope

src/Tests/CaptainHook.Application.Tests/Handlers/UpsertWebhookRequestHandlerTests.cs
```diff
@@ -0,0 +1,231 @@
...
+            _directorServiceMock.Setup(r => r.CreateReader(It.Is<SubscriberEntity>(rci => MatchReaderChangeInfo(rci, request))))
+                .ReturnsAsync(false);
+
+            var result = await Handler.Handle(request, CancellationToken.None);
+
+            result.IsError.Should().BeTrue();
+            result.Error.Should().BeOfType<UnhandledExceptionError>();
+            _repositoryMock.Verify(x => x.GetSubscriberAsync(It.IsAny<SubscriberId>()), Times.Once);
+            _directorServiceMock.Verify(x => x.CreateReader(It.IsAny<SubscriberEntity>()), Times.Never);
+            _repositoryMock.Verify(x => x.SaveSubscriberAsync(It.IsAny<SubscriberEntity>()), Times.Never);
```
tabman83, line 117: Can use an assertion scope

src/Tests/CaptainHook.Application.Tests/Handlers/UpsertWebhookRequestHandlerTests.cs
```diff
@@ -0,0 +1,231 @@
...
+            _directorServiceMock.Setup(r => r.CreateReader(It.Is<SubscriberEntity>(rci => MatchReaderChangeInfo(rci, request))))
+                .ReturnsAsync(new DirectorServiceIsBusyError());
+
+            var result = await Handler.Handle(request, CancellationToken.None);
+
+            result.IsError.Should().BeTrue();
+            result.Error.Should().BeOfType<DirectorServiceIsBusyError>();
+            _repositoryMock.Verify(x => x.GetSubscriberAsync(It.IsAny<SubscriberId>()), Times.Once);
+            _directorServiceMock.Verify(x => x.CreateReader(It.IsAny<SubscriberEntity>()), Times.Once);
+            _repositoryMock.Verify(x => x.SaveSubscriberAsync(It.IsAny<SubscriberEntity>()), Times.Never);
```
tabman83, line 138: Can use an assertion scope

src/Tests/CaptainHook.Application.Tests/Handlers/UpsertWebhookRequestHandlerTests.cs
```diff
@@ -0,0 +1,231 @@
...
+            _directorServiceMock.Setup(r => r.CreateReader(It.Is<SubscriberEntity>(rci => MatchReaderChangeInfo(rci, request))))
+                .ReturnsAsync(new ReaderCreationError(new SubscriberEntity("subscriber")));
+
+            var result = await Handler.Handle(request, CancellationToken.None);
+
+            result.IsError.Should().BeTrue();
+            result.Error.Should().BeOfType<ReaderCreationError>();
+            _repositoryMock.Verify(x => x.GetSubscriberAsync(It.IsAny<SubscriberId>()), Times.Once);
+            _directorServiceMock.Verify(x => x.CreateReader(It.IsAny<SubscriberEntity>()), Times.Once);
+            _repositoryMock.Verify(x => x.SaveSubscriberAsync(It.IsAny<SubscriberEntity>()), Times.Never);
```
tabman83, line 159: Can use an assertion scope

src/Tests/CaptainHook.Application.Tests/Handlers/UpsertWebhookRequestHandlerTests.cs
```diff
@@ -0,0 +1,231 @@
...
+            _directorServiceMock.Setup(r => r.CreateReader(It.Is<SubscriberEntity>(rci => MatchReaderChangeInfo(rci, request))))
+                 .Throws<Exception>();
+
+            var result = await Handler.Handle(request, CancellationToken.None);
+
+            result.IsError.Should().BeTrue();
+            result.Error.Should().BeOfType<UnhandledExceptionError>();
+            _repositoryMock.Verify(x => x.GetSubscriberAsync(It.IsAny<SubscriberId>()), Times.Once);
+            _directorServiceMock.Verify(x => x.CreateReader(It.IsAny<SubscriberEntity>()), Times.Once);
+            _repositoryMock.Verify(x => x.SaveSubscriberAsync(It.IsAny<SubscriberEntity>()), Times.Never);
```
tabman83, line 180: Can use an assertion scope

src/Tests/CaptainHook.Application.Tests/Handlers/UpsertWebhookRequestHandlerTests.cs
```diff
@@ -0,0 +1,231 @@
...
+                .ReturnsAsync(new BusinessError("test error"));
+            _directorServiceMock.Setup(r => r.CreateReader(It.Is<SubscriberEntity>(rci => MatchReaderChangeInfo(rci, request))))
+                .ReturnsAsync(true);
+
+            var result = await Handler.Handle(request, CancellationToken.None);
+
+            result.IsError.Should().BeTrue();
+            _repositoryMock.Verify(x => x.GetSubscriberAsync(It.IsAny<SubscriberId>()), Times.Once);
+            _directorServiceMock.Verify(x => x.CreateReader(It.IsAny<SubscriberEntity>()), Times.Once);
+            _repositoryMock.Verify(x => x.SaveSubscriberAsync(It.IsAny<SubscriberEntity>()), Times.Once);
```
tabman83, line 200: Can use an assertion scope

src/Tests/CaptainHook.Application.Tests/Handlers/UpsertWebhookRequestHandlerTests.cs
```diff
@@ -0,0 +1,231 @@
...
+                .Throws<Exception>();
+            _directorServiceMock.Setup(r => r.CreateReader(It.Is<SubscriberEntity>(rci => MatchReaderChangeInfo(rci, request))))
+                .ReturnsAsync(false);
+
+            var result = await Handler.Handle(request, CancellationToken.None);
+
+            result.IsError.Should().BeTrue();
+            _repositoryMock.Verify(x => x.GetSubscriberAsync(It.IsAny<SubscriberId>()), Times.Once);
+            _directorServiceMock.Verify(x => x.CreateReader(It.IsAny<SubscriberEntity>()), Times.Once);
+            _repositoryMock.Verify(x => x.SaveSubscriberAsync(It.IsAny<SubscriberEntity>()), Times.Once);
```
tabman83, line 220: Can use an assertion scope

src/CaptainHook.DirectorService/DirectorService.cs
dawidkeshop: Are you asking about this interface?
Request handlers will use it to make changes in ServiceFabric cluster through DirectorService.

src/CaptainHook.Domain/Errors/UnhandledExceptionError.cs
dawidkeshop: It's a wrapper for Exception which allows to catch any Exception in Handler and then return it to API controller in our standard (OperationResult) way. 
Alternatively we can have custom MediatR behaviour which will do the same, but I didn't want to introduce too many changes for a single handler.

src/CaptainHook.Application/Gateways/IDirectorServiceGateway.cs
```diff
@@ -1,18 +1,21 @@
...
+        /// Creates reader service for single webhook in Subscriber.
+        /// </summary>
+        /// <param name="subscriber">Subscriber entity defining ReaderService to be created</param>
+        /// <returns>
+        /// True if creation has been invoked
+        /// False if reader already exists
+        /// ReaderCreationError if reader creation failed
+        /// DirectorServiceIsBusyError if DirectorService is performing other operation
+        /// </returns>
+        Task<OperationResult<bool>> CreateReader(SubscriberEntity subscriber);
```
tabman83, line 19: Can you add the -Async suffix here

src/CaptainHook.Application/Handlers/Subscribers/UpsertWebhookRequestHandler.cs
```diff
@@ -0,0 +1,77 @@
...
+                var existingItem = await _subscriberRepository.GetSubscriberAsync(subscriberId);
+
+                if (existingItem.IsError)
+                {
+                    return existingItem.Error;
+                }
+
+                if (existingItem.Data != null)
+                {
+                    return new BusinessError("Updating subscribers not supported!");
```
mszymeshop, line 41: This is just a temporary update constraint? As normally I would prefer deeper message information, however, I understand this is going to go shortly, right?

src/CaptainHook.Application/Gateways/IDirectorServiceGateway.cs
```diff
@@ -0,0 +1,21 @@
+﻿using System.Threading.Tasks;
+using CaptainHook.Domain.Entities;
+using CaptainHook.Domain.Results;
+
+namespace CaptainHook.Application.Gateways
```
mszymeshop, line 5: Do you want o group gateways together in this namespace? Or does it make sense to have a namespace dedicated to a specific area (maybe Director Service or maybe something else).

src/CaptainHook.DirectorService/DirectorService.cs
mszymeshop: @dawidkeshop do you use it as a remoting service? If such, it should inherit from IService as I think you're crossing the boundary here calling it from Application layer.

Having said that, is the pattern the Gateway here or just Proxy? Does the name suggest a pattern used or is just a name you picked up for this?

src/CaptainHook.Application/Handlers/Subscribers/UpsertWebhookRequestHandler.cs
dawidkeshop: Yes, it's temporary. It will be removed when we play a story to support adding a webhook to an existing subscriber.

src/CaptainHook.Application/Gateways/IDirectorServiceGateway.cs
dawidkeshop: In this project we have such namespaces: Requests, Handlers, Validators, so as for me it was logical to create another namespace for all Gateways and single interface per each gateway.

src/CaptainHook.Application/Handlers/Subscribers/UpsertWebhookRequestHandler.cs
```diff
@@ -0,0 +1,77 @@
...
+                }
+
+                var subscriber = MapRequestToEntity(request);
+
+                var directorResult = await _directorService.CreateReader(subscriber);
```
chomick, line 46: A thought - another approach could be that you let Director Service to load the subscriber(s) that are changed/added/removed (you would only pass the names of the subscribers that should be refreshed somehow). In that flow, Director would also know that it does not need to load KVs. You would end up with the same reader creation/updates/deletes mechanism for all cases. I don't know what you plan to do in the implementation there (maybe this will exactly be it?) so maybe your solution is even better

src/CaptainHook.DirectorService/DirectorService.cs
dawidkeshop: Yes, you right it's remoting, I'll fix the missing `IService` during my current task.
It's more like a proxy, I just used the name used in clean architecture definition. I can change if you can suggest something better.

src/CaptainHook.Domain/Repositories/ISubscriberRepository.cs
```diff
@@ -14,10 +16,24 @@ public interface ISubscriberRepository
...
         /// </summary>
         /// <returns></returns>
         public Task<OperationResult<IEnumerable<SubscriberEntity>>> GetAllSubscribersAsync();
+
+        /// <summary>
+        /// Saves Subscriber
+        /// </summary>
+        /// <param name="subscriberEntity">Subscriber entity to Save</param>
+        /// <returns></returns>
+        public Task<OperationResult<SubscriberEntity>> SaveSubscriberAsync(SubscriberEntity subscriberEntity);
```
chomick, line 37: Is the idea that it creates or updates (in future)?

src/CaptainHook.Application/Handlers/Subscribers/UpsertWebhookRequestHandler.cs
dawidkeshop: My plan was to follow the flow diagram from the spike, which means first let Director to create new reader, then store new endpoint configuration in Cosmos. If I understand correctly, what you are suggesting requires the opposite flow and as far as I remember from our design conversation, we didn't want that order because it's better to not store configuration if reader creation fails.

src/CaptainHook.Domain/Repositories/ISubscriberRepository.cs
dawidkeshop: I would rather assume that Save and Update should be separate methods on repository level.

src/CaptainHook.Application/Handlers/Subscribers/UpsertWebhookRequestHandler.cs
chomick: You are right, I forgot about it and for some reason felt this was after the subscriber is updated :)
All makes sense.

src/CaptainHook.Domain/Repositories/ISubscriberRepository.cs
chomick: Maybe use different name, then - Save can mean both.
I suggest we use: 
Create, Delete, Update
-or-
Add, Remove, Modify

terminology to be consistent.

PR #158, opened 2020-07-31
Implement subscriber persistence, Unit tests

tabman83:
- extend the existing repository codebase
- base the storage on subscribers rather than endpoints
- unit tests


src/CaptainHook.Api/Controllers/EventsController.cs
```diff
@@ -35,6 +35,8 @@ public EventsController(IMediator mediator, IBigBrother bigBrother)
         public async Task<IActionResult> PutWebhook([FromRoute] string eventName, [FromRoute] string subscriberName, [FromBody] EndpointDto dto)
         {
             var request = new UpsertWebhookRequest(eventName, subscriberName, dto);
+
+            await Task.CompletedTask;
```
mszymeshop, line 39: temporary?

src/CaptainHook.Domain/ValueObjects/SubscriberId.cs
```diff
@@ -24,5 +24,10 @@ public override string ToString()
         {
             return _id;
         }
+
+        public static implicit operator string(SubscriberId subscriberId)
+        {
+            return subscriberId.ToString();
```
mszymeshop, line 30: handle nulls

src/CaptainHook.Storage.Cosmos/Models/SubscriberDocument.cs
```diff
@@ -77,5 +47,17 @@ internal class EndpointDocument
         /// </summary>
         [JsonProperty("eventName")]
         public string EventName { get; set; }
+
+        /// <summary>
+        /// Subscriber endpoints
+        /// </summary>
+        [JsonProperty(PropertyName = "endpoints")]
+        public EndpointSubdocument[] Endpoints { get; set; }
```
mszymeshop, line 55: Should we talk about distinguishment between dlq, webhook, callback before the implementation is needed?

src/CaptainHook.Storage.Cosmos/Models/SubscriberDocument.cs
```diff
@@ -77,5 +47,17 @@ internal class EndpointDocument
...
+
+        /// <summary>
+        /// Webhook selector
+        /// </summary>
+        [JsonProperty("webhookSelectionRule")]
+        public string WebhookSelectionRule { get; set; }
```
mszymeshop, line 61: What is the example usage of Rule?

src/Tests/CaptainHook.Application.Tests/ValidatorPipelineBehaviorTests.cs
```diff
@@ -63,6 +63,7 @@ private class DivideRequestHandler : IRequestHandler<DivideRequest, OperationRes
         {
             public async Task<OperationResult<DivideResult>> Handle(DivideRequest request, CancellationToken cancellationToken)
             {
+                await Task.CompletedTask;
```
mszymeshop, line 66: ?

src/CaptainHook.Storage.Cosmos/QueryBuilders/SubscriberQueryBuilder.cs
```diff
@@ -1,28 +1,42 @@
...
+        public CosmosQuery BuildSelectAllSubscribers()
         {
             var query = new QueryDefinition("select * from c where c.type = @documentType")
-                .WithParameter("@documentType", EndpointDocument.Type);
+                .WithParameter("@documentType", SubscriberDocument.Type);
 
             return new CosmosQuery(query);
         }
+
+        public CosmosQuery BuildSelectSubscriber(SubscriberId subscriberId)
```
mszymeshop, line 29: Looks like this query can be updated to be more efficient.

src/CaptainHook.Api/Controllers/EventsController.cs
tabman83: I think so. Was not part of my code, I only added this to fix the warning in sonarcloud

src/Tests/CaptainHook.Application.Tests/ValidatorPipelineBehaviorTests.cs
tabman83: Only added to fix the warning in SonarCloud.

src/CaptainHook.Storage.Cosmos/QueryBuilders/SubscriberQueryBuilder.cs
tabman83: Fixed with further commit.

src/CaptainHook.Domain/ValueObjects/SubscriberId.cs
tabman83: Fixed with further commit.

src/CaptainHook.Storage.Cosmos/Models/SubscriberDocument.cs
tabman83: The WebhookType deals with that

src/CaptainHook.Storage.Cosmos/Models/SubscriberDocument.cs
tabman83: We don't have it yet. This was carried over from the old model. At the time of definition it was agreed with Darek to leave it as a string for now until we have a full definition.

src/CaptainHook.Api/Controllers/EventsController.cs
mszymeshop: I'd much rather see Task.FromResult

src/CaptainHook.Storage.Cosmos/Models/SubscriberDocument.cs
mszymeshop: ok, got it

src/CaptainHook.Storage.Cosmos/QueryBuilders/SubscriberQueryBuilder.cs
mszymeshop: subscriber id used as partition key?

src/CaptainHook.Api/Controllers/EventsController.cs
dawidkeshop: It will be reworked in the next task anyway.

tabman83: You're right but this is a temporary endpoint from what I can see. Nothing is being done or returned at the moment

src/CaptainHook.Storage.Cosmos/QueryBuilders/SubscriberQueryBuilder.cs
tabman83: Give the way the subscriberId is build, it is also the pk.

src/CaptainHook.Storage.Cosmos/Models/SubscriberDocument.cs
chomick: @dawidkeshop @mszymeshop @tabman83 I think we should not have it unless we need it for now - in the past that was just to highlight how the 'schema' may look but at this stage I would put only data that we are using right now

src/CaptainHook.Storage.Cosmos/QueryBuilders/SubscriberQueryBuilder.cs
```diff
@@ -1,28 +1,39 @@
...
+using CaptainHook.Storage.Cosmos.Models;
 using Eshopworld.Data.CosmosDb;
 using Microsoft.Azure.Cosmos;
 
 namespace CaptainHook.Storage.Cosmos.QueryBuilders
 {
     public class SubscriberQueryBuilder: ISubscriberQueryBuilder
     {
-        public CosmosQuery BuildSelectSubscribersListEndpoints(string eventName)
+        public CosmosQuery BuildSelectSubscribersList(string eventName)
```
chomick, line 10: @dawidkeshop @mszymeshop @tabman83 a suggestion for name change: BuildSelect(For)EventSubscribers

src/CaptainHook.Storage.Cosmos/QueryBuilders/SubscriberQueryBuilder.cs
```diff
@@ -1,28 +1,39 @@
...
-                .WithParameter("@documentType", EndpointDocument.Type);
+                .WithParameter("@documentType", SubscriberDocument.Type);
 
             return new CosmosQuery(query);
         }
+
+        public CosmosQuery BuildSelectSubscriber(SubscriberId subscriberId)
+        {
+            var query = new QueryDefinition(@"select * from c where c.id = @id")
+                .WithParameter("@id", subscriberId);
```
chomick, line 32: @dawidkeshop @mszymeshop @tabman83 There seem to be duplication on how id is constructed. I think we should be using document's static GetDocumentId() method to calculate it. Also, Not certain if we need a concept of subscriber id in the domain?

src/CaptainHook.Domain/ValueObjects/SubscriberId.cs
```diff
@@ -3,9 +3,13 @@
     public class SubscriberId
```
chomick, line 3: @dawidkeshop @mszymeshop @tabman83 I know this is different PR but do we need the concept of subscriber Id in the domain? 
If we do, is the assumption that the Id in the domain is the same as Id in the database? In such case, who owns the document identifier: repo or domain?

tabman83:
This has been merged and we should address this comments in a further PR

PR #161, opened 2020-08-04
72406 - URI transformation in upsert endpoint

dawidkeshop:
It's not a full flow, this PR contains changes in:
- API contract
- validaiton
- application layer
- DirectorService interaction

What left is storing additional data in Cosmos


src/CaptainHook.Application.Infrastructure/Mappers/SubscriberEntityToConfigurationMapper.cs
```diff
@@ -32,18 +32,49 @@ await MapBasicSubscriberData(cosmosModel)
...
+                            Source = new SourceParserLocation { Replace = firstEndpoint.UriTransform.Replace },
+                            Destination = new ParserLocation { RuleAction = RuleAction.RouteAndReplace },
+                            Routes = new List<WebhookConfigRoute>
+                            {
+                                new WebhookConfigRoute
+                                {
+                                    Uri = firstEndpoint.Uri,
+                                    HttpVerb = firstEndpoint.HttpVerb,
+                                    Selector = "*",
+                                    AuthenticationConfig = await MapAuthentication(firstEndpoint.Authentication),
```
chomick, line 60: Why is that async?

src/CaptainHook.Application.Infrastructure/Mappers/SubscriberEntityToConfigurationMapper.cs
```diff
@@ -32,18 +32,49 @@ await MapBasicSubscriberData(cosmosModel)
...
         }
 
-        private async Task<SubscriberConfiguration> MapBasicSubscriberData(SubscriberEntity cosmosModel)
+        private async Task<SubscriberConfiguration> MapBasicSubscriberData(SubscriberEntity entity)
         {
-            var config = new SubscriberConfiguration
+            SubscriberConfiguration config;
+
+            var firstEndpoint = entity.Webhooks?.Endpoints?.SingleOrDefault();
+            if (firstEndpoint?.UriTransform != null)
```
chomick, line 40: I would probably encapsulate that in separate private methods as the IF-ELSE is rather large

src/CaptainHook.Application/Validators/Common/UriTransformValidator.cs
```diff
@@ -0,0 +1,31 @@
...
+        private static readonly Regex _extractSelectorsFromUri = new Regex("(?<=\\{).+?(?=\\})", RegexOptions.Compiled);
+
+        public UriTransformValidator(string uri)
+        {
+            _uri = uri;
+
+            CascadeMode = CascadeMode.StopOnFirstFailure;
+
+            RuleFor(x => x.Replace).NotEmpty()
+                .Must(x => x.ContainsKey("selector")).WithMessage("Routes dictionary must contain an item with 'selector' key")
```
chomick, line 21: I would add a note that this is temporary :)

src/CaptainHook.Domain/Entities/SubscriberEntity.cs
```diff
@@ -1,10 +1,14 @@
...
+﻿using CaptainHook.Domain.ValueObjects;
+
+namespace CaptainHook.Domain.Entities
 {
     /// <summary>
     /// Subscriber model
     /// </summary>
     public class SubscriberEntity
     {
+        public SubscriberId Id => new SubscriberId(ParentEvent?.Name, Name);
```
chomick, line 10: Should the event be a mandatory thing for the subscriber?

src/CaptainHook.Domain/Entities/UriTransformEntity.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System.Collections.Generic;
+
+namespace CaptainHook.Domain.Entities
+{
+    public class UriTransformEntity
```
chomick, line 5: Not everything is an 'entity' ddd-wise here :) but I get if we only want to have consistent naming convention

src/CaptainHook.Application/Validators/Common/UriTransformValidator.cs
```diff
@@ -0,0 +1,31 @@
...
+using System.Text.RegularExpressions;
+using CaptainHook.Contract;
+using FluentValidation;
+
+namespace CaptainHook.Application.Validators.Common
+{
+    public class UriTransformValidator : AbstractValidator<UriTransformDto>
+    {
+        private readonly string _uri;
+        private static readonly Regex _extractSelectorsFromUri = new Regex("(?<=\\{).+?(?=\\})", RegexOptions.Compiled);
```
chomick, line 12: Should we not use the same mechanism that can parse and replace the variables?
If we were to modify any syntax-related definitions, we have plenty of code that has to be changed now, including some hacks to validate URLs etc...
I think we should design code in such way that the 'service' that can replace vars in the URL can be used in many places, including:
- validity of the input
- validity of the url
- actual replacement for the url

if it exposes methods like:
- GetVars (string input) -> list of var names
- Substitute (string input, dictionary<>) -> resulting value, can throw an error if vars missing etc

src/Tests/CaptainHook.Application.Tests/RequestValidators/UpsertSubscriberRequestValidatorTests.cs
```diff
@@ -110,5 +111,61 @@ public void When_HttpVerbIsNotValidHTTPVerb_Then_ValidationFails(string httpVerb
 
             result.AssertSingleFailure(nameof(EndpointDto.HttpVerb));
         }
+
+        [Fact, IsUnit]
```
chomick, line 115: Would like to see tests where string contains "{" but no "}" or vice versa

src/CaptainHook.Application.Infrastructure/Mappers/SubscriberEntityToConfigurationMapper.cs
dawidkeshop: Because it makes async call to `secretProvider.GetSecretValueAsync` method.

src/CaptainHook.Application.Infrastructure/Mappers/SubscriberEntityToConfigurationMapper.cs
dawidkeshop: Sure

src/CaptainHook.Application/Validators/Common/UriTransformValidator.cs
dawidkeshop: Ok

src/CaptainHook.Domain/Entities/SubscriberEntity.cs
dawidkeshop: I think so, there is no such thing like Subscriber without an Event, am I wrong?

src/CaptainHook.Domain/Entities/UriTransformEntity.cs
dawidkeshop: Actually entities are: Event, Subscriber and Endpoint. All the rest: Authentication, SecretStore, UriTransform and Webhooks are value objects. We can discuss that within the team and cleanup later.

src/CaptainHook.Application/Validators/Common/UriTransformValidator.cs
dawidkeshop: I think I got the idea, but it would be better to have a separate tech debt for that, then we will be able to improve also relevant code in DirectorService.

src/Tests/CaptainHook.Application.Tests/RequestValidators/UpsertSubscriberRequestValidatorTests.cs
dawidkeshop: This also can be done as part of above-mentioned tech debt. Current validation is not so clever to detect such issue.

tabman83:
Just two notes:
- I would have the async map methods renamed with the -async suffix
- Excellent result on SonarCloud:
https://sonarcloud.io/dashboard?id=esw.captain-hook&pullRequest=161
just next time have a look at the code smell tab as well

PR #164, opened 2020-08-05
#78979 Subscriber PUT endpoint and contract

tabman83:


src/CaptainHook.Application/Requests/Subscribers/UpsertSubscriberRequest.cs
```diff
@@ -0,0 +1,22 @@
...
+using MediatR;
+
+namespace CaptainHook.Application.Requests.Subscribers
+{
+    public class UpsertSubscriberRequest : IRequest<OperationResult<EndpointDto>>
+    {
+        public string EventName { get; }
+        public string SubscriberName { get; }
+        public SubscriberType SubscriberType { get; }
+        public SubscriberDto Endpoint { get; }
```
mszymeshop, line 12: We should support multiple endpoints at this stage

src/CaptainHook.Api/Controllers/EventsController.cs
```diff
@@ -68,5 +68,41 @@ public async Task<IActionResult> PutWebhook([FromRoute] string eventName, [FromR
...
+        [ProducesResponseType(typeof(ErrorBase), StatusCodes.Status403Forbidden)]
+        [ProducesResponseType(typeof(DirectorServiceIsBusyError), StatusCodes.Status409Conflict)]
+        [ProducesResponseType(typeof(ErrorBase), StatusCodes.Status500InternalServerError)]
+        public async Task<IActionResult> PutSuscriber([FromRoute] string eventName, [FromRoute] string subscriberName, [FromRoute] SubscriberType type, [FromBody] SubscriberDto dto)
+        {
+            var request = new UpsertSubscriberRequest(eventName, subscriberName, type, dto);
+            var result = await _mediator.Send(request);
+
+            return result.Match<IActionResult>(
+                 error => error switch
```
dawidkeshop, line 94: If we are going to handle same errors as in case of single endpoint, maybe would be good to extract this whole error mapping code to a separate method?

src/CaptainHook.Application/Requests/Subscribers/UpsertSubscriberRequest.cs
```diff
@@ -0,0 +1,22 @@
...
+using CaptainHook.Domain.Results;
+using MediatR;
+
+namespace CaptainHook.Application.Requests.Subscribers
+{
+    public class UpsertSubscriberRequest : IRequest<OperationResult<EndpointDto>>
+    {
+        public string EventName { get; }
+        public string SubscriberName { get; }
+        public SubscriberType SubscriberType { get; }
```
mszymeshop, line 11: There is no subscriber type just endpoint types (if need be)

src/CaptainHook.Contract/SubscriberDto.cs
```diff
@@ -2,6 +2,28 @@
...
     public class SubscriberDto
     {
+        public string SubscriberName { get; set; }
         public WebhooksDto Webhooks { get; set; }
     }
+
+    /// <summary>
+    /// Defines different type of subscribers
+    /// </summary>
+    public enum SubscriberType
```
mszymeshop, line 12: I don't think we should have a subscriber type.

src/CaptainHook.Application/Requests/Subscribers/UpsertSubscriberRequest.cs
tabman83: Right yes

src/CaptainHook.Application/Requests/Subscribers/UpsertSubscriberRequest.cs
tabman83: Why not? Let's chat on slack about that

src/CaptainHook.Api/Controllers/EventsController.cs
tabman83: This is a good idea but I am worried that if we do that at this stage, things might change when we add support for update, so we may want to leave this change for a little later.

src/CaptainHook.Application/Requests/Subscribers/UpsertSubscriberRequest.cs
tabman83: That is actually a subscriber in the request, and it's been correctly renamed

src/CaptainHook.Contract/SubscriberDto.cs
tabman83: Addressed

src/CaptainHook.Api/Controllers/EventsController.cs
```diff
@@ -68,5 +68,39 @@ public async Task<IActionResult> PutWebhook([FromRoute] string eventName, [FromR
...
              );
         }
+
+        /// <summary>
+        /// Insert or update a subscriber
+        /// </summary>
+        /// <param name="eventName">Event name</param>
+        /// <param name="dto">Webhook configuration</param>
+        /// <returns></returns>
+        [HttpPut("{eventName}/subscriber")]
```
mszymeshop, line 78: We actually want the name of the subscriber to be specified as a resource -> please follow documentation at https://eshopworld.atlassian.net/wiki/spaces/PLAT/pages/1839005765/Captain+Hook+API+definition

src/CaptainHook.Api/Controllers/EventsController.cs
dawidkeshop: OK, makes sense.

Also another thing to consider is if it's a good idea that Controllers knows Errors defined in the Domain layer.

src/CaptainHook.Api/Controllers/EventsController.cs
tabman83: Right I was just to quick and overlooked. Fixing it now.

tabman83: Also adding to the request object so we give an opportunity to validate it is the same subscriber name of the dto

src/CaptainHook.Api/Controllers/EventsController.cs
```diff
@@ -68,5 +68,40 @@ public async Task<IActionResult> PutWebhook([FromRoute] string eventName, [FromR
...
+            return result.Match<IActionResult>(
+                 error => error switch
+                 {
+                     ValidationError validationError => BadRequest(validationError),
+                     DirectorServiceIsBusyError directorServiceIsBusyError => Conflict(directorServiceIsBusyError),
+                     ReaderCreationError readerCreationError => UnprocessableEntity(readerCreationError),
+                     CannotSaveEntityError cannotSaveEntityError => UnprocessableEntity(cannotSaveEntityError),
+                     _ => StatusCode(StatusCodes.Status500InternalServerError, error)
+                 },
+                 subscriberDto => Created($"/{eventName}/subscriber", subscriberDto)
```
mszymeshop, line 101: The created uri is not correct

src/CaptainHook.Contract/SubscriberDto.cs
```diff
@@ -2,6 +2,7 @@
 {
     public class SubscriberDto
     {
+        public string SubscriberName { get; set; }
```
mszymeshop, line 5: I don't think it should be part of the Dto, and most definitely not assignable as we don't want the request to populate this property (might be a mismatch between what's here and what's in the URL)

src/CaptainHook.Api/Controllers/EventsController.cs
tabman83: Updated, thanks

src/CaptainHook.Contract/SubscriberDto.cs
tabman83: It definitely should not be assignable but the reason why I have it here is so it can be used by a get operation

tabman83: Removing it for now as we discussed we might have separate dto for get operation

PR #165, opened 2020-08-05
Map webhook selection rule

tabman83:


src/CaptainHook.Storage.Cosmos/SubscriberRepository.cs
```diff
@@ -142,6 +142,7 @@ private SubscriberDocument Map(SubscriberEntity subscriberEntity)
             {
                 EventName = subscriberEntity.ParentEvent.Name,
                 SubscriberName = subscriberEntity.Name,
+                WebhookSelectionRule = subscriberEntity.Webhooks?.SelectionRule,
```
dawidkeshop, line 145: Do we have tests for that?

tabman83: What kind of tests were you envisioning?

dawidkeshop: Ignore, I was wrong.

chomick: Well, a unit tests should verify the mapping was correct :) If test exists, it should be extended with this field; if not, I think the tests should be here for that.
Also, Nino, while you are in the repo world, can you fix the comment I made in relation to the query builder and how we use the identifier - this may require a quick chat with the guys as currently we define the identifier build-up logic in two places: domain & repo.

chomick, line 145: Should the rule & endpoints not be wrapped in the webhook container? The subscriber will have in future DLQ & Callbacks as well and that will avoid us removing all data as we change the model. We could probably add the 'webhook' definition that contains: rule & endpoints and reference it as webhook from the subscriber.

src/CaptainHook.Storage.Cosmos/Models/SubscriberDocument.cs
```diff
@@ -10,13 +11,11 @@ internal class SubscriberDocument
...
         public static string GetPartitionKey(string eventName) => $"{Type}_{eventName}";
 
-        public static string GetDocumentId(string eventName, string subscriberName) => $"{eventName}-{subscriberName}";
-
         /// <summary>
         /// Identifier
         /// </summary>
         [JsonProperty("id")]
-        public string Id => GetDocumentId(EventName, SubscriberName);
+        public SubscriberId Id => new SubscriberId(EventName, SubscriberName);
```
dawidkeshop, line 18: I think would be better to keep it as `string Id { get; set; }` and assign its value in SubscriberRepository, in the Map method where SubscriberDocument is created. I don't think it's good to make the persistence model dependent on a type defined in the domain model. Also, SubscriberId should be created only once, here you creating it again.

mszymeshop:
@tabman83 can you merge this with my changes from https://github.com/eShopWorld/captain-hook/pull/168/files as some areas will lead to conflicts.

tabman83:
> @tabman83 can you merge this with my changes from https://github.com/eShopWorld/captain-hook/pull/168/files as some areas will lead to conflicts.

Will do!

src/CaptainHook.Storage.Cosmos/SubscriberRepository.cs
```diff
@@ -140,6 +140,7 @@ private SubscriberDocument Map(SubscriberEntity subscriberEntity)
 
             return new SubscriberDocument
             {
+                Id = new SubscriberId(subscriberEntity.ParentEvent.Name, subscriberEntity.Name),
```
dawidkeshop, line 143: Why not just `Id = subscriberEntity.Id`?

tabman83: Done

src/CaptainHook.Storage.Cosmos/Models/SubscriberDocument.cs
tabman83: Done

src/CaptainHook.Storage.Cosmos/SubscriberRepository.cs
tabman83: Following conversation with Darek I am introducing the hierarchical structure

src/CaptainHook.Storage.Cosmos/SubscriberRepository.cs
```diff
@@ -134,27 +134,40 @@ private async Task<IEnumerable<SubscriberEntity>> GetSubscribersListInternalAsyn
 
         private SubscriberDocument Map(SubscriberEntity subscriberEntity)
         {
-            var endpoints =
-                subscriberEntity.Webhooks?.Endpoints?.Select(webhookEndpoint => Map(webhookEndpoint, EndpointType.Webhook))
-                ?? Enumerable.Empty<EndpointSubdocument>();
+            //var endpoints =
```
chomick, line 137: Any reason we need this commented code?

src/CaptainHook.Application/Validators/Dtos/WebhooksDtoValidator.cs
```diff
@@ -1,6 +1,10 @@
 ﻿using CaptainHook.Contract;
 using FluentValidation;
 using Newtonsoft.Json.Linq;
+using System.Collections.Generic;
+using System.Linq;
+using System.Runtime.CompilerServices;
+using System.Security.Cryptography.X509Certificates;
```
mszymeshop, line 7: What are certificates doing here :)?

src/CaptainHook.Application/Validators/Dtos/WebhooksDtoValidator.cs
```diff
@@ -12,14 +16,35 @@ public WebhooksDtoValidator()
...
+
             RuleForEach(x => x.Endpoints)
                 .SetValidator(new EndpointDtoValidator());
+        }
 
-            RuleFor(x => x.SelectionRule).NotEmpty()
-                .Must(BeValidJsonPathExpression).WithMessage("The SelectionRule must be a valid JSONPath expression");
+        private bool ContainAtMostOneEndpointWithNoSelector(List<EndpointDto> endpoints)
+        {
+            return endpoints.Count(x => x.Selector == null) <= 1;
```
mszymeshop, line 39: Consider use of Any here so that the calculation is quicker and doesn't have to go through all the endpoints

src/CaptainHook.Storage.Cosmos/Models/WebhookSubdocument.cs
```diff
@@ -0,0 +1,22 @@
+﻿using Newtonsoft.Json;
+
+namespace CaptainHook.Storage.Cosmos.Models
+{
+    /// <summary>
+    /// Webhook model in Cosmos DB
+    /// </summary>
+    internal class WebhookSubdocument
```
mszymeshop, line 8: Is it going to be `EndpointsSubdocument` so that it can be reused between webhooks, callbacks and dlq?

src/CaptainHook.Application/Validators/Dtos/WebhooksDtoValidator.cs
tabman83: Blame intellisense... removing...

src/CaptainHook.Storage.Cosmos/SubscriberRepository.cs
tabman83: Removing

src/CaptainHook.Storage.Cosmos/Models/WebhookSubdocument.cs
tabman83: We already have an endpointsubdocument type for endpoints inside. So either we leave it as it is for now or we found a better name, which I haven't figured out yet

src/CaptainHook.Api/appsettings.json
```diff
@@ -10,10 +10,8 @@
     }
   },
   "ServiceConfigurationOptions": {
-    "ApiName": "tooling.eda.api",
-    "RequiredScopes": [
-      "tooling.eda.api.all"
-    ],
+    "ApiName": "platform.captainhook.api",
+    "RequiredScopes": [],
```
chomick, line 14: i think you may need at least one scope here?

chomick: The GET operation is now broken as it expects incorrect scope: tooling.eda.api.all

PR #166, opened 2020-08-06
Add PUT subscriber validators and unit tests

tabman83:


src/CaptainHook.Application/Validators/Dtos/WebhooksDtoValidator.cs
```diff
@@ -0,0 +1,44 @@
...
+            CascadeMode = CascadeMode.StopOnFirstFailure;
+
+            RuleFor(x => x.Endpoints).NotEmpty()
+                .Must(list => list.Count >= 1).WithMessage("Webhooks list must contain at list one endpoint");
+
+            RuleForEach(x => x.Endpoints)
+                .SetValidator(new EndpointDtoValidator());
+
+            RuleFor(x => x.SelectionRule).NotEmpty()
+                .Must(BeValidJsonPathExpression);
```
mszymeshop, line 22: Should we have a custom message here as well?

src/Tests/CaptainHook.TestsInfrastructure/Builders/SubscriberDtoBuilder.cs
```diff
@@ -0,0 +1,12 @@
+﻿using CaptainHook.Contract;
+
+namespace CaptainHook.TestsInfrastructure.Builders
+{
+    public class SubscriberDtoBuilder: SimpleBuilder<SubscriberDto>
```
dawidkeshop, line 5: I think this is better implementation than @mszymeshop has in his parallel PR, so this version should override his.

src/CaptainHook.Application/Validators/Dtos/WebhooksDtoValidator.cs
tabman83: Just added, thanks.

PR #168, opened 2020-08-06
#78982 - add handler for subscriber request

mszymeshop:
### What
The new handler is added. We have noticed that it is very similar to the Webhook Handler and it is likely that we are going to unify them at the next stage.

### Testing
Code works and stores to Cosmos however not all mapping might be valid at this stage. JSON to test with (example):
```json
{
  "webhooks": {
    "selectionRule": "$.TestSelector",
    "endpoints": [
      {
        "uri": "https://webhook.site/a86944ce-b402-493b-ae57-1d679da8d82b",
        "httpVerb": "POST",
        "authentication": {
          "type": "OIDC",
          "clientId": "abc",
          "uri": "https://webhook.site/a86944ce-b402-493b-ae57-1d679da8d82b",
          "clientSecret": {
            "vault": "abc-keyvault",
            "name": "event--1--webhookconfig--authenticationconfig--clientsecret"
          },
          "scopes": [
            "scope1"
          ]
        }
      }
    ]
  }
}
```

### DIFF
--- a/src/CaptainHook.Application/Handlers/Subscribers/UpsertSubscriberRequestHandler.cs
+++ b/src/CaptainHook.Application/Handlers/Subscribers/UpsertSubscriberRequestHandler.cs
@@ -1,4 +1,5 @@
 ﻿using System;
+using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 using CaptainHook.Application.Infrastructure.DirectorService;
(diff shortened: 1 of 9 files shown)


src/CaptainHook.Application/Handlers/Subscribers/UpsertWebhookRequestHandler.cs
```diff
@@ -60,7 +60,7 @@ public async Task<OperationResult<EndpointDto>> Handle(UpsertWebhookRequest requ
 
         private static SubscriberEntity MapRequestToEntity(UpsertWebhookRequest request)
         {
-            var subscriber = new SubscriberEntity(request.SubscriberName, null, new EventEntity(request.EventName));
```
mszymeshop, line 63: @tabman83 depending on your investigation Nino we might need to change this to something else anyway

PR #169, opened 2020-08-06
79299 remove WebhookConfig usages from Readers and Director

dawidkeshop:


src/CaptainHook.DirectorService/DirectorService.cs
```diff
@@ -60,22 +60,15 @@ public class DirectorService : StatefulService, IDirectorServiceRemoting
             _readerServiceChangeDetector = readerServiceChangeDetector ?? throw new ArgumentNullException(nameof(readerServiceChangeDetector));
         }
 
-        private async Task<(IList<WebhookConfig> newWebhookConfig, IDictionary<string, SubscriberConfiguration> newSubscriberConfigurations)> LoadConfigurationAsync()
+        private async Task<Dictionary<string, SubscriberConfiguration>> LoadConfigurationAsync()
```
chomick, line 63: Can we use IDictionary instead of concrete implementation?

src/CaptainHook.DirectorService/Infrastructure/Interfaces/ISubscriberConfigurationLoader.cs
```diff
@@ -1,11 +1,12 @@
...
 using System.Collections.Generic;
+using System.Collections.ObjectModel;
 using System.Threading.Tasks;
 
 namespace CaptainHook.DirectorService.Infrastructure.Interfaces
 {
     public interface ISubscriberConfigurationLoader
     {
-        Task<(IList<WebhookConfig>, IList<SubscriberConfiguration>)> LoadAsync(string keyVaultUri);
+        Task<ReadOnlyCollection<SubscriberConfiguration>> LoadAsync(string keyVaultUri);
```
chomick, line 10: Maybe we can simply have IEnum<>?

src/CaptainHook.EventReaderService/EventReaderService.cs
```diff
@@ -315,7 +315,6 @@ private MessageData BuildMessageData(Message message)
             messageData.CorrelationId = Guid.NewGuid().ToString();
             messageData.ServiceBusMessageId = message.MessageId;
             messageData.SubscriberConfig = _initData.SubscriberConfiguration;
-            messageData.WebhookConfig = _initData.WebhookConfig;
```
chomick, line 318: Will the hash of config be different now (not having webhook entry)?

chomick: So all readers will be recreated, right?

src/CaptainHook.DirectorService/DirectorService.cs
dawidkeshop: sure

src/CaptainHook.DirectorService/Infrastructure/Interfaces/ISubscriberConfigurationLoader.cs
dawidkeshop: sure

src/CaptainHook.EventReaderService/EventReaderService.cs
dawidkeshop: No, they won't. WebhookConfigs are not used to calculate hash, only SubscriberConfigs are considered.

PR #170, opened 2020-08-07
#78987 - handler to accept additional endpoints for subscriber

mszymeshop:
Handler gets extended so that it can handle both Additions and Updates. Updates always go through but do not have any effect in Director Service or Cosmos just yet.

### DIFF
--- a/src/CaptainHook.Application.Infrastructure/DirectorService/DirectorServiceProxy.cs
+++ b/src/CaptainHook.Application.Infrastructure/DirectorService/DirectorServiceProxy.cs
@@ -33,5 +33,10 @@ namespace CaptainHook.Application.Infrastructure.DirectorService
                 _ => new BusinessError("Director Service returned unknown result.")
             };
         }
+
+        public Task<OperationResult<bool>> UpdateReaderAsync(SubscriberEntity subscriber)
+        {
+            return Task.FromResult(new OperationResult<bool>(true));
+        }
     }
 }
--- a/src/CaptainHook.Application/Handlers/Subscribers/UpsertWebhookRequestHandler.cs
+++ b/src/CaptainHook.Application/Handlers/Subscribers/UpsertWebhookRequestHandler.cs
@@ -24,33 +24,25 @@ namespace CaptainHook.Application.Handlers.Subscribers
             _directorService = directorService;
         }
 
-        public async Task<OperationResult<EndpointDto>> Handle(UpsertWebhookRequest request, CancellationToken cancellationToken)
+        public async Task<OperationResult<EndpointDto>> Handle(UpsertWebhookRequest request,
+            CancellationToken cancellationToken)
         {
             try
             {
                 var subscriberId = new SubscriberId(request.EventName, request.SubscriberName);
                 var existingItem = await _subscriberRepository.GetSubscriberAsync(subscriberId);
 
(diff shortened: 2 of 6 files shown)


src/CaptainHook.Application/Handlers/Subscribers/UpsertWebhookRequestHandler.cs
```diff
@@ -24,49 +24,99 @@ public UpsertWebhookRequestHandler(ISubscriberRepository subscriberRepository, I
...
+        public async Task<OperationResult<EndpointDto>> Handle(UpsertWebhookRequest request,
+            CancellationToken cancellationToken)
         {
             try
             {
                 var subscriberId = new SubscriberId(request.EventName, request.SubscriberName);
                 var existingItem = await _subscriberRepository.GetSubscriberAsync(subscriberId);
 
-                if (!(existingItem.Error is EntityNotFoundError))
+                if (existingItem.IsError && existingItem.Error is EntityNotFoundError)
```
dawidkeshop, line 35: I think checking `IsError` here is redundant.

src/CaptainHook.Application.Infrastructure/DirectorService/DirectorServiceProxy.cs
```diff
@@ -33,5 +33,10 @@ public async Task<OperationResult<bool>> CreateReaderAsync(SubscriberEntity subs
                 _ => new BusinessError("Director Service returned unknown result.")
             };
         }
+
+        public Task<OperationResult<bool>> UpdateReaderAsync(SubscriberEntity subscriber)
```
dawidkeshop, line 37: Shouldn't this method be marked `async` if it has `Async` suffix?
If so, you could replace the body of this method just with `return true;`.

src/CaptainHook.Storage.Cosmos/SubscriberRepository.cs
```diff
@@ -102,6 +102,11 @@ public async Task<OperationResult<IEnumerable<SubscriberEntity>>> GetSubscribers
             return materialized;
         }
 
+        public Task<OperationResult<SubscriberEntity>> UpdateSubscriberAsync(SubscriberEntity subscriber)
```
dawidkeshop, line 105: Shouldn't this method be marked async if it has Async suffix?
If so, you could replace the body of this method just with `return subscriber;`.

src/CaptainHook.Application.Infrastructure/DirectorService/DirectorServiceProxy.cs
mszymeshop: The return type expects Task, this is why I named it Async. The actual async/await implementation will take place later and that is why Task.Fromresult is used. I cannot do return true as it would not compile.

src/CaptainHook.Storage.Cosmos/SubscriberRepository.cs
mszymeshop: Same as above

src/CaptainHook.Application/Handlers/Subscribers/UpsertWebhookRequestHandler.cs
mszymeshop: I'll be circling around this area later, so will try to simplify it

PR #171, opened 2020-08-07
Extend Subscriber Repository to handle subscriber updates

tabman83:


src/CaptainHook.Storage.Cosmos/SubscriberRepository.cs
```diff
@@ -102,12 +102,31 @@ public async Task<OperationResult<IEnumerable<SubscriberEntity>>> GetSubscribers
...
+                throw new ArgumentNullException(nameof(subscriberEntity));
+            }
+
+            try
+            {
+                return await UpdateSubscriberInternalAsync(subscriberEntity);
+            }
+            catch
+            {
+                return new CannotSaveEntityError(nameof(SubscriberEntity));
```
dawidkeshop, line 118: I think would be good to pass the original message as part of the `CannotSaveEntityError`. In other case might be hard to investigate what exactly had happen.

src/CaptainHook.Storage.Cosmos/SubscriberRepository.cs
```diff
@@ -30,111 +33,156 @@ public class SubscriberRepository : ISubscriberRepository
...
+        private async Task<OperationResult<SubscriberEntity>> UpdateSubscriberInternalAsync(SubscriberEntity subscriberEntity)
         {
-            return Task.FromResult(new OperationResult<SubscriberEntity>(subscriber));
-        }
+            try
+            {
+                var subscriberDocument = Map(subscriberEntity);
 
-        #region Private methods
+                var result = await _cosmosDbRepository.UpsertAsync(subscriberDocument);
```
chomick, line 116: Should this be a "replace" repo operation instead?
We have a separate 'create' operation and we will need etag when we do 'replace' eventaully.

src/CaptainHook.Storage.Cosmos/SubscriberRepository.cs
```diff
@@ -30,111 +33,156 @@ public class SubscriberRepository : ISubscriberRepository
...
-            return subscribers
-                .Select(x => Map(x))
-                .ToList();
+                return subscribers
+                    .Select(x => Map(x))
+                    .ToList();
+            }
+            catch (Exception exception)
+            {
+                _bigBrother.Publish(exception.ToExceptionEvent());
```
chomick, line 183: We already log the exceptions in the repository itself (although I can see incorrectly, not using the ToException extension).
I suggest we remove all the logs as they will only duplicate the entries.

src/Tests/CaptainHook.Storage.Cosmos.Tests/SubscriberRepositoryTests.cs
```diff
@@ -451,5 +456,67 @@ public async Task AddSubscriberInternalAsync_should_return_error_if_create_fails
...
+            subscriber.AddWebhooks(new WebhooksEntity("rule"));
+            _cosmosDbRepositoryMock
+                .Setup(x => x.CreateAsync(It.IsAny<SubscriberDocument>()))
+                .ReturnsAsync(new DocumentContainer<SubscriberDocument>(subscriberDocument, "etag"));
+
+            // Act
+            var result = await _repository.UpdateSubscriberAsync(subscriber);
+
+            // Assert
+            _cosmosDbRepositoryMock.Verify(x => x.UpsertAsync(It.IsAny<SubscriberDocument>()));
```
chomick, line 483: You want to verify that the repo call was with the correct data.
Also, you are mocking the 'create' repo method which is incorrect?

src/Tests/CaptainHook.Storage.Cosmos.Tests/SubscriberRepositoryTests.cs
```diff
@@ -451,5 +456,67 @@ public async Task AddSubscriberInternalAsync_should_return_error_if_create_fails
             result.Error.Should().BeOfType<CannotSaveEntityError>();
         }
 
+        [Fact, IsUnit]
+        public async Task UpdateSubscriberAsync_should_invoke_UpsertAsync()
```
chomick, line 460: I have seen some comments in the past related to test names - are we following any standard here? I thought there was a winner 

MethodName_ConditionDetails_ResultHere

?

src/Tests/CaptainHook.Storage.Cosmos.Tests/SubscriberRepositoryTests.cs
```diff
@@ -451,5 +456,67 @@ public async Task AddSubscriberInternalAsync_should_return_error_if_create_fails
...
+
+            // Act
+            var result = await _repository.UpdateSubscriberAsync(subscriber);
+
+            // Assert
+            _cosmosDbRepositoryMock.Verify(x => x.UpsertAsync(It.IsAny<SubscriberDocument>()));
+        }
+
+        [Fact, IsUnit]
+        public async Task UpdateSubscriberAsync_should_throw_for_invalid_SubscriberId()
```
chomick, line 487: The better name here would be:
UpdateSubscriberAsync_WithNullSubscriberId_ThrowsException

src/Tests/CaptainHook.Storage.Cosmos.Tests/SubscriberRepositoryTests.cs
```diff
@@ -451,5 +456,67 @@ public async Task AddSubscriberInternalAsync_should_return_error_if_create_fails
...
+                EventName = "eventName",
+                Webhooks = new WebhookSubdocument
+                {
+                    SelectionRule = "rule",
+                    Endpoints = new EndpointSubdocument[] { }
+                }
+            };
+            var subscriber = new SubscriberEntity("subscriberName", new EventEntity("eventName"));
+            _cosmosDbRepositoryMock
+                .Setup(x => x.CreateAsync(It.IsAny<SubscriberDocument>()))
```
chomick, line 512: Again, why do we setup 'create' here?

src/Tests/CaptainHook.Storage.Cosmos.Tests/SubscriberRepositoryTests.cs
chomick: UpdateSubscriberAsync_WithValidSubscriber_CallsRepoWithCorrectDocument

chomick: I also would like to see this test:
UpdateSubscriberAsync_WithValidSubscriber_ReturnsCorrectSubscriberData

src/Tests/CaptainHook.Storage.Cosmos.Tests/SubscriberRepositoryTests.cs
```diff
@@ -451,5 +456,67 @@ public async Task AddSubscriberInternalAsync_should_return_error_if_create_fails
...
+        {
+            // Act
+            Task act() => _repository.UpdateSubscriberAsync(null);
+
+            // Assert
+            await Assert.ThrowsAsync<ArgumentNullException>(act);
+        }
+
+        [Fact, IsUnit]
+        public async Task UpdateSubscriberInternalAsync_should_return_error_if_update_fails()
```
chomick, line 497: UpdateSubscriberInternalAsync_WithErrorWhileSaving_ReturnsCannotSaveEntityError

I also would like to see:
UpdateSubscriberInternalAsync_WithUnknownSubscriber_Returns??? - not sure what are the expectations here

src/CaptainHook.Domain/Repositories/ISubscriberRepository.cs
```diff
@@ -41,6 +41,6 @@ public interface ISubscriberRepository
         /// </summary>
         /// <param name="subscriber">Subscriber to update</param>
```
chomick, line 42: You did not update the name here

src/Tests/CaptainHook.Storage.Cosmos.Tests/SubscriberRepositoryTests.cs
```diff
@@ -451,5 +453,107 @@ public async Task AddSubscriberInternalAsync_should_return_error_if_create_fails
...
+            subscriber.AddWebhooks(new WebhooksEntity("rule"));
+            _cosmosDbRepositoryMock
+                .Setup(x => x.ReplaceAsync(subscriberDocument.Id, subscriberDocument, null))
+                .ThrowsAsync(new MissingDocumentException());
+
+            // Act
+            var result = await _repository.UpdateSubscriberAsync(subscriber);
+
+            // Assert
+            result.Error.Should().BeOfType<CannotUpdateEntityError>();
```
chomick, line 530: Do we want to have a different flow when we are attempting to update an entity vs flow where we fail to update it due to tech issues? Not a blocking issue now but maybe worth looking at the overall flow to see how we want to handle this case vs DB issues (I would expect to see different error message/http status code in such case)

tabman83: Be nice to distinguish, so it will require overall revisitation of the repo. Merging for now, leaving for later.

PR #172, opened 2020-08-07
#79449 Fix auth policies and scopes for API access

tabman83:


chomick:
Nino, you have to refresh your branch with latest master.

PR #173, opened 2020-08-10
#79751 Adding single webhook without selector fails

tabman83:
Steps:
- define new subscriber with a single webhook
- do not include "selector" field
- do not include "selctionRules" field
- invoke PUT on subscriber

Expected:
- new subscriber is added with default selection rules/selector


src/CaptainHook.Application/Validators/Dtos/WebhooksDtoValidator.cs
```diff
@@ -32,6 +32,11 @@ public WebhooksDtoValidator()
                 .SetValidator(new EndpointDtoValidator());
         }
 
+        private bool TheresAtLeastOneEndpointWithSelectorDefined(WebhooksDto webhooks)
+        {
+            return !string.IsNullOrWhiteSpace(webhooks.Endpoints.First().Selector);
```
chomick, line 37: This should be FirstOrDefault()

chomick: Otherwise you will get an exception

mszymeshop: I would even change it to something like
```
webhooks.Endpoints?.Any(e => !string.IsNullOrWhiteSpace(e.Selector)) ?? false;
```

You cannot guarantee your first endpoint will have the selector, right?

src/CaptainHook.Application/Validators/Dtos/WebhooksDtoValidator.cs
```diff
@@ -14,11 +14,11 @@ public WebhooksDtoValidator()
         {
             CascadeMode = CascadeMode.StopOnFirstFailure;
 
-            RuleFor(x => x.SelectionRule).NotEmpty()
-                .Must(BeValidJsonPathExpression).WithMessage("The SelectionRule must be a valid JSONPath expression");
+            RuleFor(x => x.Endpoints).NotEmpty();
```
mszymeshop, line 17: Are we happy with default messages here? The previous one was fine `Webhooks list must contain at list one endpoint` I think.

tabman83: We will get a message saying that the collection cannot be empty so I think it's clear enough

src/CaptainHook.Application/Validators/Dtos/WebhooksDtoValidator.cs
tabman83: This is happening after we enforce NotEmpty on endpoints a few lines before. 
The only reason I would change this to FirstOrDefault (or Any() as michal suggested) is if we want to be able to shuffle the rules above around. That would actually be safer.

chomick: We should code that validations themselves are independent to a degree - i.e. you don't want to get exception because you are assuming something about other validations you do (before or after) - they order may change at any time, for any reason and I would keep good coding practice in place

tabman83: Agreed and changed. Thanks.

src/Tests/CaptainHook.Application.Tests/Infrastructure/SubscriberEntityToConfigurationMapperTests.cs
```diff
@@ -65,11 +65,86 @@ public async Task When_UriTransformDefined_Then_MappedCorrectlyToRouteAndReplace
             subscriberConfiguration.WebhookRequestRules.Count.Should().Be(1);
             var rule = subscriberConfiguration.WebhookRequestRules.Single();
             rule.Routes.Count.Should().Be(1);
-            rule.Routes[0].Uri.Should().Be("https://blah-{selector}.eshopworld.com/webhook/");
+            rule.Routes[0].Uri.Should().Be("https://blah-{orderCode}.eshopworld.com/webhook/");
             rule.Routes[0].HttpMethod.Should().Be(HttpMethod.Post);
             rule.Routes[0].Selector.Should().Be("*");
             rule.Routes[0].AuthenticationConfig.Type.Should().Be(AuthenticationType.OIDC);
```
chomick, line 71: We should fully verify the authentication mappings as well - doesn't look we do that right now.

src/Tests/CaptainHook.Application.Tests/Infrastructure/SubscriberEntityToConfigurationMapperTests.cs
```diff
@@ -65,11 +65,86 @@ public async Task When_UriTransformDefined_Then_MappedCorrectlyToRouteAndReplace
...
             rule.Destination.RuleAction.Should().Be(RuleAction.RouteAndReplace);
+            rule.Source.Replace.Should().ContainKey("orderCode").WhichValue.Should().Be("$.OrderCode");
```
chomick, line 73: @mszymeshop will it work if the 'selector' variable is not defined? 

A side note (@dawidkeshop ) - I think number of validations we do for DTOs should probably be part of domain validations, e.g. combinations of valid variables/endpoints selectors etc - right now we can define incorrect (?) models/entities and not even know about it. I don't think we should duplicate them, instead we should let domain check some of those things. Right now it feels the domain is not protected at all with the validations. If I try to use the domain from different angle (not via API/DTO), there is no protection against incorrect data. This should not be the case. 

I am not advocating to change it right now but it is something I would probably do differently if we were starting from scratch.

src/CaptainHook.Application.Infrastructure/Mappers/SubscriberEntityToConfigurationMapper.cs
```diff
@@ -17,51 +17,89 @@ public SubscriberEntityToConfigurationMapper(ISecretProvider secretProvider)
             _secretProvider = secretProvider;
         }
 
-        public async Task<IEnumerable<SubscriberConfiguration>> MapSubscriber(SubscriberEntity cosmosModel)
+        public async Task<IEnumerable<SubscriberConfiguration>> MapSubscriberAsync(SubscriberEntity entity)
         {
-            return new[]
+            return new List<SubscriberConfiguration>
```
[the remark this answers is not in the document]
tabman83: Changed

src/CaptainHook.Application.Infrastructure/Mappers/SubscriberEntityToConfigurationMapper.cs
```diff
@@ -17,51 +17,89 @@ public SubscriberEntityToConfigurationMapper(ISecretProvider secretProvider)
...
+            };
         }
 
-        private async Task<SubscriberConfiguration> MapStandardWay(SubscriberEntity entity, EndpointEntity firstEndpoint)
+        private async Task<SubscriberConfiguration> MapForUriTransformAsync(SubscriberEntity entity)
         {
+            var replacements = new Dictionary<string, string>(entity.Webhooks.Endpoints
+                .SelectMany(x => x.UriTransform.Replace)
+                .GroupBy(x => x.Key)
+                .Select(x => x.First()))
```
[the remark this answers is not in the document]
chomick: I read the Linq wrong - the 'first' is applied to grouping result, not list of groups, my bad.
However, if the same variable name is used in different endpoints, and those variables have different definitions, than we are not consistent and we still have to add the extra validation (DTO or domain) to ensure all vars share the same definition in the webhook

src/CaptainHook.Application.Infrastructure/Mappers/SubscriberEntityToConfigurationMapper.cs
```diff
@@ -72,24 +110,24 @@ private async Task<SubscriberConfiguration> MapForUriTransform(SubscriberEntity
...
+                                AuthenticationConfig = await MapAuthenticationAsync(endpoint.Authentication),
                             }
                         }
                     }
                 }
             };
         }
 
-        private async Task<AuthenticationConfig> MapAuthentication(AuthenticationEntity cosmosAuthentication)
+        private async Task<AuthenticationConfig> MapAuthenticationAsync(AuthenticationEntity cosmosAuthentication)
```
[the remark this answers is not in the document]
chomick: Something to consider in future.

src/Tests/CaptainHook.Application.Tests/Infrastructure/SubscriberEntityToConfigurationMapperTests.cs
```diff
@@ -65,11 +65,86 @@ public async Task When_UriTransformDefined_Then_MappedCorrectlyToRouteAndReplace
             subscriberConfiguration.WebhookRequestRules.Count.Should().Be(1);
             var rule = subscriberConfiguration.WebhookRequestRules.Single();
             rule.Routes.Count.Should().Be(1);
-            rule.Routes[0].Uri.Should().Be("https://blah-{selector}.eshopworld.com/webhook/");
+            rule.Routes[0].Uri.Should().Be("https://blah-{orderCode}.eshopworld.com/webhook/");
             rule.Routes[0].HttpMethod.Should().Be(HttpMethod.Post);
```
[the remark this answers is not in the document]
tabman83: Add test variations for GET, PUT, POST

src/Tests/CaptainHook.Application.Tests/Infrastructure/SubscriberEntityToConfigurationMapperTests.cs
tabman83: Added tests for authentication mappings

src/CaptainHook.Application.Infrastructure/Mappers/SubscriberEntityToConfigurationMapper.cs
```diff
@@ -17,51 +17,89 @@ public SubscriberEntityToConfigurationMapper(ISecretProvider secretProvider)
...
+            {
+                Uri = endpoint.Uri,
+                HttpVerb = endpoint.HttpVerb,
+                Selector = endpoint.Selector ?? "*",
+                AuthenticationConfig = await MapAuthenticationAsync(endpoint.Authentication),
             };
         }
 
-        private async Task<SubscriberConfiguration> MapForUriTransform(SubscriberEntity entity, EndpointEntity firstEndpoint)
+        private async Task<SubscriberConfiguration> MapSingleWebhookWithUriTransformAsync(EndpointEntity endpoint, SubscriberEntity entity)
```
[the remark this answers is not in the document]
chomick: i would remove this method if i was doing some additional changes in this pr :)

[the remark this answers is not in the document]
tabman83: Done

src/Tests/CaptainHook.Application.Tests/Infrastructure/SubscriberEntityToConfigurationMapperTests.cs
```diff
@@ -25,15 +25,15 @@ public SubscriberEntityToConfigurationMapperTests()
...
                 "https://blah-blah.sts.eshopworld.com", "OIDC", new[] { "scope1" });
             var subscriber = new SubscriberBuilder()
                 .WithWebhook("https://blah-blah.eshopworld.com/webhook/", "POST", string.Empty, authentication: authentication)
                 .Create();
 
-            var result = await new SubscriberEntityToConfigurationMapper(_secretProviderMock.Object).MapSubscriber(subscriber);
+            var result = await new SubscriberEntityToConfigurationMapper(_secretProviderMock.Object).MapSubscriberAsync(subscriber);
 
             result.Should().HaveCount(1);
             var subscriberConfiguration = result.Single();
```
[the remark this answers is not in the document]
tabman83: I have added tests for different http verbs there

PR #174, opened 2020-08-10
#78994 - validate add webhook request

mszymeshop:
I am about to start working on unit tests for this change. Just wanted to share my approach here in case you find something important. Will update the PR as soon as I have unit tests.


src/CaptainHook.Application/ApplicationModule.cs
```diff
@@ -22,9 +21,9 @@ protected override void Load(ContainerBuilder builder)
             builder.RegisterType<DirectorServiceProxy>().As<IDirectorServiceProxy>();
             builder.RegisterType<SubscriberEntityToConfigurationMapper>().As<ISubscriberEntityToConfigurationMapper>();
 
-            var handlersAssembly = typeof(UpsertWebhookRequestHandler).Assembly;
-            var validatorsAssembly = typeof(UpsertWebhookRequestValidator).Assembly;
-            builder.RegisterMediatorInfrastructure(handlersAssembly, validatorsAssembly);
+            builder.RegisterMediatorInfrastructure(ThisAssembly)
```
mszymeshop, line 24: We should be using `ThisAssembly` property where it makes sense. Please note that Entity validators are registered from here even though they are actually stored in the Domain layer. Something to discuss potentially. To register Validators from the Domain Module it would require further changes along with `RegisterValidationInfrastructure` being move to Common.

src/CaptainHook.Application/Handlers/Subscribers/UpsertWebhookRequestHandler.cs
```diff
@@ -79,16 +86,18 @@ public UpsertWebhookRequestHandler(ISubscriberRepository subscriberRepository, I
             var endpoint = MapRequestToEndpoint(request, existingItem);
             existingItem.AddWebhookEndpoint(endpoint);
```
mszymeshop, line 87: Validation does not take place inside entities. It gives us better control over the flow as we do not have to throw exceptions or inject validators into entities. It introduces incoherence however as entities can be invalid until proven otherwise. I am ok with this approach but let's ensure there is a buy-in from everyone (@chomick  :)?) before we invest more effort.

src/CaptainHook.Application/ApplicationModule.cs
chomick: We would not need to register domain validators here, I would think, as they would not be invoked as part of mediator flow; instead they would be invoked manually when required in the domain layer - is that correct?

src/CaptainHook.Application/Handlers/Subscribers/UpsertWebhookRequestHandler.cs
chomick: I think the entity should return an error if operation you perform on it is incorrect.
In this case, you should not be allowed to add new endpoint to the entity and the method should respond with OperationResult<> of some sort.

mszymeshop: It's good I asked then. Let me see what I can do here. @tabman83 @dawidkeshop what do you reckon?

src/CaptainHook.Domain/Entities/WebhooksEntityValidator.cs
```diff
@@ -0,0 +1,38 @@
...
+                .WithMessage("There can be only one endpoint with no selector");
+
+            RuleFor(x => x.Endpoints)
+                .Must(NotContainMultipleEndpointsWithTheSameSelector)
+                .WithMessage("There cannot be multiple endpoints with the same selector");
+        }
+
+        private bool ContainAtMostOneEndpointWithNoSelector(IEnumerable<EndpointEntity> endpoints)
+        {
+            return endpoints.Count(x => x.Selector == null) <= 1;
```
chomick, line 27: Any ()?

src/CaptainHook.Application/Handlers/Subscribers/UpsertWebhookRequestHandler.cs
tabman83: I am kinda split between the two, on one side I don't want those entities to be too fat or have the validation in two different places

src/CaptainHook.Application/ApplicationModule.cs
mszymeshop: This has been now changed so that Entities "validate" themselves.

src/CaptainHook.Application/Handlers/Subscribers/UpsertWebhookRequestHandler.cs
mszymeshop: Please take a look at the new version, validation of entities is no longer done on the handler level.

src/CaptainHook.Domain/Entities/SubscriberEntity.cs
```diff
@@ -37,20 +46,28 @@ public void SetParentEvent(EventEntity parentEvent)
         }
 
         /// <summary>
-        /// Adds an enpoint to the list of webhook endpoints
+        /// Adds an endpoint to the list of webhook endpoints
         /// </summary>
         /// <param name="endpointModel"></param>
-        public SubscriberEntity AddWebhookEndpoint(EndpointEntity endpointModel)
+        public OperationResult<SubscriberEntity> AddWebhookEndpoint(EndpointEntity endpointModel)
```
chomick, line 52: Do we have unit tests for that domain model? We may want to have some tests that verify that valid vs invalid 'add' is handled as expected?

mszymeshop: @dawidkeshop I haven't found anything, do you have more info?

chomick: It should not block your PR, though

PR #176, opened 2020-08-10
Return appropriate error when config has not been fully loaded yet

tabman83:
Quick PR to address stuff I stumbled upon while working on something else:
- Return 500 when there's an exception (400 bad request has a different meaning)
- Correctly use ToExceptionEvent to log BB exception
- Return 503 when director has not completed loading of config (config is unavailable, subscribers is null, then we would be getting a 400)
- Set Swagger response descriptions

### DIFF
--- a/src/CaptainHook.Api/Controllers/SubscribersController.cs
+++ b/src/CaptainHook.Api/Controllers/SubscribersController.cs
@@ -39,8 +39,12 @@ namespace CaptainHook.Api.Controllers
         /// Retrieve all subscribers from current configuration
         /// </summary>
         /// <response code="200">Subscribers retrieved properly</response>
+        /// <response code="503">Configuration has not been fully loaded yet</response>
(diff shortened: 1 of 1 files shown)


src/CaptainHook.Api/Controllers/SubscribersController.cs
```diff
@@ -49,14 +53,19 @@ public async Task<IActionResult> GetAll()
                 var directorServiceClient = ServiceProxy.Create<IDirectorServiceRemoting>(directorServiceUri);
                 var subscribers = await directorServiceClient.GetAllSubscribersAsync();
 
+                if(subscribers == null)
+                {
+                    return StatusCode(StatusCodes.Status503ServiceUnavailable);
```
chomick, line 58: Would be useful to get bit more info about those issues so we can diagnose them quicker

PR #179, opened 2020-08-11
#78991 - start using eTags for updates in Cosmos DB

mszymeshop:
### What
Start using eTags for document updates/replacements. Further logic to address failures for updates is going to be addressed in a separate PR.

### DIFF
--- a/src/CaptainHook.Domain/Entities/SubscriberEntity.cs
+++ b/src/CaptainHook.Domain/Entities/SubscriberEntity.cs
@@ -24,6 +24,11 @@ namespace CaptainHook.Domain.Entities
         public string Name { get; }
 
         /// <summary>
+        /// Gets the version of the entity.
+        /// </summary>
+        public string Etag { get; }
+
+        /// <summary>
         /// Parent event for this subscriber
         /// </summary>
         public EventEntity ParentEvent { get; private set; }
@@ -33,10 +38,10 @@ namespace CaptainHook.Domain.Entities
         /// </summary>
         public WebhooksEntity Webhooks { get; private set; }
 
-        public SubscriberEntity(string name) : this(name, null) { }
-        public SubscriberEntity(string name, EventEntity parentEvent)
+        public SubscriberEntity(string name, EventEntity parentEvent = null, string etag = null)
         {
             Name = name;
+            Etag = etag;
             SetParentEvent(parentEvent);
         }
 
--- a/src/CaptainHook.Storage.Cosmos/Models/SubscriberDocument.cs
+++ b/src/CaptainHook.Storage.Cosmos/Models/SubscriberDocument.cs
@@ -46,5 +46,11 @@ namespace CaptainHook.Storage.Cosmos.Models
         /// </summary>
         [JsonProperty("webhooks")]
         public WebhookSubdocument Webhooks { get; set; }
+
+        /// <summary>
+        /// ETag of the document from Cosmos Db
+        /// </summary>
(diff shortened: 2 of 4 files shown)


src/CaptainHook.Domain/Entities/SubscriberEntity.cs
```diff
@@ -23,6 +23,11 @@ public class SubscriberEntity
         /// </summary>
         public string Name { get; }
 
+        /// <summary>
+        /// Gets the version of the entity.
+        /// </summary>
+        public string Etag { get; }
```
mszymeshop, line 29: I have decided to keep the name Etag (instead of Version or Timestamp) as I prefer consistent naming here.

src/CaptainHook.Storage.Cosmos/Models/SubscriberDocument.cs
```diff
@@ -46,5 +46,11 @@ internal class SubscriberDocument
         /// </summary>
         [JsonProperty("webhooks")]
         public WebhookSubdocument Webhooks { get; set; }
+
+        /// <summary>
+        /// ETag of the document from Cosmos Db
+        /// </summary>
+        [JsonProperty("_etag")]
+        public string Etag { get; set; }
```
chomick, line 54: You don't need it here - this is available via other constructs in the repo

src/CaptainHook.Storage.Cosmos/SubscriberRepository.cs
```diff
@@ -184,7 +184,8 @@ private SubscriberDocument Map(SubscriberEntity subscriberEntity)
                 Id = subscriberEntity.Id,
                 EventName = subscriberEntity.ParentEvent.Name,
                 SubscriberName = subscriberEntity.Name,
-                Webhooks = Map(subscriberEntity.Webhooks)
+                Webhooks = Map(subscriberEntity.Webhooks),
+                Etag = subscriberEntity.Etag
```
chomick, line 188: You should not be persisting it

src/CaptainHook.Storage.Cosmos/SubscriberRepository.cs
```diff
@@ -224,7 +225,8 @@ private SubscriberEntity Map(SubscriberDocument subscriberDocument)
 
             var subscriberEntity = new SubscriberEntity(
                 subscriberDocument.SubscriberName,
-                eventEntity);
+                eventEntity,
+                subscriberDocument.Etag);
```
chomick, line 229: If you use _cosmosDbRepository.QueryWithContainerAsync () it has the same functionality but returns your document wrapped in a container (an envelope, if you like:) ) structure that contains etag.
So you can map DocumentContainer<SubscriberDocument> to SubscriberEntity here.

src/CaptainHook.Storage.Cosmos/SubscriberRepository.cs
mszymeshop: Removed this mapping now.

src/CaptainHook.Storage.Cosmos/SubscriberRepository.cs
mszymeshop: Leaving it as it is for now

src/CaptainHook.Storage.Cosmos/Models/SubscriberDocument.cs
mszymeshop: Leaving it as it is for now

PR #180, opened 2020-08-11
#78993 - support retries for Upsert Webhook Request

mszymeshop:
### What
Operations are retried if there is a problem when updating the document in Cosmos (changes in between and eTag check fails). Please note this also applies to the subscriber being removed entirely, in that case, we will re-add it with a single endpoint.

### DIFF
--- a/src/CaptainHook.Application/Handlers/Subscribers/UpsertWebhookRequestHandler.cs
+++ b/src/CaptainHook.Application/Handlers/Subscribers/UpsertWebhookRequestHandler.cs
@@ -1,4 +1,4 @@
-﻿using System;
+using System;
 using System.Threading;
 using System.Threading.Tasks;
 using CaptainHook.Application.Infrastructure.DirectorService;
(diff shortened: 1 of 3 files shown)


src/CaptainHook.Application/Handlers/Subscribers/UpsertWebhookRequestHandler.cs
```diff
@@ -38,13 +51,23 @@ public class UpsertWebhookRequestHandler : IRequestHandler<UpsertWebhookRequest,
...
+                    return await UpdateAsync(request, existingItem.Data);
                 }
 
                 return existingItem.Error;
             }
+
+            try
+            {
+                var executionResult = await Policy
+                    .HandleResult<OperationResult<EndpointDto>>(result => result.Error is CannotUpdateEntityError)
```
chomick, line 65: I think we should have a specific error that deals with stale data exception from the repo - there may be other reasons that not sure we would care about to retry for?

src/CaptainHook.Application/Handlers/Subscribers/UpsertWebhookRequestHandler.cs
```diff
@@ -38,13 +51,23 @@ public class UpsertWebhookRequestHandler : IRequestHandler<UpsertWebhookRequest,
...
+                    .WaitAndRetryAsync(_retrySleepDurations)
+                    .ExecuteAsync(AddOrUpdateEndpointAsync);
+
+                return executionResult;
+            }
             catch (Exception ex)
             {
                 return new UnhandledExceptionError($"Error processing {nameof(UpsertWebhookRequest)}", ex);
```
chomick, line 73: Do we want to log anything here? Or will it be logged elsewhere?

src/CaptainHook.Application/Handlers/Subscribers/UpsertWebhookRequestHandler.cs
```diff
@@ -10,25 +10,38 @@
...
+            _retrySleepDurations = _sleepDurations ?? DefaultRetrySleepDurations;
         }
 
-        public async Task<OperationResult<EndpointDto>> Handle(UpsertWebhookRequest request,
+        public async Task<OperationResult<EndpointDto>> Handle(
+            UpsertWebhookRequest request,
             CancellationToken cancellationToken)
         {
-            try
+            async Task<OperationResult<EndpointDto>> AddOrUpdateEndpointAsync()
```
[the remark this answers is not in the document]
mszymeshop: I am using it as a closure so prefer it to stay here

src/CaptainHook.Application/Handlers/Subscribers/UpsertWebhookRequestHandler.cs
mszymeshop: I want to handle both stale data and an item not found (in case it is deleted in between). If I cannot update I am going to retry. What would be a reason for me to do this for those two cases and ignore for others? The code will only become complex and that part will be rarely used anyway.

src/CaptainHook.Application/Handlers/Subscribers/UpsertWebhookRequestHandler.cs
mszymeshop: yes, I might as well log something here

dawidkeshop: I think would be good to use MediatR functionality and introduce pipeline which catch all error results and logs these.

PR #181, opened 2020-08-12
Override selector replace transform with selection rule and unit test

tabman83:
- Use a case insensitive comparison when finding duplicates in the URI transform replacement dictionary
- Use a case insensitive comparer when building the aggregate dictionary so the subsequent override of the selector key will work as expected in any case
- Override the selector replacement in the URI transform with the selectionRule

### DIFF
--- a/src/CaptainHook.Application.Infrastructure/Mappers/SubscriberEntityToConfigurationMapper.cs
+++ b/src/CaptainHook.Application.Infrastructure/Mappers/SubscriberEntityToConfigurationMapper.cs
@@ -60,9 +60,9 @@ namespace CaptainHook.Application.Infrastructure.Mappers
             var replacements = entity.Webhooks.Endpoints
                 .Where(x => x.UriTransform?.Replace != null)
                 .SelectMany(x => x.UriTransform.Replace)
-                .GroupBy(x => x.Key, StringComparer.OrdinalIgnoreCase)
+                .GroupBy(x => x.Key)
                 .Select(x => x.First())
-                .ToDictionary(x => x.Key, x => x.Value, StringComparer.OrdinalIgnoreCase);
+                .ToDictionary(x => x.Key, x => x.Value);
 
             replacements["selector"] = entity.Webhooks.SelectionRule;
 
--- a/src/CaptainHook.Application.Infrastructure/Mappers/SubscriberEntityToConfigurationMapper.cs
+++ b/src/CaptainHook.Application.Infrastructure/Mappers/SubscriberEntityToConfigurationMapper.cs
@@ -58,6 +58,7 @@ namespace CaptainHook.Application.Infrastructure.Mappers
         private async Task<SubscriberConfiguration> MapForUriTransformAsync(SubscriberEntity entity)
         {
             var replacements = entity.Webhooks.Endpoints
+                .Where(x => x.UriTransform?.Replace != null)
                 .SelectMany(x => x.UriTransform.Replace)
                 .GroupBy(x => x.Key, StringComparer.OrdinalIgnoreCase)
                 .Select(x => x.First())
--- a/src/CaptainHook.Application.Infrastructure/Mappers/SubscriberEntityToConfigurationMapper.cs
+++ b/src/CaptainHook.Application.Infrastructure/Mappers/SubscriberEntityToConfigurationMapper.cs
@@ -1,3 +1,4 @@
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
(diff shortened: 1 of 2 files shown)


src/CaptainHook.Application.Infrastructure/Mappers/SubscriberEntityToConfigurationMapper.cs
```diff
@@ -56,13 +57,13 @@ private async Task<SubscriberConfiguration> MapSingleWebhookWithNoUriTransformAs
...
+            var replacements = entity.Webhooks.Endpoints
                 .SelectMany(x => x.UriTransform.Replace)
-                .GroupBy(x => x.Key)
-                .Select(x => x.First()))
-            {
-                { "selector", entity.Webhooks.SelectionRule }
-            };
+                .GroupBy(x => x.Key, StringComparer.OrdinalIgnoreCase)
+                .Select(x => x.First())
+                .ToDictionary(x => x.Key, x => x.Value, StringComparer.OrdinalIgnoreCase);
```
chomick, line 64: Are vars case insensitive?

src/Tests/CaptainHook.Application.Tests/Infrastructure/SubscriberEntityToConfigurationMapperTests.cs
```diff
@@ -212,5 +212,62 @@ public async Task MapSubscriberAsync_WithMultipleWebhooksAndNoDefaultSelectorAnd
...
+                result.Should().HaveCount(1);
+
+                var subscriberConfiguration = result.Single();
+                subscriberConfiguration.Should().NotBeNull();
+                subscriberConfiguration.SubscriberName.Should().Be("captain-hook");
+                subscriberConfiguration.EventType.Should().Be("event");
+                subscriberConfiguration.Uri.Should().BeNull();
+                subscriberConfiguration.WebhookRequestRules.Count.Should().Be(1);
+
+                var rule = subscriberConfiguration.WebhookRequestRules.Single();
```
mszymeshop, line 245: These are so difficult to read, I would consider object comparison instead

src/CaptainHook.Application.Infrastructure/Mappers/SubscriberEntityToConfigurationMapper.cs
```diff
@@ -56,13 +57,13 @@ private async Task<SubscriberConfiguration> MapSingleWebhookWithNoUriTransformAs
 
         private async Task<SubscriberConfiguration> MapForUriTransformAsync(SubscriberEntity entity)
         {
-            var replacements = new Dictionary<string, string>(entity.Webhooks.Endpoints
+            var replacements = entity.Webhooks.Endpoints
                 .SelectMany(x => x.UriTransform.Replace)
```
chomick, line 61: You may get with Null UriTransforms here, the case is:
- selectionRule defined
- selector for an endpoint defined ("*")
- no uri transform in DTO for the endpoint

There is no validation that will fail for such case but it will blow up here with null

chomick: just do x.UriTransform?.Replace ?? Enumerable.Empty<...>

src/CaptainHook.Application.Infrastructure/Mappers/SubscriberEntityToConfigurationMapper.cs
mszymeshop: Currently `BuildUriContext` is case sensitive for replacements.

chomick: ok, Nino that means we need to have exact matches for keys for now

src/Tests/CaptainHook.Application.Tests/Infrastructure/SubscriberEntityToConfigurationMapperTests.cs
chomick: I agree that they are difficult to read and rather long.
I was wondering also if we should target specific cases/mappings in the tests rather then the whole thing. 

We should have separate tests to verify authentication mapping.
Separate test cases for for endpoint mapping (that simply verifies that authentication mapping is invoked but we don't have to test all cases again here).
Separate cases for entire webhook mapping - again focus on multiple vs single endpoints and selection rules etc but do not worry about details for endpoints or authentications...

This is some refactoring to be made and beyond scope of this story/PR. I suggest that we create a tech debt item linked to this so we can refactor a bit our mapping functionality as well as tests covering it.

src/CaptainHook.Application.Infrastructure/Mappers/SubscriberEntityToConfigurationMapper.cs
tabman83: I believe they should be, to avoid confusion.

src/Tests/CaptainHook.Application.Tests/Infrastructure/SubscriberEntityToConfigurationMapperTests.cs
tabman83: This is gonna affect other tests there, so out of scope for this, leaving for another PR

src/CaptainHook.Application.Infrastructure/Mappers/SubscriberEntityToConfigurationMapper.cs
tabman83: Changed and added unit test

src/CaptainHook.Application.Infrastructure/Mappers/SubscriberEntityToConfigurationMapper.cs
chomick: Nino, I think we need to change it so StringComparer.OrdinalIgnoreCase is NOT used :)

PR #184, opened 2020-08-12
Task 80130 extend validation to selection rule

mszymeshop:
### What
The validation for SelectionRule has been added now. It was missing before and validation focused on endpoints only.


src/CaptainHook.Application/Handlers/Subscribers/UpsertWebhookRequestHandler.cs
```diff
@@ -10,6 +10,7 @@
 using CaptainHook.Domain.Results;
 using CaptainHook.Domain.ValueObjects;
 using Eshopworld.Core;
+using Kusto.Cloud.Platform.Utils;
```
tabman83, line 13: Is it needed?

src/CaptainHook.Application/Handlers/Subscribers/UpsertWebhookRequestHandler.cs
```diff
@@ -34,12 +35,12 @@ public class UpsertWebhookRequestHandler : IRequestHandler<UpsertWebhookRequest,
...
             IDirectorServiceProxy directorService,
             IBigBrother bigBrother,
-            TimeSpan[] _sleepDurations = null)
+            TimeSpan[] sleepDurations = null)
         {
             _subscriberRepository = subscriberRepository;
             _directorService = directorService;
             _bigBrother = bigBrother;
-            _retrySleepDurations = _sleepDurations ?? DefaultRetrySleepDurations;
+            _retrySleepDurations = sleepDurations?.SafeFastNullIfEmpty() ?? DefaultRetrySleepDurations;
```
tabman83, line 43: SafeFastNullIfEmpty - What is that?

mszymeshop: it returns null if the collection is empty and by default, it is empty (by IOC) and I am using this extension method from Kusto here to address this.

src/CaptainHook.Application/Handlers/Subscribers/UpsertWebhookRequestHandler.cs
mszymeshop: See the comment below.

src/CaptainHook.Application/Handlers/Subscribers/UpsertWebhookRequestHandler.cs
chomick: Are we using entire package only for that extension? :)

PR #186, opened 2020-08-13
78999 endpoint delete

dawidkeshop:


src/CaptainHook.Domain/Entities/WebhooksEntity.cs
```diff
@@ -38,5 +38,14 @@ public void AddEndpoint(EndpointEntity endpointModel)
         {
             _endpoints.Add(endpointModel);
         }
+
+        public bool RemoveEndpoint(string selector)
+        {
+            var toDelete = selector == null
```
mszymeshop, line 44: worth checking if `string.Equals(x.Selector, selector, String...)` will yield the same result and simplify at the same time

src/CaptainHook.Application/Validators/DeleteWebhookRequestValidator.cs
```diff
@@ -0,0 +1,17 @@
...
+{
+    public class DeleteWebhookRequestValidator : AbstractValidator<DeleteWebhookRequest>
+    {
+        public DeleteWebhookRequestValidator()
+        {
+            CascadeMode = CascadeMode.StopOnFirstFailure;
+
+            RuleFor(x => x.EventName).NotEmpty();
+            RuleFor(x => x.SubscriberName).NotEmpty();
+            RuleFor(x => x.Selector).NotEmpty();
```
mszymeshop, line 14: Selector can be empty for the default endpoint

dawidkeshop: Right

src/CaptainHook.Domain/Entities/WebhooksEntity.cs
dawidkeshop: Nice idea, thanks

src/CaptainHook.Application/Handlers/Subscribers/DeleteWebhookRequestHandler.cs
```diff
@@ -0,0 +1,116 @@
...
+
+        public DeleteWebhookRequestHandler(
+            ISubscriberRepository subscriberRepository,
+            IDirectorServiceProxy directorService,
+            IBigBrother bigBrother,
+            TimeSpan[] sleepDurations = null)
+        {
+            _subscriberRepository = subscriberRepository;
+            _directorService = directorService;
+            _bigBrother = bigBrother;
```
tabman83, line 42: ?? new ArgumentNullException(...)

src/CaptainHook.Application/Handlers/Subscribers/DeleteWebhookRequestHandler.cs
```diff
@@ -0,0 +1,116 @@
...
+                    .HandleResult<OperationResult<SubscriberDto>>(result => result.Error is CannotUpdateEntityError)
+                    .WaitAndRetryAsync(_retrySleepDurations)
+                    .ExecuteAsync(() => DeleteEndpointAsync(request));
+
+                return executionResult;
+            }
+            catch (Exception ex)
+            {
+                _bigBrother.Publish(ex.ToExceptionEvent());
+                return new UnhandledExceptionError($"Error processing {nameof(UpsertWebhookRequest)}", ex);
```
mszymeshop, line 60: wrong type for the error

src/CaptainHook.Application/Handlers/Subscribers/DeleteWebhookRequestHandler.cs
```diff
@@ -0,0 +1,116 @@
...
+            {
+                return saveResult.Error;
+            }
+
+            return MapToDto(saveResult);
+        }
+
+        private SubscriberDto MapToDto(SubscriberEntity subscriber)
+        {
+            return new SubscriberDto
```
mszymeshop, line 97: We will need a better mapping, there is a separate task for this.

src/Tests/CaptainHook.Application.Tests/Handlers/DeleteWebhookRequestHandlerTests.cs
```diff
@@ -0,0 +1,294 @@
...
+using Eshopworld.Core;
+using Eshopworld.Tests.Core;
+using FluentAssertions;
+using FluentAssertions.Execution;
+using Moq;
+using Xunit;
+
+namespace CaptainHook.Application.Tests.Handlers
+{
+    public class DeleteWebhookRequestHandlerTests
```
mszymeshop, line 24: Consider the approach I have taken in `UpsertSubscriberRequestHandlerTests` where I mark my mocks as Strict which allows me to control the flow of tests in a slightly different manner. Effectively I don't have to verify my mocked calls since all invalid ones are going to be rejected. This can simplify the test at the same time reducing the number of lines of code.

src/Tests/CaptainHook.Application.Tests/Handlers/DeleteWebhookRequestHandlerTests.cs
```diff
@@ -0,0 +1,294 @@
...
+        }
+
+        [Fact, IsUnit]
+        public async Task When_RepositoryThrowsAnExceptionSavingSubscriberAfterSuccessfulReaderCreation_Then_UnhandledExceptionErrorReturned()
+        {
+            _repositoryMock.Setup(x => x.UpdateSubscriberAsync(It.IsAny<SubscriberEntity>())).ThrowsAsync(new Exception());
+
+            var result = await Handler.Handle(_defaultRequest, CancellationToken.None);
+
+            using var scope = new AssertionScope();
```
mszymeshop, line 251: Check out the `BeEquivalentTo` from `UpsertSubscriberRequestHandlerTests`, maybe something worth applying.

src/Tests/CaptainHook.Application.Tests/RequestValidators/DeleteWebhookRequestValidatorTests.cs
```diff
@@ -0,0 +1,49 @@
...
+        [Theory, IsUnit]
+        [InlineData("selector")]
+        [ClassData(typeof(EmptyStrings))]
+        public void When_RequestIsValid_Then_NoFailuresReturned(string validSelector)
+        {
+            var request = new DeleteWebhookRequest("event", "subscriber", validSelector);
+
+            var result = _validator.TestValidate(request);
+
+            AssertionExtensions.Should(result.IsValid).BeTrue();
```
mszymeshop, line 24: there is an extension method `ShouldNotHaveValidationErrors()`

src/Tests/CaptainHook.Application.Tests/RequestValidators/DeleteWebhookRequestValidatorTests.cs
```diff
@@ -0,0 +1,49 @@
...
+namespace CaptainHook.Application.Tests.RequestValidators
+{
+    public class DeleteWebhookRequestValidatorTests
+    {
+        private readonly DeleteWebhookRequestValidator _validator = new DeleteWebhookRequestValidator();
+
+        [Theory, IsUnit]
+        [InlineData("selector")]
+        [ClassData(typeof(EmptyStrings))]
+        public void When_RequestIsValid_Then_NoFailuresReturned(string validSelector)
```
mszymeshop, line 18: Whether the white spaces should be allowed as a selector is debatable. We should be careful here as selector is our identifier. I suggest we allow `null`, but we don't allow `""` and `"   "`

src/Tests/CaptainHook.Application.Tests/Handlers/DeleteWebhookRequestHandlerTests.cs
mszymeshop, line 24: Tests aren't in the right namespace

src/CaptainHook.Application/Handlers/Subscribers/DeleteWebhookRequestHandler.cs
dawidkeshop: None of our handlers have such check. :)

src/CaptainHook.Application/Handlers/Subscribers/DeleteWebhookRequestHandler.cs
dawidkeshop: Why wrong? Is the same as in the other handlers.
What is wrong for me, is that we have this logic duplicated instead of use dedicated Pipeline to deal with unhandled exceptions.

src/CaptainHook.Application/Handlers/Subscribers/DeleteWebhookRequestHandler.cs
dawidkeshop: Definitely!

src/Tests/CaptainHook.Application.Tests/RequestValidators/DeleteWebhookRequestValidatorTests.cs
dawidkeshop: TBH, I have no idea where it comes form. It's not a way how I wrote assertions.

src/Tests/CaptainHook.Application.Tests/RequestValidators/DeleteWebhookRequestValidatorTests.cs
dawidkeshop: OK

src/CaptainHook.Application/Handlers/Subscribers/DeleteWebhookRequestHandler.cs
mszymeshop: I mean you have `nameof(UpsertWebhookRequest)` while you should have `nameof(DeleteWebhookRequest)`. As for your other remark, how can we address this duplication?

src/CaptainHook.Application/Handlers/Subscribers/DeleteWebhookRequestHandler.cs
mszymeshop: I'm taking care of that as a separate task.

src/Tests/CaptainHook.Application.Tests/RequestValidators/DeleteWebhookRequestValidatorTests.cs
mszymeshop: I don't see any changes here

src/Tests/CaptainHook.Application.Tests/RequestValidators/DeleteWebhookRequestValidatorTests.cs
mszymeshop: I don't see any changes here

src/Tests/CaptainHook.Application.Tests/RequestValidators/DeleteWebhookRequestValidatorTests.cs
mszymeshop: ok all good

src/CaptainHook.Application/Handlers/Subscribers/DeleteWebhookRequestHandler.cs
mszymeshop: I guess the changes for behaviour for error handling will address this so happy to accept this for now

PR #189, opened 2020-08-13
#78998 - create delete support for endpoints in EventController

mszymeshop:
### What
Adds two DELETE actions for endpoints. They are separated as optional parameters are not supported by swagger and autorest.


src/CaptainHook.Api/Controllers/EventsController.cs
```diff
@@ -69,6 +69,59 @@ public async Task<IActionResult> PutWebhook([FromRoute] string eventName, [FromR
...
+                    EndpointNotFoundInSubscriberError endpointNotFound => BadRequest(endpointNotFound),
+                    EntityNotFoundError entityNotFoundError => NotFound(entityNotFoundError),
+                    DirectorServiceIsBusyError directorServiceIsBusyError => Conflict(directorServiceIsBusyError),
+                    ReaderDeleteError readerDeleteError => UnprocessableEntity(readerDeleteError),
+                    ReaderDoesNotExistError readerDoesNotExistError => UnprocessableEntity(readerDoesNotExistError),
+                    CannotUpdateEntityError cannotUpdateEntityError => StatusCode(StatusCodes.Status500InternalServerError, cannotUpdateEntityError),
+                    _ => StatusCode(StatusCodes.Status500InternalServerError, error)
+                },
+                subscriberDto => Created($"/{eventName}/subscriber/{subscriberName}/webhooks/endpoint/{selector}", subscriberDto)
+            );
```
tabman83, line 122: Is this correct?

mszymeshop: What exactly? I think I am missing the meaning of the comment.

tabman83: Taking it offline

mszymeshop: Thanks Nino, I have now changed this to Ok as this is in sync with requirements.

dawidkeshop:
But we should reuse these error -> HTTP response mappings soon, probably once we unify errors response format.

PR #190, opened 2020-08-13
Simplify some unit tests for UpsertWebhookRequestHandlerTests

mszymeshop:
### What
Simplify some tests after going through https://github.com/eShopWorld/captain-hook/pull/186

### DIFF
--- a/src/Tests/CaptainHook.Application.Tests/Handlers/UpsertWebhookRequestHandlerTests.cs
+++ b/src/Tests/CaptainHook.Application.Tests/Handlers/UpsertWebhookRequestHandlerTests.cs
@@ -1,5 +1,4 @@
 ﻿using System;
-using System.Collections.Generic;
 using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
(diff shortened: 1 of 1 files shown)


src/Tests/CaptainHook.Application.Tests/Handlers/UpsertWebhookRequestHandlerTests.cs
```diff
@@ -31,6 +30,20 @@ public class UpsertWebhookRequestHandlerTests
         private readonly UpsertWebhookRequest _defaultUpsertRequest =
             new UpsertWebhookRequest("event", "subscriber", new EndpointDtoBuilder().With(x => x.Selector, null).Create());
 
+        private static readonly SubscriberBuilder DefaultSubscriberBuilder = new SubscriberBuilder().WithEvent("event")
```
dawidkeshop, line 33: Shouldn't it be `_defaultSubscriberBuilder` as this is a private field?

mszymeshop: I don't think that is the case for static, we should come up with something regarding coding standards finally

mszymeshop:
> I would also move default "valid" mock setups to a constructor, it would decrease amount of repeatable code in every test.

@dawidkeshop my test cases are different in each test, so there is no room for such a change

PR #192, opened 2020-08-14
Skip refresh config on builds

nikhilgirraj:
The CI deployment fails due to failed integration tests.

### DIFF
--- a/src/Tests/CaptainHook.Api.Tests/Api/RefreshConfigControllerTests.cs
+++ b/src/Tests/CaptainHook.Api.Tests/Api/RefreshConfigControllerTests.cs
@@ -17,7 +17,8 @@ namespace CaptainHook.Api.Tests
         {
         }
 
-        [Fact, IsIntegration]
+
+        [Fact(Skip = "A successful RefreshConfig makes the other tests fail until refresh is complete"), IsIntegration]
         public async Task RefreshConfig_WhenAuthenticated_Returns202Accepted()
         {
             var result = await AuthenticatedClient.ReloadConfigurationWithHttpMessagesAsync();
--- a/src/Tests/CaptainHook.Api.Tests/Api/RefreshConfigControllerTests.cs
+++ b/src/Tests/CaptainHook.Api.Tests/Api/RefreshConfigControllerTests.cs
@@ -18,7 +18,7 @@ namespace CaptainHook.Api.Tests
         }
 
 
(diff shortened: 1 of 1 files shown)


nikhilgirraj:
I like the idea of polling for a certain time after config refresh @tabman83.

nikhilgirraj:
Tested with a release on CI. ✔️

nikhilgirraj:
On second thoughts about Polling till the refresh is complete - wouldn't the other tests still fail since they might be started in parallel? @tabman83

tabman83:
> On second thoughts about Polling till the refresh is complete - wouldn't the other tests still fail since they might be started in parallel? @tabman83

You can use collections, see https://xunit.net/docs/running-tests-in-parallel.html

mszymeshop:
I have doubts regarding this, maybe I need a bit of explanation. Should we control the parallel run of the tests instead, as per @tabman83 link?

nikhilgirraj:
@mszymeshop I checked, the tests do not run in parallel so far. See [src/Tests/CaptainHook.Api.Tests/config/ApiClientCollection.cs](https://github.com/eShopWorld/captain-hook/blob/master/src/Tests/CaptainHook.Api.Tests/config/ApiClientCollection.cs).

I have added a Polly retry for when there a 409 conflict in tests. @tabman83

src/Tests/CaptainHook.Api.Tests/Api/ControllerTestBase.cs
```diff
@@ -1,21 +1,27 @@
...
+        protected readonly AsyncRetryPolicy<HttpOperationResponse> ConflictRetryPolicy;
 
         public ControllerTestBase(ApiClientFixture testFixture)
         {
             AuthenticatedClient = testFixture.GetApiClient();
             UnauthenticatedClient = testFixture.GetApiUnauthenticatedClient();
+
+            ConflictRetryPolicy = Policy /* poll until no conflict */
+                .HandleResult<HttpOperationResponse>(msg => msg.Response.StatusCode == HttpStatusCode.Conflict)
+                .WaitAndRetryAsync(30, i => TimeSpan.FromMilliseconds(2000));
```
tabman83, line 24: FromSeconds(2) is more readable.

Maybe we can even check every 5 sec
CC @mszymeshop

nikhilgirraj:
Reverted the changes for now to make the CI green.

PR #194, opened 2020-08-14
#78998 - create delete support for endpoints in EventController

mszymeshop:


src/CaptainHook.Api/Controllers/EventsController.cs
```diff
@@ -69,6 +69,59 @@ public async Task<IActionResult> PutWebhook([FromRoute] string eventName, [FromR
...
+        [ProducesResponseType(typeof(ErrorBase), StatusCodes.Status500InternalServerError)]
+        public async Task<IActionResult> DeleteWebhook([FromRoute] string eventName, [FromRoute] string subscriberName, [FromRoute] string selector)
+        {
+            var request = new DeleteWebhookRequest(eventName, subscriberName, selector);
+            var result = await _mediator.Send(request);
+
+            return result.Match<IActionResult>(
+                error => error switch
+                {
+                    ValidationError validationError => BadRequest(validationError),
```
chomick, line 111: This code starts to get lots of duplication :) any way we can share it across different endpoints? I suspect the same error type is always mapped the same way to the response?

PR #195, opened 2020-08-14
Task 80360 map subscriber to dto

mszymeshop:


src/CaptainHook.Application/Handlers/Subscribers/DeleteWebhookRequestHandler.cs
```diff
@@ -110,6 +112,19 @@ private EndpointDto MapEndpointDto(EndpointEntity endpointEntity)
                 Selector = endpointEntity.Selector,
                 Uri = endpointEntity.Uri,
                 HttpVerb = endpointEntity.HttpVerb,
+                UriTransform = new UriTransformDto { Replace = endpointEntity.UriTransform.Replace },
+                Authentication = new AuthenticationDto
```
dawidkeshop, line 116: I think this code might be reusable in several places, wouldn't be better to extract it to a separate Mapper class?
Or use AutoMapper?

chomick: Do we already have examples of duplication? I would expect endpoint mapping to be the only place we do that?

chomick: Sorry - just noticed this is for individual Handler - absolutely! This should be a shared functionality - I suggest this is extracted and shared across all handlers.

mszymeshop: I feel like we have already talked about this numerous times. :+1: to use AutoMapper. I have added a task like that as a suggestion for I&A.

mszymeshop:
@chomick I don't know which duplication you're talking about since this is SubscriberEntity -> SubscriberDto which we didn't have so far. However, this is obviously an area that can be handled outside of Handlers and reused where possible. Let's make sure that we don't reinvent the wheel.

PR #198, opened 2020-08-15
Update packages and SF packages

tabman83:


tabman83:
I haven't added anyone as reviewer yet and left it as a draft because I created this to test whether we could upgrade the SF packages to the latest version. It obviously works on our build machine because I have updated the VS build tools but it does not work on other build machines maintained by devops which haven't been updated yet. The failing builds are those against other build machines, just to prove this point.

tabman83:
Closing as not needed anymore

PR #201, opened 2020-08-18
Task 80579 extend ch api with reload status

mszymeshop:
### What
We have extended the APi with api/config/status endpoint which returns 200 if config has been loaded and 202 if the load is in progress. The choice of status codes was arbitrary and discussed within the team (also documented in UI).

### Why
The status code is going to be used in integration tests to wait for the reload to finish. It has been decided by the team that it is a better decision to extend the API than address the problem in another way.

### DIFF
--- a/src/CaptainHook.Api/Startup.cs
+++ b/src/CaptainHook.Api/Startup.cs
@@ -91,16 +91,19 @@ namespace CaptainHook.Api
                 _configuration.GetSection("ServiceConfigurationOptions").Bind(serviceConfiguration);
 
                 services.AddControllers(options =>
-                {
-                    var policy = ScopePolicy.Create(serviceConfiguration.RequiredScopes?.ToArray() ?? new string[0]);
+                    {
+                        var policy =
+                            ScopePolicy.Create(serviceConfiguration.RequiredScopes?.ToArray() ?? new string[0]);
 
-                    var filter = EnvironmentHelper.IsInFabric ?
-                        (IFilterMetadata)new AuthorizeFilter(policy) :
-                        new AllowAnonymousFilter();
+                        var filter = EnvironmentHelper.IsInFabric
+                            ? (IFilterMetadata) new AuthorizeFilter(policy)
+                            : new AllowAnonymousFilter();
 
-                    options.Filters.Add(filter);
-
-                }).AddNewtonsoftJson();
+                        options.Filters.Add(filter);
+                    })
+                    .AddNewtonsoftJson()
+                    .ConfigureApiBehaviorOptions(options =>
+                        options.SuppressMapClientErrors = true);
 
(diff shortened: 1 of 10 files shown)


src/CaptainHook.Api/Startup.cs
```diff
@@ -91,16 +91,19 @@ public void ConfigureServices(IServiceCollection services)
...
+
+                        var filter = EnvironmentHelper.IsInFabric
+                            ? (IFilterMetadata) new AuthorizeFilter(policy)
+                            : new AllowAnonymousFilter();
+
+                        options.Filters.Add(filter);
+                    })
+                    .AddNewtonsoftJson()
+                    .ConfigureApiBehaviorOptions(options =>
+                        options.SuppressMapClientErrors = true);
```
tabman83, line 106: Why do we need this?

mszymeshop: It is needed so that the DetailsProblem is not exposed in the API (which is visible in both Swagger but especially AutoRest) - https://docs.microsoft.com/en-us/aspnet/core/web-api/?view=aspnetcore-3.1#disable-problemdetails-response

nikhilgirraj: Screenshot for example - :)

image

src/CaptainHook.Application.Infrastructure/DirectorService/Remoting/ReloadConfigurationStatus.cs
```diff
@@ -0,0 +1,8 @@
+﻿namespace CaptainHook.Application.Infrastructure.DirectorService.Remoting
+{
+    public enum ReloadConfigurationStatus
+    {
+        Loaded = 0,
```
nikhilgirraj, line 5: @mszymeshop  I know we have gone over this but I had another idea for the enum - { Ready, Busy }. We don't have to change unless this sounds way better.

mszymeshop: I like it, will you make the change?

src/CaptainHook.Api/Startup.cs
tabman83: Thanks for the clarification!

src/CaptainHook.Application.Infrastructure/DirectorService/Remoting/ReloadConfigurationStatus.cs
tabman83: Since the enum is called `ReloadConfigurationStatus` I would rather have `{ Complete, InProgress }` but guys, honestly, this is such a minor change that I don't want to be splitting hairs.

PR #204, opened 2020-08-19
#80970 move URITransform from Endpoint to Webhooks

tabman83:


src/Tests/CaptainHook.Application.Tests/RequestValidators/UpsertWebhookRequestValidatorTests.cs
```diff
@@ -117,55 +117,12 @@ public void When_UriTransformIsValid_Then_NoFailuresReturned()
...
                 .Create();
             var request = new UpsertWebhookRequest("event", "subscriber", dto);
 
             var result = _validator.TestValidate(request);
 
             result.ShouldNotHaveAnyValidationErrors();
         }
-
-        [Fact, IsUnit]
-        public void When_UriTransformRoutesAreEmpty_Then_ValidationFails()
```
mszymeshop, line 130: These tests are removed but it doesn't look like new ones are added in their place.

mszymeshop (changes requested):

src/Tests/CaptainHook.Application.Tests/RequestValidators/UpsertWebhookRequestValidatorTests.cs
tabman83: Thanks @mszymeshop, I am aware, I just wanted to give an opportunity to people to start reviewing since it's such a big PR.

mszymeshop (approved):

nikhilgirraj (approved):

nikhilgirraj (approved):

src/CaptainHook.Storage.Cosmos/Models/EndpointDocument.cs
```diff
@@ -1,12 +1,11 @@
 ﻿using Newtonsoft.Json;
-using Newtonsoft.Json.Converters;
 
 namespace CaptainHook.Storage.Cosmos.Models
 {
     /// <summary>
     /// Endpoint model in Cosmos DB
     /// </summary>
-    internal class EndpointSubdocument
+    internal class EndpointDocument
```
chomick, line 8: Are these becoming stand-alone documents now?

tabman83: Because the UriTransforDocument was called like that when I started working on this PR, ([see here](https://github.com/eShopWorld/captain-hook/blob/0b1a0c1de2029a3a8947f8397cf492ac2013fb0d/src/CaptainHook.Storage.Cosmos/Models/UriTransformDocument.cs#L7)) I assumed we wanted to keep it consistent with that. However I see it is more meaningful to use -subdocument so I have renamed those models.

PR #205, opened 2020-08-19
Return NoContent when deleting an endpoint

tabman83:
Return NoContent when deleting an endpoint as per Swagger attribute.


src/CaptainHook.Api/Controllers/EventsController.cs
```diff
@@ -117,7 +117,7 @@ public async Task<IActionResult> DeleteWebhook([FromRoute] string eventName, [Fr
                     CannotUpdateEntityError cannotUpdateEntityError => StatusCode(StatusCodes.Status500InternalServerError, cannotUpdateEntityError),
                     _ => StatusCode(StatusCodes.Status500InternalServerError, error)
                 },
-                Ok
```
mszymeshop, line 120: Nino, what do the requirements say? I was returning OK here as the entity was returned along with it. Your change is correct as you have no content. But frankly, both are ok so it boils down to requirements. I cannot seem to be able to access ADO today to verify myself.

tabman83: We either change that or the Swagger attribute. I'd say NoContent looks correct as we are not returning anything back, whatever the reqs says we can agree on the most correct approach here anyway.

tabman83: Sorry I mis-interpreted your comment. I see that the handler is returning the updated entity now. Reqs on ADO do not mention the result here. So it would make sense to me to return the entity (as you were doing) and just change the Swagger attribute. Let me know your opinion @mszymeshop and I can change that here.

tabman83:
Closing this one as we are gonna change the Swagger attribute to reflect the requirements.

PR #207, opened 2020-08-20
Bug 77982 stale auth header after reload

nikhilgirraj:
**What**
Issue - Bug 77982: Reload on Authentication changes does not work
1. Added unit test to replicate the bug. 
2. Added `HasConfigChanged` method to `IAuthenticationHandler` interface
3. *Adding a check for `HasConfigChanged` in `AuthenticationHandlerFactory`

**Why draft PR**
1. Should I keep these changes or implement an IEqualityComparer?
2. Thoughts on extending the unit tests for OIDC auth handler as well?


src/CaptainHook.EventHandlerActor/Handlers/Authentication/BasicAuthenticationHandler.cs
```diff
@@ -33,5 +34,16 @@ public virtual async Task<string> GetTokenAsync(CancellationToken cancellationTo
 
             return await Task.FromResult(token);
         }
+
+        public bool HasConfigChanged(AuthenticationConfig newConfig)
+        {
+            if (newConfig?.Type == this.BasicAuthenticationConfig.Type && newConfig is BasicAuthenticationConfig authConfig)
```
mszymeshop, line 40: Is the Type property check necessary? The object type should be enough.

src/CaptainHook.EventHandlerActor/Handlers/Authentication/OidcAuthenticationHandler.cs
```diff
@@ -83,6 +85,23 @@ public virtual async Task<string> GetTokenAsync(CancellationToken cancellationTo
             return $"Bearer {OidcAuthenticationToken.AccessToken}";
         }
 
+        public bool HasConfigChanged(AuthenticationConfig newConfig)
+        {
+            if (newConfig is OidcAuthenticationConfig oidcNewConfig)
```
mszymeshop, line 90: You are not checking the type here so whatever is decided I think this should be consistent with BasicAuthenticationConfig.

src/CaptainHook.EventHandlerActor/Handlers/Authentication/OidcAuthenticationHandler.cs
```diff
@@ -83,6 +85,23 @@ public virtual async Task<string> GetTokenAsync(CancellationToken cancellationTo
...
+            {
+                var equality = oidcNewConfig.Uri == OidcAuthenticationConfig.Uri
+                               && oidcNewConfig.ClientId == OidcAuthenticationConfig.ClientId
+                               && oidcNewConfig.ClientSecret == OidcAuthenticationConfig.ClientSecret
+                               && oidcNewConfig.GrantType == OidcAuthenticationConfig.GrantType
+                               && oidcNewConfig.RefreshBeforeInSeconds == OidcAuthenticationConfig.RefreshBeforeInSeconds
+                               && oidcNewConfig.Scopes.Length == OidcAuthenticationConfig.Scopes.Length
```
mszymeshop, line 97: Use of `SequenceEqual` might be clearer here.

src/Tests/CaptainHook.Tests/Web/Authentication/AuthenticationFactoryTests.cs
```diff
@@ -61,5 +75,25 @@ public async Task NoAuthentication(WebhookConfig config)
...
             Assert.Null(handler);
         }
+
+        /// <summary>
+        /// Checks that the auth token changes with any change in basic auth params
+        /// </summary>
+        /// <param name="webhookConfigs"></param>
+        [IsUnit]
+        [Theory]
+        [MemberData(nameof(AuthenticationChangeTestData))]
```
mszymeshop, line 85: At the moment this test does not constitute a theory, do you plan on extending it? You have only a single test case here.

src/Tests/CaptainHook.Tests/Web/Authentication/AuthenticationFactoryTests.cs
```diff
@@ -61,5 +75,25 @@ public async Task NoAuthentication(WebhookConfig config)
...
+        public async Task ChangeAuthentication(IEnumerable<WebhookConfig> webhookConfigs)
+        {
+            var factory = new AuthenticationHandlerFactory(new HttpClientFactory(), new Mock<IBigBrother>().Object);
+            string lastToken = null;
+            foreach (WebhookConfig webhookConfig in webhookConfigs)
+            {
+                var handler = await factory.GetAsync(webhookConfig, CancellationToken.None);
+                var token = await handler.GetTokenAsync(CancellationToken.None);
+                Assert.NotEqual(token, lastToken);
```
mszymeshop, line 94: As long as it does its job I am thinking that this test is hard to reason about when a failure happens. If it happens for the second token you will not know that.

What I would suggest you do is create all 4 tokens and put in an array. Then compare that array with the expected array of tokens. Benefits: clear identification which element in the array is incorrect. Also you get to do a single assert for an entire test.

src/Tests/CaptainHook.Tests/Web/Authentication/AuthenticationFactoryTests.cs
nikhilgirraj: Yes, I was trying to add another set of tests in there. I'll remove it since the OIDC requires a much different flow. Working on it at the moment.

src/CaptainHook.EventHandlerActor/Handlers/Authentication/BasicAuthenticationHandler.cs
nikhilgirraj: The type check indeed isn't necessary. Removing it.

src/CaptainHook.EventHandlerActor/Handlers/Authentication/OidcAuthenticationHandler.cs
nikhilgirraj: yep.

src/Tests/CaptainHook.Tests/Web/Authentication/AuthenticationFactoryTests.cs
nikhilgirraj: Makes sense. Please have a look at final changes.

nikhilgirraj:
Also, do I create a real PR instead of draft? Or is this one okay to merge?

src/Tests/CaptainHook.Tests/Web/Authentication/AuthenticationFactoryTests.cs
```diff
@@ -61,5 +78,132 @@ public async Task NoAuthentication(WebhookConfig config)
 
             Assert.Null(handler);
         }
+
+        /// <summary>
+        /// Checks that the auth token changes with any change in basic auth params
+        /// </summary>
+        [Fact, IsUnit]
+        public async Task ChangeBasicAuthentication()
+        {
```
tabman83, line 87: Please use this convention for test names:
`When_StateUnderTest_Expect_ExpectedBehavior`
as it's the one we aim to spread across the codebase

tabman83 (changes requested):

src/Tests/CaptainHook.Tests/Web/Authentication/AuthenticationFactoryTests.cs
nikhilgirraj: Changing.

nikhilgirraj: Changed.

src/Tests/CaptainHook.Tests/Web/Authentication/AuthenticationFactoryTests.cs
```diff
@@ -37,7 +55,7 @@ public class AuthenticationFactoryTests
         [IsUnit]
         [Theory]
         [MemberData(nameof(AuthenticationTestData))]
-        public async Task GetTokenProvider(WebhookConfig config, IAuthenticationHandler expectedHandler)
+        public async Task AuthenticationHandler_Returns_RespectiveHandlerForConfigType(WebhookConfig config, IAuthenticationHandler expectedHandler)
         {
             var factory = new AuthenticationHandlerFactory(new HttpClientFactory(), new Mock<IBigBrother>().Object);
```
tabman83, line 61: The bigbrother mock is something you can declare at the beginning and reuse across all of your tests, just to make them more readable. See example from other test classes.

src/Tests/CaptainHook.Tests/Web/Authentication/AuthenticationFactoryTests.cs
```diff
@@ -37,7 +55,7 @@ public class AuthenticationFactoryTests
         [IsUnit]
         [Theory]
         [MemberData(nameof(AuthenticationTestData))]
-        public async Task GetTokenProvider(WebhookConfig config, IAuthenticationHandler expectedHandler)
+        public async Task AuthenticationHandler_Returns_RespectiveHandlerForConfigType(WebhookConfig config, IAuthenticationHandler expectedHandler)
```
tabman83, line 58: AuthenticationHandlerFactory_WhenInvoked_ReturnsCorrectType

src/CaptainHook.EventHandlerActor/Handlers/Authentication/AuthenticationHandlerFactory.cs
```diff
@@ -79,15 +79,20 @@ public async Task<IAuthenticationHandler> GetAsync(WebhookConfig config, Cancell
...
+                var key = newConfig.Uri;
+
                 await _semaphore.WaitAsync(cancellationToken);
 
-                if (_handlers.ContainsKey(key))
+                if (_handlers.TryGetValue(key, out var handler))
                 {
-                    return;
+                    if (handler.HasConfigChanged(newConfig.AuthenticationConfig))
+                        _handlers.TryRemove(key, out _);
```
chomick, line 93: I don't think you need `TryRemove` as you are certain the element exists and you are in critical section

src/CaptainHook.EventHandlerActor/Handlers/Authentication/OidcAuthenticationHandler.cs
tabman83: Reopening as I don't think you need the order by

tabman83 (changes requested):

src/Tests/CaptainHook.Tests/Web/Authentication/AuthenticationFactoryTests.cs
```diff
@@ -37,29 +63,154 @@ public class AuthenticationFactoryTests
...
+                var authConfig = webhookConfig.AuthenticationConfig as BasicAuthenticationConfig;
+                var handler = await factory.GetAsync(webhookConfig, CancellationToken.None);
+                var token = await handler.GetTokenAsync(CancellationToken.None);
+
+                expectedTokens.Add(
+                    $"Basic {BasicAuthenticationHandler.CreateBasicAuthToken(authConfig.Username, authConfig.Password)}");
+                receivedTokens.Add(token);
+            }
+
+            receivedTokens.Should().Equal(expectedTokens);
```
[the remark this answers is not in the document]
mszymeshop: What I meant is Fluent Assertions documentation. Checked now and Equals verifies the ordering, so closing here.

src/Tests/CaptainHook.Tests/Web/Authentication/AuthenticationFactoryTests.cs
```diff
@@ -1,25 +1,41 @@
...
 
 namespace CaptainHook.Tests.Web.Authentication
 {
     public class AuthenticationFactoryTests
     {
+        private readonly BigBrother _bigBrother;
+
+        public AuthenticationFactoryTests()
+        {
+            _bigBrother = Mock.Of<BigBrother>();
```
mszymeshop, line 30: I think you want to go with IBigBrother here instead (although this seems to work, we should be mocking the types your system under tests relies on, and this is IBigBrother here).

src/Tests/CaptainHook.Tests/Web/Authentication/AuthenticationFactoryTests.cs
```diff
@@ -37,29 +63,150 @@ public class AuthenticationFactoryTests
...
+            }
+
+            receivedTokens.Should().Equal(expectedTokens);
+        }
+
+        /// <summary>
+        /// Checks that the auth token changes with changes in OIDC auth parameters
+        /// </summary>
+        [Fact, IsUnit]
+        public async Task When_OidcAuthParamsUpdated_ExpectUpdatedTokenInRequests()
```
dawidkeshop, line 119: Maybe I'm missing something, but couldn't we just have a `Theory` here and use `GetOidcAuthChangeTestData` as `MemberData`?
I think it's easier to follow and fix if we have multiple separate test cases instead of one big test with couple of loops inside.

src/Tests/CaptainHook.Tests/Web/Authentication/AuthenticationFactoryTests.cs
```diff
@@ -37,29 +63,154 @@ public class AuthenticationFactoryTests
...
+
+            foreach (var kvPair in tokenWebhookConfigMap)
+            {
+                var webhookConfig = kvPair.Value;
+                var expectedAccessToken = kvPair.Key;
+
+                var handler = await factory.GetAsync(webhookConfig, cancellationToken);
+                var token = await handler.GetTokenAsync(cancellationToken);
+
+                expectedTokens.Add($"Bearer {expectedAccessToken}");
```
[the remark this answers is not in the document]
mszymeshop: I don't see this change for some reason.

src/CaptainHook.EventHandlerActor/Handlers/Authentication/OidcAuthenticationHandler.cs
nikhilgirraj: I added order by for situations where scopes in the two configurations are specified as 
config1 -> { "scope1", "scope2" }
config2 -> { "scope2", "scope1" }

Here config1 should match config2.

src/CaptainHook.EventHandlerActor/Handlers/Authentication/AuthenticationHandlerFactory.cs
nikhilgirraj: I'm happy to change this but seems like it would internally do the same. Or would it not?

src/CaptainHook.EventHandlerActor/Handlers/Authentication/OidcAuthenticationHandler.cs
chomick: It is not necessary but it will work properly - it just allows code to ignore change in the same scopes order - this is NULL change from auth point of view - so, code is correct but it also would be fine if we refreshed (unnecessarily) config for this case. (makes the code simpler).

src/CaptainHook.EventHandlerActor/Handlers/Authentication/AuthenticationHandlerFactory.cs
nikhilgirraj: I'm happy to change this but seems like it would internally do the same since it is an extension method. Or would it not?

chomick (approved):
Few outstanding questions from guys but overall looks good

src/Tests/CaptainHook.Tests/Web/Authentication/AuthenticationFactoryTests.cs
nikhilgirraj: Right, I'll change it.

src/Tests/CaptainHook.Tests/Web/Authentication/AuthenticationFactoryTests.cs
nikhilgirraj: The loop tests whether changes to AuthenticationConfig reflect in GetTokenAsync. Making it MemberData would reset the test context for every run which is not what I wanted?

Unless you meant to put various combinations of these into different tests. Did you?

src/CaptainHook.EventHandlerActor/Handlers/Authentication/OidcAuthenticationHandler.cs
nikhilgirraj: In that case, I can remove it.

src/CaptainHook.EventHandlerActor/Handlers/Authentication/AuthenticationHandlerFactory.cs
nikhilgirraj: I'm happy to change this but seems like it would internally do the same since a `.Remove` is an extension method. Or would it not?

nikhilgirraj:
As per @mszymeshop's suggestions, 
- I have changed the tests from "check-for-a-different-handler-for-a-different-auth-config" instead of "check-for-a-different-token-for-a-different-auth-config". 
- Removed the public static method in Basic Auth handler.

src/Tests/CaptainHook.Tests/Web/Authentication/AuthenticationFactoryTests.cs
```diff
@@ -37,29 +53,145 @@ public class AuthenticationFactoryTests
...
+
+            // Act
+            var handlers = new List<IAuthenticationHandler>();
+            foreach (var (_, webhookConfig) in tokenWebhookConfigMap)
+            {
+                handlers.Add(await factory.GetAsync(webhookConfig, cancellationToken));
+            }
+
+            // Assert 
+            handlers.Should().OnlyHaveUniqueItems();
```
nikhilgirraj, line 139: Checks for reference since `.Equals` is not overridden. Let me know if there's a better way.

src/Tests/CaptainHook.Tests/Web/Authentication/AuthenticationFactoryTests.cs
```diff
@@ -37,29 +50,145 @@ public class AuthenticationFactoryTests
...
+            // Arrange
+            const string uri = "http://localhost/api/v2/oidc";
+            var cancellationToken = new CancellationToken();
+
+            var tokenWebhookConfigMap = GetOidcAuthChangeTestData(uri);
+            var mockHttp = new MockHttpMessageHandler(BackendDefinitionBehavior.Always);
+
+            foreach (var (expectedAccessToken, webhookConfig) in tokenWebhookConfigMap)
+            {
+                SetupMockHttpResponse(mockHttp, webhookConfig.AuthenticationConfig as OidcAuthenticationConfig, expectedAccessToken);
```
mszymeshop, line 120: Now that you don't have the expected token being used, is this still required?

src/Tests/CaptainHook.Tests/Web/Authentication/AuthenticationFactoryTests.cs
mszymeshop: ok, as long as you have tested that.

src/Tests/CaptainHook.Tests/Web/Authentication/AuthenticationFactoryTests.cs
nikhilgirraj: Good point. Since now I'm only testing the handlers, this should be fine.

src/CaptainHook.EventHandlerActor/Handlers/Authentication/OidcAuthenticationHandler.cs
```diff
@@ -83,6 +84,30 @@ public virtual async Task<string> GetTokenAsync(CancellationToken cancellationTo
...
+        {
+            if (newConfig is OidcAuthenticationConfig oidcNewConfig)
+            {
+                var equals = oidcNewConfig.Uri == OidcAuthenticationConfig.Uri
+                               && oidcNewConfig.ClientId == OidcAuthenticationConfig.ClientId
+                               && oidcNewConfig.ClientSecret == OidcAuthenticationConfig.ClientSecret
+                               && oidcNewConfig.GrantType == OidcAuthenticationConfig.GrantType
+                               && oidcNewConfig.RefreshBeforeInSeconds == OidcAuthenticationConfig.RefreshBeforeInSeconds;
+                
+                if (oidcNewConfig.Scopes != null)
```
mszymeshop, line 97: I think you might have overcomplicated that. A use of `??` and `?.` would be sufficient here, unless I am not seeing something.

mszymeshop (approved):

src/CaptainHook.EventHandlerActor/Handlers/Authentication/OidcAuthenticationHandler.cs
nikhilgirraj: Just wanted to cover the situation where (that.Scopes?.SequenceEqual(other.Scopes) ?? false) is what I thought, but this would not cover the situation where both are null. Of course that's a hypothetical one.

src/Tests/CaptainHook.Tests/Web/Authentication/AuthenticationFactoryTests.cs
nikhilgirraj: Removed it. :-)

PR #211, opened 2020-08-21
#81037 Update auth mappings and validators

tabman83:


src/CaptainHook.Storage.Cosmos/CaptainHook.Storage.Cosmos.csproj
```diff
@@ -5,7 +5,7 @@
   </PropertyGroup>
 
   <ItemGroup>
-    <PackageReference Include="Eshopworld.Data.CosmosDb" Version="3.0.1" />
+    <PackageReference Include="Eshopworld.Data.CosmosDb" Version="3.0.3-preview3" />
```
dawidkeshop, line 8: Should this be a `preview` version?

dawidkeshop (approved):
Looks OK

src/CaptainHook.Application/Handlers/Subscribers/UpsertWebhookRequestHandler.cs
```diff
@@ -143,5 +142,15 @@ private static OperationResult<SubscriberEntity> MapRequestToSubscriber(UpsertWe
 
             return addWebhookResult?.Error ?? addWebhookResult;
         }
+
```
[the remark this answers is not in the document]
chomick: No reason to keep adding the wrong pattern here

src/CaptainHook.Contract/AuthenticationDtoJsonConverter.cs
```diff
@@ -0,0 +1,45 @@
...
+        public override bool CanConvert(Type objectType)
+        {
+            return AuthenticationDtoType.IsAssignableFrom(objectType);
+        }
+
+        public override object ReadJson(JsonReader reader, Type objectType, object existingValue, JsonSerializer serializer)
+        {
+            JObject jObject = JObject.Load(reader);
+
+            var typeDesc = jObject["type"]?.Value<string>();
```
chomick, line 23: Will it work as expected if JObject does not contain 'type'? Or will it throw exception?

src/CaptainHook.Contract/AuthenticationDtoJsonConverter.cs
```diff
@@ -0,0 +1,45 @@
+﻿using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+using System;
+
+namespace CaptainHook.Contract
+{
+    public class AuthenticationDtoJsonConverter : JsonConverter
```
chomick, line 7: Do you have unit tests for this?

src/CaptainHook.Storage.Cosmos/CaptainHook.Storage.Cosmos.csproj
tabman83: I have updated the CosmosDb package to support custom deserialization to enable polymorphism on the auth db entities. Since the PR is still outstanding, I have released a preview version of that package in order to unblock the work here. Once te PR for the CosmosDb is approved and merged, I will updated to non-preview version here.

src/CaptainHook.Application/Validators/PolymorphicValidator.cs
```diff
@@ -0,0 +1,64 @@
...
+using FluentValidation.Results;
+using FluentValidation.Validators;
+using System;
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace CaptainHook.Application.Validators
+{
+	public class PolymorphicValidator<T, TInterface> : ChildValidatorAdaptor<T, TInterface>
```
chomick, line 11: Do we need this wrapper class or can we use many ChildValidatorAdaptor on your EndpointDtoValidator?
I don't know the fluent validators but maybe they support it out of the box?

src/CaptainHook.Contract/BasicAuthenticationDto.cs
```diff
@@ -2,6 +2,7 @@
 {
     public class BasicAuthenticationDto : AuthenticationDto
     {
+        public const string Type = "Basic";
```
chomick, line 5: Are we not using field on base class?

src/CaptainHook.Storage.Cosmos/CaptainHook.Storage.Cosmos.csproj
dawidkeshop: OK, makes sense for me.

src/CaptainHook.Storage.Cosmos/Models/BasicAuthenticationSubdocument.cs
```diff
@@ -0,0 +1,28 @@
...
+    /// Basic authentication model in cosmos db
+    /// </summary>
+    internal class BasicAuthenticationSubdocument : AuthenticationSubdocument
+    {
+        public const string Type = "Basic";
+
+        /// <summary>
+        /// Authentication type
+        /// </summary>
+        [JsonProperty("type")]
```
chomick, line 15: Do you need to repeat the attribute here?

src/Tests/CaptainHook.Application.Tests/Validators/Dtos/AuthenticationDtoValidatorTests.cs
```diff
@@ -0,0 +1,140 @@
...
+        {
+            var dto = new OidcAuthenticationDtoBuilder().With(x => x.Uri, invalidString).Create();
+
+            var result = _oidcValidator.TestValidate(dto);
+
+            result.ShouldHaveValidationErrorFor(x => x.Uri);
+        }
+
+        [Theory, IsUnit]
+        [ClassData(typeof(InvalidUris))]
```
chomick, line 60: We should also validate that url with `{abc}` will also be considered invalid in auth setup

src/CaptainHook.Application/Handlers/Subscribers/UpsertWebhookRequestHandler.cs
```diff
@@ -143,5 +142,15 @@ private static OperationResult<SubscriberEntity> MapRequestToSubscriber(UpsertWe
...
+        private static AuthenticationEntity MapAuthentication(AuthenticationDto authenticationDto)
```
chomick, line 146: Do you have unit tests for the auth mappings? for any/all handlers?

src/CaptainHook.Storage.Cosmos/JsonCosmosSerializer.cs
```diff
@@ -0,0 +1,51 @@
+﻿using Microsoft.Azure.Cosmos;
+using Newtonsoft.Json;
+using System;
+using System.IO;
+using System.Text;
+
+namespace CaptainHook.Storage.Cosmos
+{
+    internal class JsonCosmosSerializer : CosmosSerializer
```
chomick, line 9: Do you have unit tests for that?

src/CaptainHook.Storage.Cosmos/AuthenticationSubdocumentJsonConverter.cs
```diff
@@ -0,0 +1,42 @@
+﻿using CaptainHook.Storage.Cosmos.Models;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+using System;
+
+namespace CaptainHook.Storage.Cosmos
+{
+    internal class AuthenticationSubdocumentJsonConverter : JsonConverter
```
chomick, line 8: Do you have unit tests for that?

src/CaptainHook.Contract/AuthenticationDtoJsonConverter.cs
```diff
@@ -0,0 +1,45 @@
+﻿using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+using System;
+
+namespace CaptainHook.Contract
```
chomick, line 5: I don't think serialiser belongs to the contract - contract is just POCOs; serialiser would be within API or Application namespace.

src/CaptainHook.Storage.Cosmos/JsonCosmosSerializer.cs
chomick, line 9: What's the reason for this class?

chomick (changes requested):
Thanks Nino, good job getting this implementation ready.

However, this PR is way too big. You should have broken it into smaller PRs, even if they are chained one after another, if you did not have anyone to review it.

There would be many ways to break it down, but at least would focus on getting individual layers PRed.

Couple of questions to be answered. 
Also, can you confirm if you tested it locally?

src/CaptainHook.Contract/AuthenticationDtoJsonConverter.cs
tabman83: It will be null and will fall to the default case on the switch (tested)

src/CaptainHook.Contract/AuthenticationDtoJsonConverter.cs
tabman83: I don't - do we want it? Also for the cosmos json converter?

src/CaptainHook.Storage.Cosmos/JsonCosmosSerializer.cs
tabman83: Nope!

tabman83: I don't because this is a copy of MS implementation here: https://github.com/Azure/azure-cosmos-dotnet-v3/blob/06b961a6c181995590097f764b296403169974f8/Microsoft.Azure.Cosmos/src/Serializer/CosmosJsonDotNetSerializer.cs 
which is internal

src/CaptainHook.Storage.Cosmos/AuthenticationSubdocumentJsonConverter.cs
tabman83: Nope

src/CaptainHook.Contract/AuthenticationDtoJsonConverter.cs
tabman83: I can move it to API

src/CaptainHook.Storage.Cosmos/JsonCosmosSerializer.cs
tabman83: I need that to specify serializer settings under which I have my custom AuthenticationSubdocumentJsonConverter

src/CaptainHook.Contract/AuthenticationDtoJsonConverter.cs
tabman83: If I move it to API or Application it would create a circular dependency as those 2 projects already depend on Contract

src/CaptainHook.Application/Validators/PolymorphicValidator.cs
tabman83: There's no support out of the box although the developer claims it is going to be added in v. 9.2
There's a discussion which started here https://github.com/FluentValidation/FluentValidation/issues/1237 where I borrowed this approach. ChildValidatorAdaptor is not something we can use directly.

src/CaptainHook.Contract/BasicAuthenticationDto.cs
tabman83: Not anymore, these I need to reuse when matching the auth type in the mappers and deserializer

src/CaptainHook.Storage.Cosmos/Models/BasicAuthenticationSubdocument.cs
tabman83: I can double check if the attribute is inferred from the base class

src/CaptainHook.Application.Infrastructure/Mappers/SubscriberEntityToConfigurationMapper.cs
```diff
@@ -99,20 +100,38 @@ private async Task<WebhookConfigRoute> MapEndpointToRouteAsync(EndpointEntity en
...
+        {
+            if (authenticationEntity?.ClientSecretKeyName == null)
                 return null;
 
-            var secretValue = await _secretProvider.GetSecretValueAsync(cosmosAuthentication.ClientSecretKeyName);
+            var secretValue = await _secretProvider.GetSecretValueAsync(authenticationEntity.ClientSecretKeyName);
 
             return new OidcAuthenticationConfig
             {
-                Type = AuthenticationType.OIDC,
```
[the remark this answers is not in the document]
tabman83: Yes, will change.

src/CaptainHook.Contract/AuthenticationDtoJsonConverter.cs
chomick: Yes, we should - otherwise, how do you know it works :)
These should be quick to add, cases I would test:
- username/password type
- oidc type
- invalid type

src/CaptainHook.Storage.Cosmos/JsonCosmosSerializer.cs
chomick: Doesn't matter if internal or not :) It is code that should be tested noramlly. However, I would question why do we need that?

src/CaptainHook.Contract/AuthenticationDtoJsonConverter.cs
chomick: Contract should not depend on anything.
API/App should depend on contract.

What is it in contract that we need from API/App?

src/CaptainHook.Storage.Cosmos/JsonCosmosSerializer.cs
chomick: Can you use RequestOptions?
https://docs.microsoft.com/en-us/dotnet/api/microsoft.azure.documents.client.requestoptions?view=azure-dotnet

You would have to extend the cosmos db package again to allow for serialisation settings to be specified in the repository (you could expose it via repo protected property so it can be redefined in subclasses)

src/CaptainHook.Application.Infrastructure/Mappers/SubscriberEntityToConfigurationMapper.cs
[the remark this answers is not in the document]
tabman83: Yes it is!

tabman83:
> Thanks Nino, good job getting this implementation ready.
> 
> However, this PR is way too big. You should have broken it into smaller PRs, even if they are chained one after another, if you did not have anyone to review it.
> 
> There would be many ways to break it down, but at least would focus on getting individual layers PRed.
> 
> Couple of questions to be answered.
> Also, can you confirm if you tested it locally?

Yes I tested the API locally, request is persisted to the database and on subsequent restart the new subscriber is correctly pulled from db.

src/CaptainHook.Application/Handlers/Subscribers/UpsertSubscriberRequestHandler.cs
```diff
@@ -56,9 +56,7 @@ private static SubscriberEntity MapRequestToEntity(UpsertSubscriberRequest reque
         {
             var webhooks = new WebhooksEntity(
                 request.Subscriber.Webhooks.SelectionRule,
-                request.Subscriber.Webhooks.Endpoints?.Select(MapEndpointEntity) ?? Enumerable.Empty<EndpointEntity>(),
-                MapUriTransformEntity(request.Subscriber.Webhooks.UriTransform));
```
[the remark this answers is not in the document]
tabman83: It may have slipped. Re-added now.

src/CaptainHook.Contract/AuthenticationDtoJsonConverter.cs
tabman83: Added tests for this

src/CaptainHook.Storage.Cosmos/AuthenticationSubdocumentJsonConverter.cs
tabman83: Added tests for this

src/CaptainHook.Storage.Cosmos/Models/BasicAuthenticationSubdocument.cs
tabman83: Unit tests highlighted that this can be safely removed. Done.

src/CaptainHook.Contract/AuthenticationDtoJsonConverter.cs
tabman83: The AuthenticationDtoJsonConverter is specified as an attribute on the AuthenticationDto, so if I move the AuthenticationDtoJsonConverter to a different project it would require a dependency on that project.

src/CaptainHook.Storage.Cosmos/JsonCosmosSerializer.cs
tabman83: That is defined in the Azure DocumentDB library, while our Cosmos package has been already upgraded to Azure Cosmos library.

src/CaptainHook.Application/Handlers/Subscribers/UpsertSubscriberRequestHandler.cs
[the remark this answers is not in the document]
chomick: There should be unit test that caught it

src/CaptainHook.Contract/AuthenticationDtoJsonConverter.cs
chomick: Is there any other way to specify the converter than via attribute? It feels wrong that our contract needs to have the converter defined in the same project. 
I like to think of the contract almost as a auto-generated code based on some abstract contract definition :)

src/CaptainHook.Contract/BasicAuthenticationDto.cs
chomick: But the property is still part of the class - should it not be set (in ctor) to the right value?

src/CaptainHook.Contract/AuthenticationDtoJsonConverter.cs
tabman83: The AuthenticationDtoJsonConverter has been moved to the API project

src/Tests/CaptainHook.Application.Tests/Validators/Dtos/AuthenticationDtoValidatorTests.cs
tabman83: Added test for that

src/CaptainHook.Application/Handlers/Subscribers/UpsertWebhookRequestHandler.cs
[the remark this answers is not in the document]
tabman83: Will be handled in https://dev.azure.com/eshopworld/evo-core/_workitems/edit/81792

src/CaptainHook.Application/Handlers/Subscribers/UpsertWebhookRequestHandler.cs
tabman83: Will be handled in https://dev.azure.com/eshopworld/evo-core/_workitems/edit/81792

src/CaptainHook.Contract/BasicAuthenticationDto.cs
tabman83: Right, changed that.

src/CaptainHook.Storage.Cosmos/JsonCosmosSerializer.cs
chomick: As discussed, Nino will try to serialise directly in the repository using the custom converter rather than substitute the serializer for Cosmos SDK.

chomick: As discussed, Nino will try to serialise directly in the repository using the custom converter rather than substitute the serializer for Cosmos SDK.

chomick (commented):
Looks good with the exception for the serializer - once you manage to replace that code with repository serialisation, we should be good.

src/CaptainHook.Storage.Cosmos/SubscriberRepository.cs
```diff
@@ -185,7 +185,10 @@ private async Task<OperationResult<IEnumerable<SubscriberEntity>>> GetSubscriber
 
         private static SubscriberDocument Deserialize(dynamic document)
         {
-            return JsonConvert.DeserializeObject(document?.ToString(), SubscriberDocumentType, AuthenticationSubdocumentJsonConverter);
+            return (SubscriberDocument)JsonConvert.DeserializeObject(
```
chomick, line 188: I think there should be generic version of that method as well?

src/CaptainHook.Storage.Cosmos/SubscriberRepository.cs
```diff
@@ -176,6 +183,14 @@ private async Task<OperationResult<IEnumerable<SubscriberEntity>>> GetSubscriber
             }
         }
 
+        private static SubscriberDocument Deserialize(dynamic document)
+        {
+            return (SubscriberDocument)JsonConvert.DeserializeObject(
```
chomick, line 188: I think I made a comment on particular commit - I think there should be generic version of that method, no need for Type to be used then.

src/CaptainHook.Storage.Cosmos/Models/WebhookSubdocument.cs
```diff
@@ -10,19 +10,17 @@ internal class WebhookSubdocument
...
         /// <summary>
         /// Webhook selector
         /// </summary>
-        [JsonProperty("selectionRule")]
         public string SelectionRule { get; set; }
 
         /// <summary>
         /// Defines Uri transformation definition
         /// </summary>
-        [JsonProperty("uriTransform", NullValueHandling = NullValueHandling.Ignore)]
```
chomick, line 25: Any reason we are changing the name of the field in DB?

src/CaptainHook.Storage.Cosmos/Models/OidcAuthenticationSubdocument.cs
```diff
@@ -3,38 +3,35 @@
...
+
+        /// <summary>
+        /// Authentication type
+        /// </summary>
+        public override string AuthenticationType => Type;
+
         /// <summary>
         /// Client id for authentication
         /// </summary>
-        [JsonProperty("clientId")]
```
chomick, line 13: I actually likes the fact that the contract is well defined here. Right now, we will rely on the fact that some options are set somewhere and if they change, the DB doc contract will also change.
No action needed, just something to consider in future.

chomick (approved):
We MUST avoid such massive PRs - very difficult to read/reason about it - I had difficulty reviewing it and possibly I missed things

src/CaptainHook.Storage.Cosmos/SubscriberRepository.cs
tabman83: Changed

src/CaptainHook.Storage.Cosmos/SubscriberRepository.cs
tabman83: Changed

PR #212, opened 2020-08-24
#81052 Remove payload logging for SAND and PROD envs

tabman83:
- SandBox should not have message payload logging to Application Insights contained within FailureWebhookEvents
- Production should not have message payload logging to Application Insights contained within FailureWebhookEvents
- Additionally, the FailureWebHook event has been renamed to FailureWebhookEvent to make it consistent with the WebhookEvent (indicating success)


src/CaptainHook.EventHandlerActor/Handlers/RequestLogger.cs
```diff
@@ -36,6 +37,7 @@ WebhookConfig config
         )
         {
             var webhookRules = CollectRules (config);
+            var canLogPayload = (EswDevOpsSdk.GetEnvironment() != DeploymentEnvironment.Prod && EswDevOpsSdk.GetEnvironment() != DeploymentEnvironment.Sand);
```
chomick, line 40: I would prefer to have a feature flag rather than hardcoded envs.

src/CaptainHook.EventHandlerActor/Handlers/RequestLogger.cs
```diff
@@ -56,9 +58,9 @@ WebhookConfig config
                 _bigBrother.Publish(new FailedWebHookEvent(
```
mszymeshop, line 58: Could you add a unit test that verifies this? Somebody could change that in the future by mistake and we could end up logging these by mistake.

mszymeshop (changes requested):
The rename of `FailedWebHookEvent` which you have mentioned in the description is not visible in the PR.

src/CaptainHook.Common/Telemetry/FailedWebHookEvent.cs
```diff
@@ -37,13 +37,14 @@ public FailedWebHookEvent()
 
         public string ResponseHeaders { get; set; }
 
+        // sent for all environments except for SAND and PROD
```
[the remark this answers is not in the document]
tabman83: Done!

src/CaptainHook.EventHandlerActor/Handlers/RequestLogger.cs
tabman83: Done!

src/CaptainHook.EventHandlerActor/Handlers/RequestLogger.cs
tabman83: This is now controlled by feature flags

src/CaptainHook.EventHandlerActor/Handlers/RequestLogger.cs
tabman83: Done

tabman83:
> The rename of `FailedWebHookEvent` which you have mentioned in the description is not visible in the PR.

Not sure why it did not get through in the first place, I have re-added now.

src/CaptainHook.Common/Configuration/TempConfigLoader.cs
```diff
@@ -33,6 +33,9 @@ public static IConfigurationRoot Load ()
                 AddEntry (client, "CaptainHook--DbConfiguration--DatabaseKey"),
                 AddEntry (client, "CaptainHook--DbConfiguration--Databases--EDA--1--CollectionName"),
                 AddEntry (client, "CaptainHook--DbConfiguration--Databases--EDA--1--PartitionKey"),
+
+                AddEntry (client, "CaptainHook--FeatureFlags--DisablePayloadLoggingForPROD"),
```
mszymeshop, line 37: Is there a reason you use 2 flags per environment? Instead of having a single one which just controls the payload? It would then be set to false for CI-PREP and true SAND-PROD?

tabman83: I have re-used a FeatureFlagsConfiguration class which is already there and it uses boolean flags

src/CaptainHook.Common/Telemetry/FailedWebHookEvent.cs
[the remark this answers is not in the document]
tabman83: Done

src/CaptainHook.EventHandlerActor/Handlers/RequestLogger.cs
tabman83: Using feature flags now

src/CaptainHook.Common/Configuration/TempConfigLoader.cs
mszymeshop: That is not what I am saying :) Why do you ask for SAND and PROD for each environment? Why should CI care about these entries?

dawidkeshop: Couldn't we just have a single flag like `DisablePayloadLogging` which will be `false` in CI, TEST and PREP and `true` on SAND and PROD?
It looks a bit confusing to have flags with `SAND` and `PROD` in the name on every environment.

src/CaptainHook.EventHandlerActor/Handlers/RequestLogger.cs
```diff
@@ -73,13 +80,20 @@ WebhookConfig config
             }
         }
 
+        private bool IsPayloadDisabled()
+        {
+            return
```
dawidkeshop, line 85: Do we have to check the environment type if we have a feature flag?

src/CaptainHook.Common/Configuration/TempConfigLoader.cs
tabman83: Hey guys you're absolutely, right I just made a bit of confusion, let me sort this out!

src/CaptainHook.EventHandlerActor/Handlers/RequestLogger.cs
tabman83: I have changed that

src/CaptainHook.Common/Configuration/TempConfigLoader.cs
tabman83: I have changed that

chomick: Still seeing old code?

tabman83: Changed

chomick (approved):

src/CaptainHook.EventHandlerActor/Handlers/RequestLogger.cs
```diff
@@ -58,7 +58,8 @@ WebhookConfig config
             }
             else
             {
-                var canLogPayload = !IsPayloadDisabled();
+                var canLogPayload = !(_configurationSettings.FeatureFlags.GetFlag<DisablePayloadLoggingFeatureFlag>()?.IsEnabled ?? false);
```
dawidkeshop, line 61: Hmm, so the flag name is `DisablePayloadLogging`, so if `IsEnabled` is `true`, it means that feature is actually disabled, right?
Bit confusing as for me, maybe try to rename it just to `PayloadLogging` and don't have this `Disable`-`IsEnabled` construction in a single line of code?

dawidkeshop (approved):
OK, but I would do something with the name as I suggested in the comment.

src/CaptainHook.EventHandlerActor/Handlers/RequestLogger.cs
tabman83: I know it gets confusing to read this, but we are taking from the assumption that all feature flags are disabled by default, and we don't want to disable payload logging in case the deploy goes ahead without the ff in.

PR #213, opened 2020-08-24
Reset connection retry fix

dawidkeshop:


src/CaptainHook.EventReaderService/EventReaderService.cs
```diff
@@ -260,20 +261,16 @@ protected override async Task RunAsync(CancellationToken cancellationToken)
...
-                    catch (ServiceBusException serviceBusException)
+                    catch (Exception exception) when (exception is ServiceBusException || exception is SocketException)
                     {
-                        _bigBrother.Publish(serviceBusException.ToExceptionEvent());
-
-                        await Policy
-                            .Handle<ServiceBusException>()
-                            .WaitAndRetryForeverAsync(_exponentialBackoff, OnServiceBusExceptionRetryAsync)
-                            .ExecuteAsync(ct => SetupServiceBusAsync(ct), cancellationToken);
+                        await ResetConnection(exception, cancellationToken);
```
tabman83, line 273: -Async suffix is missing

tabman83 (commented):
When I worked on that piece of code I was under the impression that the SocketException always was an inner exception of the containing one, ServiceBusException. My understanding was that ServiceBusException was the one always being thrown under that condition.

Do you then have any AI log substantiating your approach?

chomick (approved):

src/Tests/CaptainHook.Tests/Services/Reliable/EventReaderTests.cs
```diff
@@ -293,62 +209,25 @@ public async Task CanAddMoreHandlersDynamically(string eventName, int messageCou
...
+            var mockServiceBusManager = CreateMockServiceBusManager(eventName, messageCount);
             var cancellationTokenSource = new CancellationTokenSource(TimeSpan.FromSeconds(1));
+            var service = CreateService(mockServiceBusManager);
+
             await service.InvokeOnOpenAsync(ReplicaOpenMode.New, cancellationTokenSource.Token);
             await service.InvokeRunAsync(cancellationTokenSource.Token);
 
-            actor.MessageDataInstances.Select(m => m.SubscriberConfig).Should().AllBeEquivalentTo(subscriberConfiguration);
+            var subscriberConfigurations = actor.MessageDataInstances.Select(m => m.SubscriberConfig).ToList();
+            subscriberConfigurations.Should().HaveCountGreaterOrEqualTo(1);
```
mszymeshop, line 229: Is it possible to combine it with `And` so that you don't have two individual assertions?

dawidkeshop: Good idea!

PR #214, opened 2020-08-25
Event reader service tests fix

dawidkeshop:
EventReaderServiceTests contains a lot of code duplicated in every method.
Also one of the tests (`InitSubscriberDataIsPassedToHandlers`) was false positive because of invalid assertion logic.

Such assertion: 
```
collection.Should().AllBeEquivalentTo(new SubscriberConfiguration { EventType = eventName, SubscriberName = "subA" });
```
won't fail if `collection` is empty. It might seem counter-intuitive at the first glance but it's logical. 
So if we'd like to perform `AllBeEquivalentTo` assertion we should first check if collection is not empty.


mszymeshop (approved):
It's high time somebody refactored that. Great work.

chomick (approved):

PR #215, opened 2020-08-25
Story 73228 update entire subscriber

nikhilgirraj:
**Changes**

- Modifies existings API  (PUT .../subscriber/{name}) to allow updating the subscriber when it exists.  
- Removed logic to no longer reject requests when passing existing subscriber.


nikhilgirraj:
Some tests fail. Please hold off reviews until after I fix them.

nikhilgirraj:
Ready for review!

src/CaptainHook.Api/Controllers/EventsController.cs
```diff
@@ -126,18 +127,22 @@ public async Task<IActionResult> PutSuscriber([FromRoute] string eventName, [Fro
...
+                        ValidationError validationError => BadRequest(validationError),
+                        DirectorServiceIsBusyError directorServiceIsBusyError => Conflict(directorServiceIsBusyError),
+                        ReaderCreateError readerCreationError => UnprocessableEntity(readerCreationError),
+                        CannotSaveEntityError cannotSaveEntityError => UnprocessableEntity(cannotSaveEntityError),
+                        _ => StatusCode(StatusCodes.Status500InternalServerError, error)
+                    },
+                    upsertResult => upsertResult.UpsertType switch
+                    {
+                        UpsertType.Created => Created($"/{eventName}/subscriber/{subscriberName}", upsertResult.Dto),
+                        UpsertType.Updated => Accepted($"/{eventName}/subscriber/{subscriberName}", upsertResult.Dto),
```
nikhilgirraj, line 141: @chomick I did not change this Accepted to Ok since the current behavior requires both the route and response object. Cannot specify the resource URL with a 200Ok response.

mszymeshop: @nikhilgirraj I don't think you need the resource URL at all for Ok, so to me you can go ahead with Ok

src/CaptainHook.Application/Results/UpsertResult.cs
```diff
@@ -0,0 +1,20 @@
+﻿namespace CaptainHook.Application.Results
+{
+    public class UpsertResult<T>
```
mszymeshop, line 3: Should we consider extending `OperationResult` instead so that it also carried additional information on the data - created, deleted, updated etc? @dawidkeshop what do you reckon?

dawidkeshop (approved):

src/CaptainHook.Application/Results/UpsertResult.cs
dawidkeshop: We need this distinction here, because we have a single endpoint to do insert/update. Delete is a separate endpoint, so we don't need a special class to represent the result of the delete operation. 
As for me what Nikhil implemented is good.  `OperationResult` has been designed as a wrapper for any data, so if we need something to be wrapped we can use this class directly. I don't see any profit of introducing inheritance over composition, but I see disadvantages like lacking the consistency of method return types which might cause some difficulties when we decide to move some code (like conversion to HTTP result) to a single place.

src/CaptainHook.Application/Handlers/Subscribers/UpsertSubscriberRequestHandler.cs
```diff
@@ -25,31 +26,63 @@ public UpsertSubscriberRequestHandler(ISubscriberRepository subscriberRepository
...
         {
             var subscriberId = new SubscriberId(request.EventName, request.SubscriberName);
             var existingItem = await _subscriberRepository.GetSubscriberAsync(subscriberId);
 
-            if (!(existingItem.Error is EntityNotFoundError))
+            var subscriber = MapRequestToEntity(request);
+            OperationResult<bool> saveResult;
+            UpsertType upsertType;
+
+            if (existingItem.IsError)
```
chomick, line 38: A small note: it would be cleaner if your code would return an error first and you would not need so much nesting, i.e.

```
if (is error and not entity-not-found)
  return error;

if (entity-not-found) {
  ...
}
else {
  ...
}

```

src/Tests/CaptainHook.Application.Tests/Handlers/Subscribers/UpsertSubscriberRequestHandlerTests.cs
```diff
@@ -54,12 +55,12 @@ public async Task When_SubscriberDoesNotExist_Then_NewOneWillBePassedToDirectorS
...
             var result = await _handler.Handle(request, CancellationToken.None);
 
-            var expectedResult = new OperationResult<SubscriberDto>(subscriberDto);
+            var expectedResult = new OperationResult<UpsertResult<SubscriberDto>>(new UpsertResult<SubscriberDto>(subscriberDto, UpsertType.Created));
             result.Should().BeEquivalentTo(expectedResult);
         }
 
         [Fact, IsUnit]
-        public async Task When_Subscriber_DoesExist_Then_OperationFails()
+        public async Task When_Subscriber_DoesExist_Then_SubscriberIsUpdated()
```
chomick, line 63: When_Subscriber_DoesExist_Then_SubscriberIsCreated

chomick (approved):

src/CaptainHook.Api/Controllers/EventsController.cs
nikhilgirraj: I am 50-50 on this one. On one hand I believe the functionality on the receiving side should be consistent and on the other hand, the Ok thing makes sense as well. 
If you want I can change this to `Ok`.

dawidkeshop: I think the idea was to return the whole subscriber object after every operation (including delete and operations on single webhook). With this approach we probably don't need additional location anywhere.

PR #216, opened 2020-08-25
#80115 - provide endpoint selector identifier in the uri, use * as default

mszymeshop:
This PR focuses on two main aspects:

1. Specify endpoint selector in the uri
2. Use * as the default selector (instead of null previously)

### What

The API endpoints are not duplicated now as both PUT and DELETE can handle selectors:
image

The information on * is also provided is Swagger UI for visual support:
image

### DIFF
--- a/src/CaptainHook.Application/Validators/Dtos/WebhooksDtoValidator.cs
+++ b/src/CaptainHook.Application/Validators/Dtos/WebhooksDtoValidator.cs
@@ -1,9 +1,11 @@
-﻿using CaptainHook.Application.Validators.Common;
+﻿using System;
 using CaptainHook.Contract;
 using FluentValidation;
 using Newtonsoft.Json.Linq;
 using System.Collections.Generic;
 using System.Linq;
+using CaptainHook.Domain.Entities;
+using UriTransformValidator = CaptainHook.Application.Validators.Common.UriTransformValidator;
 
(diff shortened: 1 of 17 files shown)


src/CaptainHook.Application/Validators/Dtos/WebhooksDtoValidator.cs
```diff
@@ -17,48 +19,47 @@ public WebhooksDtoValidator()
 
             RuleFor(x => x.SelectionRule).NotEmpty()
                 .Must(BeValidJsonPathExpression).WithMessage("The SelectionRule must be a valid JSONPath expression")
-                .When(TheresAtLeastOneEndpointWithSelectorDefined);
+                .When(ThereIsAtLeastOneEndpointWithSelectorDefined);
 
             RuleFor(x => x.Endpoints)
                 .NotNull()
                 .WithMessage("Webhooks list must contain at list one endpoint")
```
chomick, line 26: Typo (not this PR):
"at list" => "at least"

chomick (approved):

src/CaptainHook.Application/Validators/Dtos/WebhooksDtoValidator.cs
mszymeshop: hahahaha, good one 🥇

PR #218, opened 2020-08-25
#81057 - enable endpoint update in subscriber

mszymeshop:
### What
This PR will enable to Upsert Endpoints if they already exist on the subscriber.

### Disclaimer
I have identified a case where when you try to upsert endpoint with the same config as already in place, your SF Reader is going to disappear. I am going to address this in a separate PR.


src/CaptainHook.Domain/Entities/Comparers/SelectorEndpointEntityEqualityComparer.cs
```diff
@@ -0,0 +1,22 @@
+﻿using System;
+using System.Collections.Generic;
+
+namespace CaptainHook.Domain.Entities.Comparers
+{
+    public sealed class SelectorEndpointEntityEqualityComparer: IEqualityComparer<EndpointEntity>
```
mszymeshop, line 6: I am going with EqualityComparer as it is very specific (compares on Selector only), cannot be treated as a default compared (hence individual class) and allows me to reuse the same comparison logic where I need it.

src/CaptainHook.Domain/Entities/SubscriberEntity.cs
```diff
@@ -51,24 +53,29 @@ public void SetParentEvent(EventEntity parentEvent)
...
 
         /// <summary>
-        /// Adds an endpoint to the list of webhook endpoints
+        /// Adds an endpoint to the list of webhook endpoints if it is not on the list already.
+        /// Removes the existing endpoint and adds the new one to the list if the item is already present.
         /// </summary>
-        /// <param name="endpointModel"></param>
-        public OperationResult<SubscriberEntity> AddWebhookEndpoint(EndpointEntity endpointModel)
+        /// <remarks>The identification is made on selector using case-insensitive comparison.</remarks>
+        public OperationResult<SubscriberEntity> UpsertWebhookEndpoint(EndpointEntity endpointModel)
```
chomick, line 60: On the entity level, I would expect to have explicit operations to:
a) find an endpoint based on selector
b) add new endpoint (with validation that selector is correct)
c) update existing endpoint based on selector.

Those operations would be used by the handler then to perform correct action.

If you prefer to keep this high-level method here, I suggest we rename it to be "Set..." as upsert has strong DB connotation

src/CaptainHook.Domain/Entities/SubscriberEntity.cs
```diff
@@ -51,24 +53,29 @@ public void SetParentEvent(EventEntity parentEvent)
...
         {
             if (Webhooks == null)
             {
                 Webhooks = new WebhooksEntity();
             }
 
-            var extendedEndpointsCollection = Webhooks.Endpoints.Concat(new[] { endpointModel });
+            var extendedEndpointsCollection = Webhooks.Endpoints
```
chomick, line 67: This logic does not belong in this entity - it belongs in WebhooksEntity - this is the guy that manages endpoints, not subscriber?

src/CaptainHook.Domain/Entities/SubscriberEntity.cs
mszymeshop: @chomick I was considering the same thing but came to a conclusion that since we do not need those granular operations (I believe they would rather complicate the implementation + testing) then I shouldn't be following that path. Ultimately what we care about to keep changing that list and it should remain coherent. If the need arises, then absolutely this is something we can circle back to.

Not a fan of 'Upsert', 'Set' isn't ideal but probably better. 'AddOrUpdate' might be an overkill :/

src/CaptainHook.Domain/Entities/SubscriberEntity.cs
mszymeshop: The reason I have it here is that I validate the entire WebhooksEntity. I didn't want to change the contents of the existing object. This is something I would like to discuss though. Will reach out on Slack.

mszymeshop: If we don't have time to discuss I would like to go ahead with the current approach. I have taken another look and the change would also affect the other methods in this class. Probably better to handle that in the separate PR.

chomick: Sure, let's go ahead but we may want to review options for further updates

src/CaptainHook.Domain/Entities/SubscriberEntity.cs
chomick: I vote for "Set" if it matters :)

chomick (approved):
Nothing blocking the PR

src/CaptainHook.Domain/Entities/SubscriberEntity.cs
```diff
@@ -51,24 +53,29 @@ public void SetParentEvent(EventEntity parentEvent)
...
+                .Concat(new[] { endpointModel });
             var endpointsEntityForValidation = new WebhooksEntity(Webhooks.SelectionRule, extendedEndpointsCollection, Webhooks.UriTransform);
 
             var validationResult = WebhooksValidator.Validate(endpointsEntityForValidation);
 
             if (validationResult.IsValid)
             {
                 endpointModel.SetParentSubscriber(this);
+                
+                Webhooks.RemoveEndpoint(endpointModel.Selector);
```
tabman83, line 78: We now have a comparer, so why don't we pass the full endpoint model, so we are consistent with 'AddEnpoint' signature

tabman83 (approved):

dawidkeshop (approved):

src/CaptainHook.Domain/Entities/SubscriberEntity.cs
mszymeshop: I am going to work around that area, so will go ahead with how it is at the moment. There will be a separate PR.

src/CaptainHook.Domain/Entities/SubscriberEntity.cs
```diff
@@ -51,24 +53,29 @@ public void SetParentEvent(EventEntity parentEvent)
...
+        public OperationResult<SubscriberEntity> SetWebhookEndpoint(EndpointEntity endpointModel)
         {
             if (Webhooks == null)
             {
                 Webhooks = new WebhooksEntity();
             }
 
-            var extendedEndpointsCollection = Webhooks.Endpoints.Concat(new[] { endpointModel });
+            var extendedEndpointsCollection = Webhooks.Endpoints
+                .Except(new [] { endpointModel }, new SelectorEndpointEntityEqualityComparer())
```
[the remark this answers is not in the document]
mszymeshop: ok, there will be another PR

src/CaptainHook.Domain/Entities/SubscriberEntity.cs
mszymeshop: I'll prepare a suggestion as a separate PR

PR #219, opened 2020-08-26
#82054 - detect reader update no change needed

mszymeshop:
### What
The problem would occur where you would do PUT Webhook with the same details as currently in DB. In that case current-reader and expected-reader would be the same and for an update operation this wouldn't be a supported scenario leading to unwanted behaviour.

In this PR I am only focusing on the Director Service - assuming that updates that come in via API with the existing value are still allowed.


dawidkeshop (approved):

src/CaptainHook.DirectorService/DirectorService.cs
```diff
@@ -221,7 +221,15 @@ public async Task<ReaderChangeResult> ApplyReaderChange(ReaderChangeBase readerC
...
                                     if (!existingReader.IsValid)
+                                    {
                                         return ReaderChangeResult.ReaderDoesNotExist;
+                                    }
+
+                                    var noChanges = desiredReader.IsUnchanged(existingReader);
+                                    if (noChanges)
+                                    {
+                                        return ReaderChangeResult.NoChangeNeeded;
+                                    }
```
tabman83, line 233: We could use pattern matching here to improve readability

tabman83 (approved):

PR #221, opened 2020-08-26
Fix FeatureFlags loading logic

tabman83:
Although the code looked correct before, Feature Flags were not being loaded. This PR addresses the issue and has been tested locally.


src/CaptainHook.Common/Configuration/TempConfigLoader.cs
```diff
@@ -34,7 +34,7 @@ public static IConfigurationRoot Load ()
                 AddEntry (client, "CaptainHook--DbConfiguration--Databases--EDA--1--CollectionName"),
                 AddEntry (client, "CaptainHook--DbConfiguration--Databases--EDA--1--PartitionKey"),
 
-                AddEntry (client, "CaptainHook--FeatureFlags--DisablePayloadLogging"),
+                AddEntry (client, "CaptainHook--FeatureFlags--1"),
```
mszymeshop, line 37: Why have you stopped using the name which gives a bit of context? Just 1 suggest the order matters :/

dawidkeshop (approved):

mszymeshop (approved):

src/CaptainHook.Common/Configuration/TempConfigLoader.cs
tabman83: As discussed separately, this is needed by the binder to work. Will highlight this in a tech debt task.

src/CaptainHook.Common/Configuration/TempConfigLoader.cs
```diff
@@ -62,8 +62,15 @@ private static SecretClient GetClient ()
...
-            var value = client.GetSecretAsync (key).ConfigureAwait (false).GetAwaiter ().GetResult ();
-            return value?.Value?.Value;
+            try
+            {
+                var value = client.GetSecretAsync(key).ConfigureAwait(false).GetAwaiter().GetResult();
+                return value?.Value?.Value;
+            }
+            catch
+            {
+                return null;
```
chomick, line 72: I think we should log it - it may be really difficult to find issues that result in missing config entries

tabman83: Added logging. BigBrother is not injected since TempLoader is static and we will be moving away from that.

PR #224, opened 2020-08-27
#82281 Fix deserializers

tabman83:


mszymeshop:
@tabman83 the root cause of the issue is not clear to me, could you expand on this?

tabman83:
> @tabman83 the root cause of the issue is not clear to me, could you expand on this?

The Cosmos deserializer failed when the authentication object is null. The db has subscribers with null auth object from int tests. I addressed this and while there I applied the same fix to the DTO deserializer.

mszymeshop (approved):

mszymeshop:
> I addressed this and while there I applied the same fix to the DTO deserializer.
If I could suggest something, maybe next time worth having 2 PRs. One which directly addresses a pressing issue like a build break, and another one with improvements. Otherwise, they get glued together and it is more difficult to reason about the change that was actually necessary.

chomick (approved):

PR #225, opened 2020-08-27
Task 81057 refactor endpoints

mszymeshop:
### What
This is a suggestion to move the entity operations and validation to WebhooksEntity instead of SubscriberEntity. Once WebhooksEntity is actually renamed and reused for Callback and Dlq, we will be able to reuse that logic this way.


dawidkeshop (approved):

src/CaptainHook.Domain/Entities/SubscriberEntity.cs
```diff
@@ -57,41 +49,35 @@ public void SetParentEvent(EventEntity parentEvent)
...
+            => OperationOnWebhooks(webhooks => webhooks.SetEndpoint(entity.SetParentSubscriber(this)));
+
+        /// <summary>
+        /// Removes the existing endpoint from the list if the item is present.
+        /// </summary>
+        /// <remarks>The identification is made on selector using case-insensitive comparison.</remarks>
+        public OperationResult<SubscriberEntity> RemoveWebhookEndpoint(EndpointEntity entity)
+            => OperationOnWebhooks(webhooks => webhooks.RemoveEndpoint(entity));
+
+        private OperationResult<SubscriberEntity> OperationOnWebhooks(Func<WebhooksEntity, OperationResult<WebhooksEntity>> funcToRun)
```
chomick, line 62: Not sure I understand why we need that method? Could you simply call Webhooks.Xxx () instead of the Func<>?

src/CaptainHook.Domain/Entities/SubscriberEntity.cs
```diff
@@ -57,41 +49,35 @@ public void SetParentEvent(EventEntity parentEvent)
...
-            if (validationResult.IsValid)
+            if (webhooksResult.IsError)
             {
-                endpointModel.SetParentSubscriber(this);
-                
-                Webhooks.RemoveEndpoint(endpointModel.Selector);
-                Webhooks.AddEndpoint(endpointModel);
-
-                return this;
+                return webhooksResult.Error switch
```
chomick, line 73: Could these be simply returned from the webhookEnitty?

src/CaptainHook.Domain/Entities/WebhooksEntity.cs
```diff
@@ -42,15 +51,49 @@ public WebhooksEntity(string selectionRule, IEnumerable<EndpointEntity> endpoint
...
+            _endpoints.Add(endpoint);
+            return this;
         }
 
-        public bool RemoveEndpoint(string selector)
+        public OperationResult<WebhooksEntity> RemoveEndpoint(EndpointEntity endpoint)
         {
-            var toDelete = _endpoints.SingleOrDefault(x => string.Equals(x.Selector, selector, StringComparison.InvariantCultureIgnoreCase));
-            return toDelete != null && _endpoints.Remove(toDelete);
+            if (_endpoints.Count == 1)
```
chomick, line 80: This may become optional check as for DLQ/Callbacks, it will be possible. We could initialise the entity with the flags/options whether this is allowed or not (not this PR)

src/CaptainHook.Domain/Entities/WebhooksEntity.cs
```diff
@@ -42,15 +51,49 @@ public WebhooksEntity(string selectionRule, IEnumerable<EndpointEntity> endpoint
...
+            {
+                return new EndpointNotFoundInSubscriberError(endpoint.Selector);
+            }
+
+            _endpoints.Remove(toDelete);
+
+            return this;
         }
+
+        private EndpointEntity FindMatchingEndpoint(EndpointEntity endpoint)
```
chomick, line 96: Can you search by 'selector'? No need for equality comparer then or, if you must, hide that detail in your private method rather than in the contract

src/CaptainHook.Domain/Errors/CannotRemoveLastEndpointFromSusbscriberError.cs
```diff
@@ -3,10 +3,15 @@
 
 namespace CaptainHook.Domain.Errors
 {
-    public class CannotRemoveLastEndpointFromSubscriberError : ErrorBase
+    public class CannotRemoveLastEndpointFromSubscriberError: ErrorBase
     {
-        public CannotRemoveLastEndpointFromSubscriberError(SubscriberEntity entity) :
-            base($"Cannot remove last endpoint from Subscriber: {entity.Id}")
+        public CannotRemoveLastEndpointFromSubscriberError()
```
chomick, line 8: You could pass two params: webhook entity (this will give you subscriber id) and type of endpoints collection: webhook, dlq, callback etc. Then you do not have to map the error twice

src/CaptainHook.Domain/Errors/EndpointNotFoundInSubscriberError.cs
```diff
@@ -3,11 +3,20 @@
...
+    public class EndpointNotFoundInSubscriberError: ErrorBase
     {
-        public EndpointNotFoundInSubscriberError(string selector, SubscriberEntity entity) : 
+        public EndpointNotFoundInSubscriberError(string selector)
+            : base($"Endpoint with selector: '{selector}' does not exist in the collection")
+        {
+            Selector = selector;
+        }
+
+        public EndpointNotFoundInSubscriberError(string selector, SubscriberEntity entity):
```
chomick, line 14: As above - just accept the webhook entity (maybe even that could be part of the entity: webhookType?)

src/CaptainHook.Application/Handlers/Subscribers/DeleteWebhookRequestHandler.cs
```diff
@@ -64,7 +64,7 @@ private async Task<OperationResult<SubscriberDto>> DeleteEndpointAsync(DeleteWeb
                 return existingItem.Error;
             }
 
-            var removeResult = existingItem.Data.RemoveWebhookEndpoint(request.Selector);
+            var removeResult = existingItem.Data.RemoveWebhookEndpoint(EndpointEntity.FromSelector(request.Selector));
```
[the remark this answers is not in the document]
tabman83: Something I asked on a different PR to be consistent with AddWebhookEndpoint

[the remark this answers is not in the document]
mszymeshop: I was on a verge regarding this one. The main reason was that I wanted to use the equality comparer which takes the entire object.

src/CaptainHook.Domain/Entities/EndpointEntity.cs
```diff
@@ -43,9 +42,12 @@ public class EndpointEntity
             SetParentSubscriber(subscriber);
         }
 
-        public void SetParentSubscriber(SubscriberEntity subscriber)
+        public static EndpointEntity FromSelector(string selector) => new EndpointEntity(null, null, null, selector);
```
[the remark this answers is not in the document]
mszymeshop: The available constructor allows to do the same thing, objects are not validated until a validator is used. Simply validation of their coherence is too difficult just in the constructor.

This area is, however, something I am not going to fight over as I was on a verge regarding this one as well.

src/CaptainHook.Domain/Entities/SubscriberEntity.cs
mszymeshop: `OperationOnWebhooks` executes some common logic, for instance, checks if `Webhooks` property is available, matches errors and enriches those with `subscriber` level information which enables the subscriber id to appear in the error message. If the list of endpoints were in SubscriberEntity I believe I could simplify this the way you suggested since it is embedded in the other object, I cannot do that directly.

src/CaptainHook.Domain/Entities/SubscriberEntity.cs
mszymeshop: The reason I am handling them this way is that when I started making these changes the errors had SubscriberEntity as a dependency -> they would use the subscriber's id in the error message. I did not want to push SubscriberEntity to WebhooksEntity class as a dependency as this would create a snowball effect where I need to introduce lots of changes across the codebase (and I don't think I should have that reference available from WebhooksEntity anyway). Decided to go with this approach instead.

Another viable option is to ignore subscriber identifier in the message entirely and then I can just return any error that will be returned from WebhooksEntity here.

src/CaptainHook.Domain/Entities/WebhooksEntity.cs
mszymeshop: Sounds reasonable

src/CaptainHook.Domain/Errors/CannotRemoveLastEndpointFromSusbscriberError.cs
mszymeshop: `WebhooksEntity` does not have a subscriber id available. EndpointEntity does but (a) I don't think it should, (b) this would require me to traverse the collection of endpoints which smells fishy.

src/CaptainHook.Domain/Errors/EndpointNotFoundInSubscriberError.cs
mszymeshop: If you mean the `EndpointEntity` here then the problem is that for delete operation I don't have a valid EndpointEntity available if I am asked to move to the selector-based delete operation. For EndpointEntity with Selector (and SubscriberEntity embedded), I could try this approach.

mszymeshop:
It has been agreed that I am going to merge this and introduce changes from comments in a separate PR so that I don't block the relying changes.

PR #226, opened 2020-08-27
#82366 DTO to Entity mapper service & tests

tabman83:
The mapping code from DTOs to Entities has been extracted from handlers and moved to a separate service so all handlers can share the same piece of code.


mszymeshop (approved):

src/Tests/CaptainHook.Application.Tests/Handlers/Subscribers/UpsertWebhookRequestHandlerTests.cs
```diff
@@ -62,13 +64,22 @@ public async Task When_SubscriberDoesNotExist_Then_NewOneWillBePassedToDirectorS
                 .ReturnsAsync(new SubscriberEntity("subscriber"));
             _directorServiceMock.Setup(r => r.CreateReaderAsync(It.Is<SubscriberEntity>(rci => MatchReaderChangeInfo(rci, _defaultUpsertRequest))))
                 .ReturnsAsync(true);
+            _dtoToEntityMapper.Setup(r => r.MapEndpoint(It.IsAny<EndpointDto>()))
```
dawidkeshop, line 67: If I understand correctly this setup is repeated in every method. Wouldn't be better to move it to the constructor to avoid code duplication?

tabman83: Done!

dawidkeshop (approved):

PR #228, opened 2020-08-27
#82398 Fix Cosmos DB serialization

tabman83:
- Cosmos serialization was failing due to missing attributes
- Added missing Swagger attribute

### DIFF
--- a/src/CaptainHook.Storage.Cosmos/Models/AuthenticationSubdocument.cs
+++ b/src/CaptainHook.Storage.Cosmos/Models/AuthenticationSubdocument.cs
@@ -1,10 +1,13 @@
 ﻿using Newtonsoft.Json;
+using System.Runtime.Serialization;
 
 namespace CaptainHook.Storage.Cosmos.Models
 {
     /// <summary>
     /// Authentication model in cosmos db
     /// </summary>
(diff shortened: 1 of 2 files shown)


mszymeshop (approved):

nikhilgirraj (approved):

src/CaptainHook.Storage.Cosmos/Models/AuthenticationSubdocument.cs
```diff
@@ -1,10 +1,13 @@
 ﻿using Newtonsoft.Json;
+using System.Runtime.Serialization;
 
 namespace CaptainHook.Storage.Cosmos.Models
 {
     /// <summary>
     /// Authentication model in cosmos db
     /// </summary>
+    [KnownType(typeof(OidcAuthenticationSubdocument))]
```
chomick, line 9: Was that needed for json converter to work? I would not expect it as we do custom conversions?

chomick (approved):

src/CaptainHook.Storage.Cosmos/Models/AuthenticationSubdocument.cs
tabman83: We do custom de-serialization. I got this error while saving the document, thus serialization

PR #229, opened 2020-08-27
Task 82243 client regen fix integration test

nikhilgirraj:
Changes -
1. Added Nuget Package to CaptainHook.Api -> **Swashbuckle.AspNetCore.Newtonsoft**
2. Added subclass resolver to StartUp.cs
3. Regenerated CH.API.Client.
4. Fixed integration tests.


src/CaptainHook.Api/Startup.cs
```diff
@@ -146,6 +146,19 @@ public void ConfigureServices(IServiceCollection services)
                             }
                         });
 
+                        c.GeneratePolymorphicSchemas(
+                            type =>
+                            {
+                                if (type == typeof(AuthenticationDto))
```
mszymeshop, line 152: A candidate for reflection

PR #233, opened 2020-08-31
Add RemoveSubscriber method to SubscriberRepository & unit tests

tabman83:


src/CaptainHook.Storage.Cosmos/SubscriberRepository.cs
```diff
@@ -104,6 +104,26 @@ public Task<OperationResult<SubscriberEntity>> UpdateSubscriberAsync(SubscriberE
             return UpdateSubscriberInternalAsync(subscriberEntity);
         }
 
+        public Task<OperationResult<SubscriberEntity>> RemoveSubscriberAsync(SubscriberEntity subscriberEntity)
```
chomick, line 107: Do we need entire entity or can we just pass the minimum, which is the identifier?

tabman83: Needed to construct the partition key

chomick (approved):
I would reduce the contract for delete to mini. i.e. identifier (that should give info event name as well, i guess?)

src/CaptainHook.Storage.Cosmos/SubscriberRepository.cs
tabman83: Other option would be to accept subscriberId and eventName

chomick: is subscriber id (the object - do we still have it?) not built of event & subscriber names?

tabman83: Yes can use SubscriberId.EventName !! - SubscriberId would be enough then

chomick: if not, we may need to pass id & event name.

dawidkeshop (approved):

nikhilgirraj:
Changed the parameter to RemoveSubscriber.

PR #234, opened 2020-08-31
Downgrade Eshopworld.Devops

tabman83:
Downgrade Eshopworld.Devops package from v. 5.1.6 to v. 5.0.1 for common project which is failing to authenticate against KV.

### DIFF
--- a/src/CaptainHook.Common/CaptainHook.Common.csproj
+++ b/src/CaptainHook.Common/CaptainHook.Common.csproj
@@ -15,7 +15,7 @@
     <PackageReference Include="Autofac.ServiceFabric" Version="3.0.0" />
     <PackageReference Include="Azure.Identity" Version="1.2.2" />
     <PackageReference Include="Azure.Security.KeyVault.Secrets" Version="4.1.0" />
-    <PackageReference Include="Eshopworld.DevOps" Version="5.1.6" />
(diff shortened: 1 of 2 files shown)


dawidkeshop (approved):

src/CaptainHook.Common/CaptainHook.Common.csproj
```diff
@@ -15,7 +15,7 @@
     <PackageReference Include="Autofac.ServiceFabric" Version="3.0.0" />
     <PackageReference Include="Azure.Identity" Version="1.2.2" />
     <PackageReference Include="Azure.Security.KeyVault.Secrets" Version="4.1.0" />
```
nikhilgirraj, line 17: Can we rule out the possibility of upgrading `Azure.Security.KeyVault.Secrets` **4.0.3** -> **4.1.0**  being the issue?

nikhilgirraj (approved):

src/CaptainHook.Common/CaptainHook.Common.csproj
chomick: what version is used by DevOps package? Do we also have to downgrade this package not to have conflicts?

tabman83: yes that is not in the dependency chain of the classes used in method which is failing

PR #235, opened 2020-08-31
82663 DirectorService to support deleting readers

dawidkeshop:


src/CaptainHook.DirectorService/DirectorService.cs
```diff
@@ -218,42 +218,32 @@ public async Task<ReaderChangeResult> ApplyReaderChange(ReaderChangeBase readerC
                                     changeInfo = ReaderChangeInfo.ToBeCreated(desiredReader);
                                     break;
                                 }
+                            case DeleteReader _:
```
chomick, line 221: Don't you need to remove the subscriber from the list of configs? Right now we only add/update existing key in the list.

dawidkeshop: Yes, you're right

chomick (approved):

src/CaptainHook.DirectorService/DirectorService.cs
```diff
@@ -207,7 +207,7 @@ public async Task<ReaderChangeResult> ApplyReaderChange(ReaderChangeBase readerC
                         var desiredReader = new DesiredReaderDefinition(readerChange.Subscriber);
                         var existingReader = existingReaders.SingleOrDefault(r => desiredReader.IsTheSameService(r));
 
-                        ReaderChangeInfo changeInfo = default;
+                        ReaderChangeInfo changeInfo;
                         switch (readerChange)
                         {
```
tabman83, line 212: could you use switch pattern matching here

src/CaptainHook.DirectorService/DirectorService.cs
```diff
@@ -266,5 +258,33 @@ public async Task<ReaderChangeResult> ApplyReaderChange(ReaderChangeBase readerC
...
+                case RefreshReaderResult.Success:
+                    return ReaderChangeResult.Success;
+                case RefreshReaderResult.CreateFailed:
+                    return ReaderChangeResult.CreateFailed;
+                case RefreshReaderResult.DeleteFailed:
+                    return ReaderChangeResult.DeleteFailed;
+                default:
+                    return ReaderChangeResult.None;
+            }
+        }
```
tabman83, line 288: definitely switch pattern matching here

tabman83 (approved):

src/CaptainHook.DirectorService/DirectorService.cs
dawidkeshop: I believe you mean to use switch expression, but I don't think it's possible in that case because of failure flow there. (We're returning either ReaderChangeInfo or ReaderChangeResult, expression can have only one type returned.)

PR #236, opened 2020-08-31
#82729 Have Basic Autenthication accept KV secret name and retrieve t…

tabman83:
…he password

### DIFF
--- a/src/Tests/CaptainHook.Application.Tests/Infrastructure/EntityToDtoMapperTests.cs
+++ b/src/Tests/CaptainHook.Application.Tests/Infrastructure/EntityToDtoMapperTests.cs
@@ -19,7 +19,7 @@ namespace CaptainHook.Application.Tests.Infrastructure
         [IsUnit]
         public void MapAuthentication_When_BasicAuthenticationEntityIsUsed_Then_IsMappedToBasicAuthenticationDto()
         {
-            var entity = new BasicAuthenticationEntity("Username", "Password");
+            var entity = new BasicAuthenticationEntity("Username", "password-key-name");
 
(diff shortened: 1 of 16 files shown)


dawidkeshop (approved):

src/Tests/CaptainHook.Application.Tests/Infrastructure/EntityToDtoMapperTests.cs
```diff
@@ -28,7 +28,7 @@ public void MapAuthentication_When_BasicAuthenticationEntityIsUsed_Then_IsMapped
                 dto.Should().BeOfType(typeof(BasicAuthenticationDto));
                 BasicAuthenticationDto basicDto = (BasicAuthenticationDto)dto;
                 basicDto.Username.Should().Be("Username");
-                basicDto.Password.Should().Be("Password");
+                basicDto.PasswordKeyName.Should().Be("password-key-name");
```
chomick, line 31: How can this be passing? You need to mock the secret provider, no?

chomick: It would be quicker if you checked unit tests locally :)

chomick: Sorry, I see this is mapping entity to DTO not the other way round - the comments above are invalid here.
So the entity itself never has the value of the password/secret? Ideally entity should have those values populated at some stage but this is not part of this PR.

chomick (approved):

PR #237, opened 2020-08-31
#82730 Make SelectionRule optional and unit tests

tabman83:


src/CaptainHook.Application/Validators/Dtos/WebhooksDtoValidator.cs
```diff
@@ -11,7 +11,7 @@ namespace CaptainHook.Application.Validators.Dtos
 {
     public class WebhooksDtoValidator : AbstractValidator<WebhooksDto>
     {
-        private static JObject _jObject = new JObject();
+        private static readonly JObject _jObject = new JObject();
```
chomick, line 14: Is the usage of the jObject thread-safe?

src/Tests/CaptainHook.Application.Tests/RequestValidators/UpsertSubscriberRequestValidatorTests.cs
```diff
@@ -137,39 +193,71 @@ public void When_TwoEndpointsHaveTheSameSelector_Then_ValidationFails()
...
+            var subscriber = new SubscriberDtoBuilder().With(x => x.Webhooks, webhooks).Create();
+            var request = new UpsertSubscriberRequest("event", "subscriber", subscriber);
+
+            var result = _validator.TestValidate(request);
+
+            result.ShouldHaveValidationErrorFor(x => x.Subscriber.Webhooks.SelectionRule);
+        }
+
+        [Fact, IsUnit]
+        public void WhenOneEndpointHasDefaultSelectorAndOtherEndpointsHaveCustomSelectorsAndSelectionRuleIsDefined_Then_ValidationSucceeds()
```
chomick, line 228: convention typo: When_.... :)

chomick (approved):
I would double-check the static jObject field - to be on safe side, I would create instance field instead of class one as we don't know how thread-safe the methods used on it are (unless they are guaranteed to be thread safe).

src/CaptainHook.Application/Validators/Dtos/WebhooksDtoValidator.cs
dawidkeshop: @tabman83 - it was your change, do you know the answer?

tabman83: I only added 'readonly', while @chomick refers to it being static. The method itself does not seem to be using any internal state (see: [here](https://github.com/JamesNK/Newtonsoft.Json/blob/master/Src/Newtonsoft.Json/Linq/JToken.cs)) although the documentation does not say anything about its thread safety.

dawidkeshop (approved):

PR #238, opened 2020-08-31
#82729 - Propagate kv loading errors

tabman83:
Return proper error message when we fail to get password (or client secret) from KV - right now we get some generic error and it is unclear why it failed


dawidkeshop (approved):

src/CaptainHook.Common/CaptainHook.Common.csproj
```diff
@@ -15,7 +15,7 @@
     <PackageReference Include="Autofac.ServiceFabric" Version="3.0.0" />
     <PackageReference Include="Azure.Identity" Version="1.2.2" />
     <PackageReference Include="Azure.Security.KeyVault.Secrets" Version="4.1.0" />
-    <PackageReference Include="Eshopworld.DevOps" Version="5.1.6" />
+    <PackageReference Include="Eshopworld.DevOps" Version="5.0.1" />
```
chomick, line 18: Do we know what was the actual issue there? Should we investigate bit more?

src/CaptainHook.DirectorService/DirectorService.cs
```diff
@@ -88,7 +94,15 @@ protected override async Task RunAsync(CancellationToken cancellationToken)
                     _refreshInProgress = true;
                 }
 
-                _subscriberConfigurations = await LoadConfigurationAsync();
+                var loadConfigurationResult = await LoadConfigurationAsync();
+                if(loadConfigurationResult.IsError)
+                {
+                    // should log through bb;
```
chomick, line 100: Can you add the logging for that, please?

src/CaptainHook.Common/CaptainHook.Common.csproj
tabman83: I have created a separate tech debt task for that and I would like to leave it out of the scope of this PR

chomick (approved):

src/CaptainHook.DirectorService/DirectorService.cs
tabman83: Done

src/CaptainHook.Common/CaptainHook.Common.csproj
tabman83: Also, downgrade is only shown here because of a merge from master

chomick (approved):

PR #239, opened 2020-09-01
#82855 Fix EventReader deserializer

tabman83:


dawidkeshop (approved):

src/CaptainHook.Common/Json/AuthenticationConfigConverter.cs
```diff
@@ -8,60 +7,51 @@ namespace CaptainHook.Common.Json
...
+                AuthenticationType.Basic => new BasicAuthenticationConfig(),
+                AuthenticationType.OIDC => new OidcAuthenticationConfig(),
+                AuthenticationType.Custom => new AuthenticationConfig(),
+                AuthenticationType.None => new AuthenticationConfig(),
+                _ => new AuthenticationConfig()
+            };
 
-            return (AuthenticationType)oType;
-        }
+            serializer.Populate(jToken.CreateReader(), item);
```
nikhilgirraj, line 52: Tiny doubt - Isn't `jToken.CreateReader` going to return what `reader` variable already contains?

nikhilgirraj (approved):

nikhilgirraj (approved):

src/CaptainHook.Common/Json/AuthenticationConfigConverter.cs
chomick: I think it does - we got the token out of the reader :)

src/Tests/CaptainHook.Tests/Services/Reliable/EventReaderInitDataTests.cs
```diff
@@ -100,6 +111,48 @@ public void CanPassSubscriberConfiguration()
             eventReaderInitData.SubscriberConfiguration.Should().BeEquivalentTo(_subscriberConfiguration);
         }
 
+        [Fact]
+        [IsUnit]
+        public void CanPassAuthDataConfiguration()
```
chomick, line 116: Not a biggie but can you keep the naming convention in sync with the file?

chomick (approved):
I would probably also follow Nikhils comment.

src/CaptainHook.Common/Json/AuthenticationConfigConverter.cs
tabman83: When we do JToken.ReadFrom(reader) the reader we are passing is advanced to next position in order to read the token. So cannot reuse the same. Need to create another one from the same JToken.

src/CaptainHook.Common/Json/AuthenticationConfigConverter.cs
```diff
@@ -8,60 +7,54 @@ namespace CaptainHook.Common.Json
...
+                AuthenticationType.OIDC => new OidcAuthenticationConfig(),
+                AuthenticationType.Custom => new AuthenticationConfig(),
+                AuthenticationType.None => new AuthenticationConfig(),
+                _ => new AuthenticationConfig()
+            };
+
+            if (canParse)
             {
-                throw new JsonSerializationException("Unknown authentication type");
+                serializer.Populate(jToken.CreateReader(), item);
```
chomick, line 54: Are the readers not maintaining the state? I would imagine that using reader you are changing its state otherwise they would be useless and would have to be called 'state' not 'reader'.

PR #240, opened 2020-09-01
82617 delete subscriber handler

dawidkeshop:


src/CaptainHook.Api/Controllers/EventsController.cs
```diff
@@ -180,7 +179,7 @@ public async Task<IActionResult> DeleteSubscriber([FromRoute] string eventName,
                     CannotUpdateEntityError cannotUpdateEntityError => StatusCode(StatusCodes.Status500InternalServerError, cannotUpdateEntityError),
                     _ => StatusCode(StatusCodes.Status500InternalServerError, error)
                 },
-                Ok
+                data => NoContent()
```
tabman83, line 182: Can you apply the same change to DeleteWebhook

src/CaptainHook.Domain/Results/OperationResult.cs
```diff
@@ -23,14 +24,14 @@ public T Match<T>(Func<ErrorBase, T> leftFunc, Func<TData, T> rightFunc)
...
-                throw new ArgumentNullException(nameof(action));
-            }
+            return IsError ? Error : func(Data);
+        }
 
-            return !IsError ? (OperationResult<TResult>) action(Data) : Error;
+        public async Task<OperationResult<TResult>> Then<TResult>(Func<TData, Task<OperationResult<TResult>>> func)
+        {
+            return IsError ? Error : await func(Data);
         }
```
tabman83, line 35: Brilliant. Functional programming all the way!

src/CaptainHook.Application/Handlers/Subscribers/DeleteSubscriberRequestHandler.cs
```diff
@@ -0,0 +1,57 @@
...
+
+        private async Task<OperationResult<bool>> DeleteSubscriberAsync(DeleteSubscriberRequest request)
+        {
+            var subscriberId = new SubscriberId(request.EventName, request.SubscriberName);
+
+            return await _subscriberRepository.GetSubscriberAsync(subscriberId)
+                .Then(async subscriber => await _directorService.DeleteReaderAsync(subscriber))
+                .Then(async _ => await _subscriberRepository.RemoveSubscriberAsync(subscriberId))
+                .Then(_ => new OperationResult<bool>(true));
+        }
```
tabman83, line 55: Is there a case where we will be returning false?
If false means deletion didn't go through then we would need a reason.

----

I personally think it does not make sense to return a boolean here as we do have OperationResult.Error to indicate a failure and describe the reason.

I would return the deleted subscriber if the operation is successful (or nothing) and an Error object from OperationResult.Error if the operation failed.

src/CaptainHook.Domain/Results/OperationResult.cs
```diff
@@ -39,4 +40,17 @@ public OperationResult<TResult> IfValid<TResult>(Func<TData, TResult> action)
...
+        public static async Task<OperationResult<TOutput>> Then<TInput, TOutput>(this Task<OperationResult<TInput>> task, Func<TInput, OperationResult<TOutput>> func)
+        {
+            return (await task).Then(func);
+        }
+
+        public static async Task<OperationResult<TOutput>> Then<TInput, TOutput>(this Task<OperationResult<TInput>> task, Func<TInput, Task<OperationResult<TOutput>>> func)
+        {
+            return await (await task).Then(func);
+        }
+    }
```
tabman83, line 55: I would probably prefer having them in a separate file otherwise it feels like they are hidden

tabman83 (changes requested):

src/CaptainHook.Api/Controllers/EventsController.cs
dawidkeshop: But why? The initial plan was to return subscriber data from every endpoint, so an API consumer can immediately see his  changes. We did this for all upserts and for delete webhook andpoint.
Here we have no longer subscriber to return, so we can't return it, so only here we must return NoContent.

tabman83: I understand yes

src/CaptainHook.Application/Handlers/Subscribers/DeleteSubscriberRequestHandler.cs
dawidkeshop: I've used `bool` just because I can't use `void` there and `false` is never returned.

Regarding returning deleted subscriber - won't it be confusing to return obsolete data?

src/CaptainHook.Domain/Results/OperationResult.cs
dawidkeshop: Nice to read that. :)
I think if other people also like that way we can rewrite other handlers to avoid all these ` if (result.IsError) return result.Error` duplications.

src/CaptainHook.Domain/Results/OperationResult.cs
dawidkeshop: These are tightly coupled with `OperationResult`, but I can move it if you wish.

tabman83: May be just OperationResultExtensions in the same folder?

src/CaptainHook.Application/Handlers/Subscribers/DeleteSubscriberRequestHandler.cs
chomick: I would agree that returning the deleted record would be helpful/useful - you are passing just Id and if for whatever reason you deleted wrong record, you have a way to recreate it somehow and have its content/value.

src/CaptainHook.Domain/Results/OperationResult.cs
dawidkeshop: Actually these are `TaskExtensions`. In a separate file now.

src/CaptainHook.Application/Handlers/Subscribers/DeleteSubscriberRequestHandler.cs
dawidkeshop: Ok, that flow makes sense. Code changed.

src/CaptainHook.Domain/Results/OperationResult.cs
tabman83: Right

tabman83 (approved):

PR #242, opened 2020-09-02
#82854 Implement GET Subscriber API

tabman83:


src/CaptainHook.Api/Controllers/EventsController.cs
```diff
@@ -183,5 +188,47 @@ public async Task<IActionResult> DeleteSubscriber([FromRoute] string eventName,
...
+        {
+            try
+            {
+                var directorServiceUri = new Uri(ServiceNaming.DirectorServiceFullName);
+                var directorServiceClient = ServiceProxy.Create<IDirectorServiceRemoting>(directorServiceUri);
+                var subscribers = await directorServiceClient.GetAllSubscribersAsync();
+
+                if (subscribers.Count == 0)
+                {
+                    return StatusCode(StatusCodes.Status503ServiceUnavailable);
```
chomick, line 215: Is that what we return today? I would simply return 404 - what if you have no subscribers? why 503?

src/CaptainHook.Api/Controllers/EventsController.cs
```diff
@@ -183,5 +188,47 @@ public async Task<IActionResult> DeleteSubscriber([FromRoute] string eventName,
...
+                }
+
+                if (!subscribers.TryGetValue($"{eventName};{subscriberName}", out SubscriberConfiguration subscriberConfiguration))
```
chomick, line 218: Is the way the key is generated something that can be extracted/shared? 
Also, you will have callback and DLQ cases as well - the key generation algorithm should be shared here I think

src/CaptainHook.Api/Controllers/EventsController.cs
```diff
@@ -183,5 +188,47 @@ public async Task<IActionResult> DeleteSubscriber([FromRoute] string eventName,
...
+                {
+                    return NotFound();
+                }
+
+                AuthenticationConfigSanitizer.Sanitize(subscriberConfiguration);
+
+                return Ok(subscriberConfiguration);
```
chomick, line 225: Here you may have more than 1 entry: dlq, callback as well are considered separate subscribers.
Unless you want to have independent endpoint for getting those?

chomick (commented):
DLQ/Callback support missing.
Will you add unit tests for the controller & integration tests for security?
Also,. integration tests could be updated to use this method rather than getAll for some verifications.

src/CaptainHook.Api/Controllers/EventsController.cs
```diff
@@ -183,5 +188,47 @@ public async Task<IActionResult> DeleteSubscriber([FromRoute] string eventName,
...
+        [ProducesResponseType(StatusCodes.Status200OK)]
+        [ProducesResponseType(StatusCodes.Status503ServiceUnavailable)]
+        [ProducesResponseType(StatusCodes.Status500InternalServerError)]
+        [ProducesResponseType(StatusCodes.Status401Unauthorized)]
+        [ProducesResponseType(StatusCodes.Status404NotFound)]
+        public async Task<IActionResult> GetSubscriber([FromRoute] string eventName, [FromRoute] string subscriberName)
+        {
+            try
+            {
+                var directorServiceUri = new Uri(ServiceNaming.DirectorServiceFullName);
```
dawidkeshop, line 209: I don't think it's a good idea to make such inconsistency by not following our architecture. We should not call DirectorService directly from controllers, this should be placed in an Application Handler.

src/CaptainHook.Fabric/ApplicationParameters/Local.1Node.xml
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="utf-8"?>
 <Application xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" Name="fabric:/CaptainHook" xmlns="http://schemas.microsoft.com/2011/01/fabric">
   <Parameters>
-    <Parameter Name="KeyVault_Base_Uri" Value="https://esw-tooling-ci-we.vault.azure.net/" />
+    <Parameter Name="KeyVault_Base_Uri" Value="https://esw-tooling-ci-we-ap.vault.azure.net/" />
```
dawidkeshop, line 4: I don't think this is a good idea.

dawidkeshop (changes requested):

src/CaptainHook.Fabric/ApplicationParameters/Local.1Node.xml
tabman83: That slipped through, thanks

src/CaptainHook.Api/Controllers/EventsController.cs
tabman83: I can implement handler here but it's gonna grow more than the original estimation of course. I followed the original story description of not touching too many layers for this. @chomik?

src/CaptainHook.Api/Controllers/EventsController.cs
tabman83: 503 represents director service not ready yet (busy loading configuration).
It was initially returned when subscribers was null, but a different situation prompted to have the subscribers collection to be initialized on the director side. So this became broken and checking the count was a quick solution to get it back working.

src/CaptainHook.Api/Controllers/EventsController.cs
chomick: @dawidkeshop 
We are replicating the existing GET functionality , just adding filtering to improve usage.
These endpoints are going to be reimplemented as they do not return correct contracts.
I want to minimise effort on those so I am happy to follow the same approach we did for first GET.

When/if we implement the GET properly, we will follow the correct pattern/arch.

src/CaptainHook.Api/Controllers/EventsController.cs
tabman83: Callback is part of the subscriber configuration.
Dlq is a subscriber per se, so you just use 'dlq' as a subscriber name to get it

src/CaptainHook.Api/Controllers/EventsController.cs
tabman83: we get DLQ through the subscriber name, see below

src/CaptainHook.Api/Controllers/EventsController.cs
dawidkeshop: So it was my misunderstanding of the scope. Apologies.

dawidkeshop (approved):

src/CaptainHook.Api/Controllers/EventsController.cs
chomick: As discussed, let's keep it simple - the subscriber name (different from subscriber name used in the PUT operations) will have to be explicit. We also agreed to add a very simple endpoint to list names of existing subscribers (KV config models) so they can be queried by the GET for details.

chomick (approved):

PR #243, opened 2020-09-02
#83125 Return error with invalid auth key name

tabman83:


src/CaptainHook.Application.Infrastructure/Mappers/SubscriberEntityToConfigurationMapper.cs
```diff
@@ -154,8 +152,8 @@ private async Task<OperationResult<AuthenticationConfig>> MapAuthenticationAsync
         {
             return authenticationEntity switch
             {
-                OidcAuthenticationEntity ent => (AuthenticationConfig) await MapOidcAuthenticationAsync(ent),
-                BasicAuthenticationEntity ent => (AuthenticationConfig) await MapBasicAuthenticationAsync(ent),
+                OidcAuthenticationEntity ent => (await MapOidcAuthenticationAsync(ent)).Then<AuthenticationConfig>(x => x),
```
chomick, line 155: Why do we need "Then<>"? I thought the methods were already returning operation result?
Or is it not related to the bug?

tabman83: We need Then() because you cannot cast OperationResult<OidcAuthenticationConfig> to OperationResult<AuthenticationConfig>

chomick: what do you mean? are these different classes with  the same name? it doesn't make sense?

tabman83: Wrong comment before.

Let me explain.

OidcAuthenticationEntity ent => await MapOidcAuthenticationAsync(ent),

The version above is invalid because cannot case OperationResult<OidcAuthenticationConfig> to OperationResult<AuthenticationConfig>

That's why the old version shows:

OidcAuthenticationEntity ent => (AuthenticationConfig) await MapOidcAuthenticationAsync(ent),

which is wrong because MapOidcAuthenticationAsync returns an OperationResult<OidcAuthenticationConfig> which when cast to AuthenticationConfig will basically skip the error and return null data (when there's an error)

The latter version with Then takes care of casting to AuthenticationConfig whenever there's data available and returns the Error if that's the case.

I have tested that locally.

dawidkeshop (approved):

PR #245, opened 2020-09-03
Task 83097 Remove ServiceBus subscription on remove subscriber

nikhilgirraj:
Moved ServiceBusNamespaceExtensions to CaptainHook.Common.
Added DeleteSubscription method to DirectorService (on Delete)


src/CaptainHook.DirectorService/DirectorService.cs
```diff
@@ -285,12 +291,20 @@ private async Task<RefreshReaderResult> WrapManagerCall(ReaderChangeInfo changeI
 
             if (readerChange is DeleteReader && value == RefreshReaderResult.None)
             {
+                await DeleteSubscription(readerChange, _cancellationToken);
                 return RefreshReaderResult.Success;
             }
 
             return value;
```
tabman83, line 298: @dawidkeshop  Is this correct?
I mean - Is it correct to have the delete inside this block where we also make this check (value == RefreshReaderResult.None) ?

src/CaptainHook.Common/ServiceBus/ServiceBusNamespaceExtensions.cs
```diff
@@ -87,33 +87,51 @@ await topic.Subscriptions
...
 
             if (sbNamespace == null)
             {
-                throw new InvalidOperationException($"Couldn't find the service bus namespace {serviceBusNamespace} in the subscription with ID {azureSubscriptionId}");
+                throw new InvalidOperationException(
+                    $"Couldn't find the service bus namespace {serviceBusNamespace} in the subscription with ID {azureSubscriptionId}");
             }
 
-            return await sbNamespace.CreateTopicIfNotExists(entityName, cancellationToken);
+            return sbNamespace;
```
tabman83, line 126: Why did you remove `CreateTopicIfNotExists()` ?

src/CaptainHook.Common/ServiceBus/ServiceBusNamespaceExtensions.cs
```diff
@@ -87,33 +87,51 @@ await topic.Subscriptions
...
         }
 
+        public static async Task DeleteSubscription(string azureSubscriptionId, string sbNamespace, string topicName, string subscriptionName,
+            CancellationToken cancellationToken)
+        {
+            var sbNamespaceObj = await GetServiceBusNamespace(azureSubscriptionId, sbNamespace, cancellationToken);
```
tabman83, line 132: I don't think you should do this all over again here.

tabman83 (changes requested):
This build is also failing. Can you please check?

src/CaptainHook.DirectorService/DirectorService.cs
dawidkeshop: Yes, because `RefreshReaderResult.None` means that there were no explicit errors during deletion and there was no creation also. 
The problem is that deletion works indefinitely and doesn't provide any actual result of the operation, it can't return Success, so we must consider None as a valid situation. It's very confusing and we have a tech debt 80123 to fix that.

tabman83:
Suggested to move the ServiceBus manager to Common project and have it injected in Director service as well (other than Reader service) - the whole idea is to keep the SB interaction abstracted away in a common service/place.

src/CaptainHook.DirectorService/DirectorService.cs
```diff
@@ -285,12 +291,20 @@ private async Task<RefreshReaderResult> WrapManagerCall(ReaderChangeInfo changeI
...
         }
 
+        private async Task DeleteSubscription(ReaderChangeBase changeInfo, CancellationToken cancellationToken)
+        {
+            await ServiceBusNamespaceExtensions.DeleteSubscription(_configurationSettings?.AzureSubscriptionId, _configurationSettings?.ServiceBusNamespace,
```
dawidkeshop, line 303: I think we don't have to use `?.` here since we are sure that `_configurationSettings` is not null.

tabman83 (commented):
I have tested this locally and the EventReader does not seem to be able to read messages.
Messages sent to a newly created topic are just sitting there until they end up in DLQ.
Will try and more tests later.

As a consequence, I cannot approve this for now.

src/CaptainHook.Common/ServiceBus/ServiceBusManager.cs
```diff
@@ -0,0 +1,127 @@
...
+            }
+
+            return sbNamespace;
+        }
+
+        public async Task CreateSubscriptionAsync(string subscriptionName, string topicName, CancellationToken cancellationToken)
+        {
+            await CreateTopicIfNotExists(TypeExtensions.GetEntityName(topicName), cancellationToken);
+        }
+
```
tabman83, line 80: - This method is called CreateSubscription but it does not seem to create any subscription (hence why it was not working for me locally)
- By the look of the original one this is responsible for creating both the topic and the subscription so I would call it CreateTopicAndSubscriptionAsync

src/CaptainHook.Common/ServiceBus/ServiceBusManager.cs
```diff
@@ -0,0 +1,127 @@
...
+            return topicsList.SingleOrDefault(t => t.Name == name.ToLowerInvariant());
+        }
+
+        /// <summary>
+        /// Setups a ServiceBus <see cref="ITopic"/> given a subscription Id, a namespace topicName and the topicName of the entity we want to work with on the topic.
+        /// </summary>
+        /// <param name="topicName">The topicName of the topic entity that we are working with.</param>
+        /// <param name="cancellationToken"></param>
+        /// <returns>The <see cref="ITopic"/> contract for use of future operation if required.</returns>
+        private async Task<ITopic> CreateTopicIfNotExists(string topicName, CancellationToken cancellationToken = default)
```
tabman83, line 111: Please follow the convention and rename this to end in -Async

tabman83 (changes requested):

tabman83:
Can you please have a look at the changes suggested and test it locally. Give a me a shout if you need help for running & testing it locally.

nikhilgirraj:
Manually tested readers and subscribers removal. Could not mock the calls to write unit tests for this part.

src/CaptainHook.Common/ServiceBus/ServiceBusManager.cs
```diff
@@ -0,0 +1,147 @@
...
+                await GetServiceBusNamespaceAsync(configurationSettings.AzureSubscriptionId, configurationSettings.ServiceBusNamespace));
+
+            static TimeSpan ExponentialBackoff(int x) => TimeSpan.FromSeconds(Math.Clamp(Math.Pow(2, x), 0, RetryCeilingSeconds));
+
+            _findTopicPolicy = Policy
+                .HandleResult<ITopic>(b => b == null)
+                .WaitAndRetryForeverAsync(ExponentialBackoff);
+        }
+
+        private async Task<IServiceBusNamespace> GetServiceBusNamespaceAsync(string azureSubscriptionId, string serviceBusNamespace,
```
chomick: small suggestion - can you move private methods below public?

src/CaptainHook.Common/ServiceBus/IServiceBusManager.cs
```diff
@@ -38,5 +36,14 @@ public interface IServiceBusManager
...
         string GetLockToken(Message message);
+        
+        /// <summary>
+        /// Deletes the subscription from the topic on the current ServiceBus
+        /// </summary>
+        /// <param name="topicName">The topic where subscription exists</param>
+        /// <param name="subscriptionName">Subscription name</param>
+        /// <param name="cancellationToken"></param>
+        /// <returns></returns>
+        Task DeleteSubscriptionAsync(string topicName, string subscriptionName, CancellationToken cancellationToken);
```
chomick, line 47: tiny one - maybe group the subscription related methods together in the code

src/CaptainHook.DirectorService/DirectorService.cs
```diff
@@ -285,7 +291,9 @@ private async Task<RefreshReaderResult> WrapManagerCall(ReaderChangeInfo changeI
 
             if (readerChange is DeleteReader && value == RefreshReaderResult.None)
             {
-                return RefreshReaderResult.Success;
+                value = RefreshReaderResult.Success;
+                await _serviceBusManager.DeleteSubscriptionAsync(readerChange.Subscriber.EventType,
```
chomick, line 295: What does it mean that refresh result is None? Can it be returned as success for delete operation? And in such case, should we also remove the subscription?

chomick, line 295: What will happen if the subscription is not there? Is there an exception returned? 
In such case, we should continue as the end result is as expected - subscription is gone

src/Tests/CaptainHook.Tests/Director/DirectorServiceTests.cs
```diff
@@ -221,5 +225,30 @@ public async Task DeleteReader_When_DeletionFails_Then_DeleteFailedIsReturned()
 
             result.Should().Be(ReaderChangeResult.DeleteFailed);
         }
+
+        [Fact, IsUnit]
+        public async Task DeleteReader_When_ReaderExists_Then_DeleteSubscriptionIsCalled()
```
chomick, line 230: What about tests for deletion failures? Can they happen?

src/CaptainHook.Common/ServiceBus/ServiceBusManager.cs
```diff
@@ -0,0 +1,147 @@
...
+using Microsoft.Rest;
+using Nito.AsyncEx;
+using Polly;
+using Polly.Retry;
+using ISubscription = Microsoft.Azure.Management.ServiceBus.Fluent.ISubscription;
+
+namespace CaptainHook.Common.ServiceBus
+{
+    /// <inheritdoc/>
+    public class ServiceBusManager : IServiceBusManager
```
chomick, line 16: We may need to exclude this code from code coverage - not sure if it is testable

src/CaptainHook.Common/ServiceBus/ServiceBusManager.cs
nikhilgirraj: aren't those already below the public ones?

src/CaptainHook.DirectorService/DirectorService.cs
nikhilgirraj: For the result None see @dawidkeshop's reply here https://github.com/eShopWorld/captain-hook/pull/245#discussion_r482862130
It only results in None afaik.

nikhilgirraj: The underlying API doesn't throw any exception in case the subscription does not exist.

src/Tests/CaptainHook.Tests/Director/DirectorServiceTests.cs
nikhilgirraj: I'm sure there can be deletion failures. If the subscriber does not exist the API spits a 404 I guess. I'll try to write tests to check the behavior of the methods though.

src/CaptainHook.Common/ServiceBus/ServiceBusNamespaceExtensions.cs
nikhilgirraj: Fixed.

src/CaptainHook.Common/CaptainHook.Common.csproj
```diff
@@ -14,13 +14,14 @@
   <ItemGroup>
     <PackageReference Include="Autofac.ServiceFabric" Version="3.0.0" />
     <PackageReference Include="Azure.Extensions.AspNetCore.Configuration.Secrets" Version="1.0.0" />
-    <PackageReference Include="Azure.Identity" Version="1.2.2" />
     <PackageReference Include="Azure.Security.KeyVault.Secrets" Version="4.1.0" />
     <PackageReference Include="Eshopworld.DevOps" Version="5.0.1" />
     <PackageReference Include="Eshopworld.Telemetry" Version="3.1.5" />
-    <PackageReference Include="IdentityModel" Version="4.3.1" />
+	  <PackageReference Include="Microsoft.Azure.Management.Fluent" Version="1.34.0" />
```
mszymeshop, line 20: The formatting seems to be off here.

src/CaptainHook.Common/ServiceBus/ServiceBusManager.cs
mszymeshop: @nikhilgirraj you can use this: https://docs.microsoft.com/en-us/dotnet/api/system.diagnostics.codeanalysis.excludefromcodecoverageattribute?view=netcore-3.1

src/CaptainHook.Common/ServiceBus/ServiceBusManager.cs
```diff
@@ -0,0 +1,117 @@
...
+        public ServiceBusManager(IMessageProviderFactory factory, AsyncLazy<IServiceBusNamespace> serviceBusNamespace)
+        {
+            _factory = factory;
+            _serviceBusNamespace = serviceBusNamespace;
+
+            static TimeSpan ExponentialBackoff(int x) => TimeSpan.FromSeconds(Math.Clamp(Math.Pow(2, x), 0, RetryCeilingSeconds));
+
+            _findTopicPolicy = Policy
+                .HandleResult<ITopic>(b => b == null)
+                .WaitAndRetryForeverAsync(ExponentialBackoff);
```
mszymeshop, line 33: Do we want to keep retrying forever?

src/CaptainHook.Common/ServiceBus/ServiceBusManager.cs
```diff
@@ -0,0 +1,117 @@
...
+        }
+
+        public async Task CreateSubscriptionAsync(string subscriptionName, string topicName, bool createTopicIfNotExists, CancellationToken cancellationToken)
```
mszymeshop, line 36: Do we have a usage where `createTopicIfNotExists` is `false`? By the look at usages, it is only `true` which means we probably should not support the other case at the moment.

src/CaptainHook.Common/ServiceBus/ServiceBusManager.cs
```diff
@@ -0,0 +1,117 @@
...
+            await CreateSubscriptionIfNotExistsAsync(topic, subscriptionName, cancellationToken);
+        }
+
+        public async Task DeleteSubscriptionAsync(string topicName, string subscriptionName, CancellationToken cancellationToken)
+        {
+            var serviceBusNamespace = await _serviceBusNamespace;
+            var topic = await FindTopicAsync(serviceBusNamespace, topicName, cancellationToken);
+            if (topic != null)
+            {
+                await topic.Subscriptions.DeleteByNameAsync(subscriptionName + "3", cancellationToken);
```
mszymeshop, line 54: What does `subscriptionName + "3"` mean?

src/CaptainHook.Common/ServiceBus/ServiceBusManager.cs
```diff
@@ -0,0 +1,117 @@
...
+            }
+        }
+
+        private async Task<ISubscription> CreateSubscriptionIfNotExistsAsync(ITopic topic, string subscriptionName,
+            CancellationToken cancellationToken)
+        {
+            var subscriptionsList = await topic.Subscriptions.ListAsync(cancellationToken: cancellationToken);
+            var subscription = subscriptionsList.SingleOrDefault(s => s.Name == subscriptionName.ToLowerInvariant());
```
mszymeshop, line 62: Instead of going `ToLowerInvariant` use `StringComparison.OrdinalIgnoreCase` (there is invariant option available but for these entries we don't expect national characters)

src/CaptainHook.Common/ServiceBus/ServiceBusManager.cs
```diff
@@ -0,0 +1,117 @@
...
+            if (subscription != null) return subscription;
+
+            await topic.Subscriptions
+                .Define(subscriptionName.ToLowerInvariant())
+                .WithMessageLockDurationInSeconds(60)
```
mszymeshop, line 67: Are there defaults available or are you copying from the previous code? I would just look out for duplications of the algorithm which might rely on similar configuration. For instance, if you say 60 seconds here and in the other place it says `use default`, this might go out of sync at some point.

To me: (a) look out for duplications (might not be applicable), (b) rely on defaults if they can be used (I am not sure what the SDK allows)

src/CaptainHook.Common/ServiceBus/ServiceBusManager.cs
```diff
@@ -0,0 +1,117 @@
...
+                .WithExpiredMessageMovedToDeadLetterSubscription()
+                .WithMessageMovedToDeadLetterSubscriptionOnMaxDeliveryCount(10)
+                .CreateAsync(cancellationToken);
+
+            await topic.RefreshAsync(cancellationToken);
+            subscriptionsList = await topic.Subscriptions.ListAsync(cancellationToken: cancellationToken);
+            return subscriptionsList.Single(t => t.Name == subscriptionName.ToLowerInvariant());
```
mszymeshop, line 74: Consider StringComparison instead. Fewer temp string are created.

src/CaptainHook.Common/ServiceBus/ServiceBusManager.cs
```diff
@@ -0,0 +1,117 @@
...
+        private async Task<ITopic> CreateTopicIfNotExistsAsync(string topicName, CancellationToken cancellationToken = default)
+        {
+            var serviceBusNamespace = await _serviceBusNamespace;
+            var topic = await FindTopicAsync(serviceBusNamespace, topicName, cancellationToken);
+
+            if (topic != null) return topic;
+
+            await serviceBusNamespace.Topics
+                .Define(topicName.ToLowerInvariant())
+                .WithDuplicateMessageDetection(TimeSpan.FromMinutes(10))
```
mszymeshop, line 94: Again, is this the default or taken from the old code?

src/CaptainHook.Common/ServiceBus/ServiceBusManager.cs
```diff
@@ -0,0 +1,117 @@
...
+
+            return await _findTopicPolicy.ExecuteAsync(ct => FindTopicAsync(serviceBusNamespace, topicName, ct), cancellationToken);
+        }
+
+        public IMessageReceiver CreateMessageReceiver(string serviceBusConnectionString, string topicName, string subscriptionName, bool dlqMode)
+        {
+            return _factory.Create(serviceBusConnectionString, TypeExtensions.GetEntityName(topicName), subscriptionName.ToLowerInvariant(), dlqMode);
+        }
+
+        public string GetLockToken(Message message)
```
mszymeshop, line 105: If this is a public method it should probably be guarded against nulls etc.

mszymeshop (changes requested):

src/CaptainHook.Common/ServiceBus/ServiceBusManager.cs
mszymeshop: I can see this is old code now.

src/CaptainHook.Common/ServiceBus/ServiceBusManager.cs
mszymeshop: I can see this is old code.

mszymeshop:
A general suggestion, this is too big of a PR. The code you're moving and changing at the same time is represented as new code which makes it more difficult to review. Maybe this could have been divided into two PRs instead to emphasise clarity.

src/CaptainHook.Common/ServiceBus/ServiceBusManager.cs
nikhilgirraj: I'm not sure, the code was present there and I didn't question it. Seems like if we plan to try limited number of times, we would need to define the failure mechanism as well, like fail silently or throw the exception to the user?

src/CaptainHook.Common/ServiceBus/ServiceBusManager.cs
nikhilgirraj: It's a tricky one. Read @chomick's comment here -https://github.com/eShopWorld/captain-hook/pull/245#discussion_r486162960

src/CaptainHook.Common/ServiceBus/ServiceBusManager.cs
nikhilgirraj: That's fixed in my latest commit, I was trying to see if I need to introduce tests for non-existing subscriptions.

src/CaptainHook.Common/ServiceBus/ServiceBusManager.cs
nikhilgirraj: Old code.

src/CaptainHook.Common/ServiceBus/ServiceBusManager.cs
nikhilgirraj: Added conditional access.

src/CaptainHook.Common/ServiceBus/ServiceBusManager.cs
nikhilgirraj: This is previous code. The history is lost in refactoring but I understand what you are trying to say.

src/Tests/CaptainHook.Tests/Director/DirectorServiceTests.cs
chomick: When 404 is returned, is that translated to an exception on our side or we simply swallow the error and go as nothing happened?
The error handling for service bus manager is bit unclear to me in case of deletion errors.

src/CaptainHook.Common/ServiceBus/ServiceBusManager.cs
chomick: If we always plan to create topic, I would not use the param - no need if this is our normal behaviour. It was meant to be added only in case we need/want to use the service in different cases/modes and to have control over topic creation

mszymeshop (approved):

tabman83 (approved):

PR #246, opened 2020-09-03
Refactor servicebus extensions

nikhilgirraj:


src/CaptainHook.Common/ServiceBus/ServiceBusManager.cs
```diff
@@ -0,0 +1,127 @@
...
+
+            _exponentialBackoff = x =>
+            TimeSpan.FromSeconds(Math.Clamp(Math.Pow(2, x), 0, _retryCeilingSeconds));
+            _findTopicPolicy = Policy
+                .HandleResult<ITopic>(b => b == null)
+                .WaitAndRetryForeverAsync(_exponentialBackoff);
+
+        }
+
+        private async Task<IServiceBusNamespace> GetServiceBusNamespace(string azureSubscriptionId, string serviceBusNamespace,
```
tabman83, line 44: use -Async suffix

tabman83 (commented):
Don't forget to inject the ServiceBusManager into the container

src/CaptainHook.Common/ServiceBus/ServiceBusManager.cs
nikhilgirraj: done

src/CaptainHook.Common/ServiceBus/ServiceBusManager.cs
```diff
@@ -0,0 +1,127 @@
...
+using Nito.AsyncEx;
+using Polly;
+using Polly.Retry;
+
+namespace CaptainHook.Common.ServiceBus
+{
+    /// <inheritdoc/>
+    public class ServiceBusManager : IServiceBusManager
+    {
+        private int _retryCeilingSeconds = 30;
```
tabman83, line 23: This can be a const

src/CaptainHook.DirectorService/DirectorService.cs
```diff
@@ -289,19 +289,20 @@ private async Task<RefreshReaderResult> WrapManagerCall(ReaderChangeInfo changeI
...
         }
 
-        private async Task DeleteSubscription(ReaderChangeBase changeInfo, CancellationToken cancellationToken)
+        private async Task DeleteSubscriptionAsync(ReaderChangeBase changeInfo, CancellationToken cancellationToken)
         {
-            await ServiceBusNamespaceExtensions.DeleteSubscription(_configurationSettings?.AzureSubscriptionId, _configurationSettings?.ServiceBusNamespace,
-                changeInfo.Subscriber.EventType,
+            await _serviceBusManager.DeleteSubscriptionAsync(changeInfo.Subscriber.EventType,
                 changeInfo.Subscriber.SubscriberName, cancellationToken);
         }
```
tabman83, line 308: Do we need a method just to call one line?

tabman83 (commented):
Check SonarCloud for this PR
https://sonarcloud.io/project/issues?id=esw.captain-hook&pullRequest=246
to see where you can improve the quality and coverage

tabman83 (changes requested):
@nikhilgirraj 
You still need to inject the ServiceBusManager for both the DirectorService and the EventReaderService

src/CaptainHook.DirectorService/DirectorService.cs
```diff
@@ -289,19 +289,20 @@ private async Task<RefreshReaderResult> WrapManagerCall(ReaderChangeInfo changeI
...
 
-            if (readerChange is DeleteReader && value == RefreshReaderResult.None)
+            if (readerChange is DeleteReader)
             {
-                await DeleteSubscription(readerChange, _cancellationToken);
-                return RefreshReaderResult.Success;
+                if(value == RefreshReaderResult.None)
+                    value= RefreshReaderResult.Success;
+
+                await DeleteSubscriptionAsync(readerChange, _cancellationToken);
```
dawidkeshop, line 297: Are we sure that this will be called only in case of successful scenario?

nikhilgirraj: I am not so sure about that. Do you mean it should be called unconditionally?

src/CaptainHook.DirectorService/DirectorService.cs
nikhilgirraj: I was assuming it would be more than a line initially. Changed.

src/CaptainHook.DirectorService/DirectorService.cs
```diff
@@ -289,10 +289,12 @@ private async Task<RefreshReaderResult> WrapManagerCall(ReaderChangeInfo changeI
             var refreshResult = await _readerServicesManager.RefreshReadersAsync(new[] { changeInfo }, _cancellationToken);
             var (_, value) = refreshResult.SingleOrDefault();
 
-            if (readerChange is DeleteReader && value == RefreshReaderResult.None)
+            if (readerChange is DeleteReader)
             {
+                if(value == RefreshReaderResult.None)
+                    value= RefreshReaderResult.Success;
+
```
tabman83: Can you explain this change?

nikhilgirraj: The unit tests were written wrong which led me to believe that Deletion can result in a .None or a .Success result. but it is changed now, as @dawidkeshop explained, Deletion in ServiceFabric results in a .None but we return .Sucess to reflect successful delete event.

src/CaptainHook.DirectorService/DirectorService.cs
nikhilgirraj: Made changes with Dawid on call.

dawidkeshop (approved):

PR #250, opened 2020-09-04
82803 KV loading refactoring and tests

dawidkeshop:


src/CaptainHook.Common/Configuration/SubscribersKeyVaultProvider.cs
```diff
@@ -1,64 +1,43 @@
...
                 {
-                    this.SubscriberConfigurations.Add(
+                    subscribers.Add(
                         SubscriberConfiguration.Key(eventHandlerConfig.Type, subscriber.SubscriberName),
                         subscriber);
 
                     var path = subscriber.WebHookConfigPath;
                     ConfigParser.ParseAuthScheme(subscriber, configurationSection, $"{path}:authenticationconfig");
                     subscriber.EventType = eventHandlerConfig.Type;
                     subscriber.PayloadTransformation = subscriber.DLQMode != null ? PayloadContractTypeEnum.WrapperContract : PayloadContractTypeEnum.Raw;
```
tabman83, line 40: There's no error catching here in case anything goes wrong during:
- loading from KV
- parsing of events
Would be nice to catch those situations, log them via BB and report them back through OperationResult.

tabman83:
Suggest to use new Azure SDK which accepts SecretClient (which is already injected) and can easily be mocked

src/CaptainHook.Common/Configuration/SubscribersKeyVaultProvider.cs
```diff
@@ -0,0 +1,68 @@
...
+                        }
+                    }
+                    catch (Exception ex)
+                    {
+                        var exceptionEvent = ex.ToExceptionEvent<FailedKeyVaultConfigurationEvent>();
+                        exceptionEvent.ConfigKey = configurationSection.Key;
+                        exceptionEvent.SubscriberKey = subscriberKey;
+
+                        _bigBrother.Publish(exceptionEvent);
+                        throw;
```
tabman83, line 60: Good stuff, we finally get to get some logs when things go south.
I'd say, since we already have that nice OperationResult, we don't we use it here so we just return an error instead of crashing the DirectorService upon loading?

tabman83 (changes requested):
At this stage I don't see the need to have a separate SubscribersKeyVaultProvider and KeyVaultConfigurationLoader, I think you can unify them both under a KeyVaultConfigurationLoader

dawidkeshop:
>
> At this stage I don't see the need to have a separate SubscribersKeyVaultProvider and KeyVaultConfigurationLoader, I think you can unify them both under a KeyVaultConfigurationLoader

@tabman83: sure I can, but in that case I won't be able to write unit tests to verify logic currently placed in `SubscribersKeyVaultProvider` (foreach loop, error handling, etc.).  If we don't want to test this code I can unify both classes and the whole story will be completed.

src/CaptainHook.Common/CaptainHook.Common.csproj
```diff
@@ -13,6 +13,7 @@
 
   <ItemGroup>
     <PackageReference Include="Autofac.ServiceFabric" Version="3.0.0" />
+    <PackageReference Include="Azure.Extensions.AspNetCore.Configuration.Secrets" Version="1.0.0" />
     <PackageReference Include="Azure.Identity" Version="1.2.2" />
```
nikhilgirraj, line 17: Do we still use it?

nikhilgirraj: Seems like we don't need these packages in the Common.csproj anymore.

1. `Microsoft.Extensions.Configuration.EnvironmentVariables`
2. `Azure.Identity`
3. `IdentityModel`

nikhilgirraj (approved):

tabman83 (approved):

PR #255, opened 2020-09-09
Tests for SubscribersKeyVaultProvider

dawidkeshop:


tabman83:
Can we have a negative test? Say for example we have an event with webhooks where one is missing in the array, very much like what happened few months ago in the real KV. That's to test the error correctly bubbles up.

src/CaptainHook.Common/Configuration/EventHandlerConfig.cs
```diff
@@ -154,7 +154,7 @@ public static string Key(string typeName, string subscriberName)
         /// <remarks>
         /// It's for backward compatiblity only.
         /// </remarks>
-        public bool IsMainConfiguration { get; private set; }
+        public bool IsMainConfiguration { get; set; }
```
mszymeshop, line 157: Is this property actively used? We expect the names to be provided, or do we?

src/Tests/CaptainHook.Tests/Common/Configuration/KeyVault/SubscribersKeyVaultProviderTests.cs
```diff
@@ -0,0 +1,229 @@
...
+                ""1:webhookconfig:uri"": ""https://blah.blah.eshopworld.com"",
+                ""1:webhookconfig:httpverb"": ""POST"",
+                ""1:webhookconfig:authenticationconfig:type"": ""OIDC""
+            }}";
+
+            var provider = new SubscribersKeyVaultProvider(CreateLoaderMock(configurationData));
+            var result = provider.Load("keyvault");
+
+            result.IsError.Should().BeTrue();
+            result.Error.Should().BeOfType<KeyVaultConfigurationError>().Which.Failures.Should().HaveCount(1);
```
mszymeshop, line 215: Would be good to lock the error with the expected message. Many things can slip under if we just check the type and count I think.

dawidkeshop: I've changed this to check the actual type of the Exception returned inside the Failure. Which is always `NullReferenceException` in this case.

src/CaptainHook.Common/Configuration/EventHandlerConfig.cs
dawidkeshop: I didn't want to use the `FromWebhookConfig` method in tests to build expected results, so I've exposed this property to be able to set the value at the Builder level.
Alternatively I can try to ignore this property in `BeEquivalentTo` assertions. What do you think?

tabman83 (approved):
Thank you very much I believe this is a great improvement towards reliability

PR #258, opened 2020-09-12
#84934 Update domain entities to support callbacks

tabman83:


src/CaptainHook.Domain/Entities/SubscriberEntity.cs
```diff
@@ -59,22 +64,59 @@ public OperationResult<SubscriberEntity> SetWebhookEndpoint(EndpointEntity entit
...
+                    EndpointNotFoundInSubscriberError notFound => new EndpointNotFoundInSubscriberError(notFound.Selector, this),
+                    CannotRemoveLastEndpointFromSubscriberError _ => new CannotRemoveLastEndpointFromSubscriberError(this),
+                    _ => result.Error
+                };
+            }
+
+            return this;
+        }
+
+        private OperationResult<SubscriberEntity> OperationOnCallbacks(Func<WebhooksEntity, OperationResult<WebhooksEntity>> funcToRun)
```
chomick, line 104: i think those two methods, with small refactoring, can be merged into one

tabman83: Done with last commit

dawidkeshop (approved):

nikhilgirraj (approved):

src/CaptainHook.Domain/Entities/SubscriberEntity.cs
```diff
@@ -50,31 +55,69 @@ public void SetParentEvent(EventEntity parentEvent)
         /// </summary>
         /// <remarks>The identification is made on selector using case-insensitive comparison.</remarks>
         public OperationResult<SubscriberEntity> SetWebhookEndpoint(EndpointEntity entity)
-            => OperationOnWebhooks(webhooks => webhooks.SetEndpoint(entity.SetParentSubscriber(this)));
+            => OperationOnWebhooks(() =>
+            {
+                if(Webhooks == null)
```
chomick, line 60: That's ok but I still believe we should create instances of those properties in ctor and just maintain them on the entity via methods.

src/CaptainHook.Domain/Entities/SubscriberEntity.cs
```diff
@@ -92,5 +135,17 @@ public SubscriberEntity AddWebhooks(WebhooksEntity webhooks)
...
         }
+
+        public SubscriberEntity AddCallbacks(WebhooksEntity callbacks)
+        {
+            foreach (var callbacksEndpoint in callbacks.Endpoints)
+            {
+                callbacksEndpoint.SetParentSubscriber(this);
+            }
+
+            Callbacks = new WebhooksEntity(callbacks.SelectionRule, callbacks.Endpoints, callbacks.UriTransform);
```
chomick, line 146: If you were to have this as readonly instance created in ctor, I would expect a method on WebhookEntity to 'reset' it with new data.
Also, not a fan of creating child entities outside of the aggregate root, even for passing params - if anything, you should have a data object with bits that your method needs here (list of endpoint, selection rules and uri transforms). Your webhook entity may have more complex relationship with a parent entity and you do not want to leak/maintain that relationship outside of the aggregate root.

chomick (approved):
Acceptable :) but could be improved imo

src/CaptainHook.Domain/Entities/SubscriberEntity.cs
dawidkeshop: @chomick - I think that `SubscriberEntity` needs refactoring and more test coverage (now there are some methods which aren't covered even by a single test). I can do that as part of my story, since that will be probably the last one story touching this code for a time being.

tabman83:
I am going to merge this since @dawidkeshop offered to work to improve these entities.

PR #259, opened 2020-09-12
#84395 Update Cosmos repository to support callbacks

tabman83:


src/Tests/CaptainHook.Storage.Cosmos.Tests/SubscriberRepositoryTests.cs
```diff
@@ -443,7 +443,48 @@ public async Task AddSubscriberAsync_WithErrorWhileSaving_ReturnsCannotSaveEntit
...
+                    Endpoints = new EndpointSubdocument[] { }
+                },
+                Callbacks = new WebhookSubdocument
+                {
+                    SelectionRule = "rule",
+                    Endpoints = new EndpointSubdocument[] { }
+                }
+            };
+
+            subscriber.AddWebhooks(new WebhooksEntity("rule", new List<EndpointEntity>()));
```
chomick, line 468: what about callback? also would be good to have some actual endpoint in each. do we test mapping for different auth. options for endpoint somewhere? i think we need to have solid tests for endpoint mapping and then we can just check that correct mapping was applied to webhooks, callbacks etc - no need to repeat all tests then.

tabman83: I have improved tests for mapping in repo with the latest commit

src/CaptainHook.Storage.Cosmos/SubscriberRepository.cs
```diff
@@ -220,12 +220,20 @@ private static SubscriberDocument Deserialize(dynamic document)
 
         private SubscriberDocument Map(SubscriberEntity subscriberEntity)
         {
+            var webhooks = Map(subscriberEntity.Webhooks);
+            WebhookSubdocument callbacks = null;
+            if(null != subscriberEntity.Callbacks)
```
dawidkeshop, line 225: Why `null` on the left?

src/CaptainHook.Storage.Cosmos/SubscriberRepository.cs
```diff
@@ -269,6 +277,11 @@ private SubscriberEntity Map(SubscriberDocument subscriberDocument)
 
             subscriberEntity.AddWebhooks(Map(subscriberDocument.Webhooks, subscriberEntity));
 
+            if (null != subscriberDocument.Callbacks)
```
dawidkeshop, line 280: Why `null` on the left?

src/Tests/CaptainHook.Storage.Cosmos.Tests/SubscriberRepositoryTests.cs
```diff
@@ -130,26 +161,14 @@ public async Task GetSubscribersListAsync_WithValidEventName_MapsSubscriberDocum
...
                                 Uri = "http://test",
                                 Selector = "selector",
-                                Authentication = new OidcAuthenticationSubdocument
-                                {
-                                    ClientId = "clientid",
-                                    Scopes = new[] { "scope" },
-                                    SecretName = "secret",
-                                    Uri = "uri"
-                                }
+                                Authentication = OidcAuthenticationSubdocument
```
dawidkeshop, line 164: Maybe would be better to use `SimpleBuilder<T>` to build so complicated object graph? We can even create a new, dedicated Builder which derives from SimpleBuilder and provides default values. It can help make code shorter.

dawidkeshop (commented):
Few cosmetic suggestions.

src/CaptainHook.Storage.Cosmos/SubscriberRepository.cs
tabman83: Old habit from when you say `null ==` to avoid mistakes from  `x = null`
I can definitely change that.

src/Tests/CaptainHook.Storage.Cosmos.Tests/SubscriberRepositoryTests.cs
tabman83: Yeah, can change that

tabman83: Done

src/CaptainHook.Storage.Cosmos/SubscriberRepository.cs
tabman83: Changed

src/CaptainHook.Storage.Cosmos/SubscriberRepository.cs
tabman83: Changed

src/CaptainHook.Fabric/ApplicationParameters/CI.xml
```diff
@@ -1,7 +1,7 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Application xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" Name="fabric:/CaptainHook" xmlns="http://schemas.microsoft.com/2011/01/fabric">
   <Parameters>
-    <Parameter Name="KeyVault_Base_Uri" Value="https://esw-tooling-ci-we.vault.azure.net/" />
+    <Parameter Name="KeyVault_Base_Uri" Value="https://esw-tooling-ci-we-ap.vault.azure.net/" />
```
dawidkeshop, line 4: Is this for some test purposes or just by mistake?

dawidkeshop (approved):
The KV uri is probably a mistake, beside that it's great job!

src/CaptainHook.Fabric/ApplicationParameters/CI.xml
tabman83: Fixed!

src/CaptainHook.Storage.Cosmos/SubscriberRepository.cs
```diff
@@ -220,12 +220,21 @@ private static SubscriberDocument Deserialize(dynamic document)
 
         private SubscriberDocument Map(SubscriberEntity subscriberEntity)
         {
+            var webhooks = Map(subscriberEntity.Webhooks);
```
chomick, line 223: You could probably shrink this code to be executed as part of the assignment below, with callbacks being assigned as 
 `.. ? .. : ..`  
or simply map function to return null on null input

src/CaptainHook.Storage.Cosmos/SubscriberRepository.cs
```diff
@@ -238,6 +247,7 @@ private WebhookSubdocument Map(WebhooksEntity webhooksEntity)
             return new WebhookSubdocument
             {
                 SelectionRule = webhooksEntity.SelectionRule,
+                UriTransform = Map(webhooksEntity.UriTransform),
```
chomick, line 250: Did we miss that before? Hmm - good catch but I think our tests are lacking then - can you ensure we have unit test to cover that (and all other) mappings?

chomick (approved):

src/CaptainHook.Storage.Cosmos/SubscriberRepository.cs
tabman83: That was actually caught by tests in a subsequent PR

src/CaptainHook.Storage.Cosmos/SubscriberRepository.cs
tabman83: Done

PR #261, opened 2020-09-13
#84938 Update SubscriberEntityToConfiguration mapper for callbacks

tabman83:


tabman83:
Still in draft state because I want to add few tests.

chomick (changes requested):
we should not be duplicating mapping code. also, for dlq, we will need more then one subscriber config, right?

src/CaptainHook.Application.Infrastructure/Mappers/ISubscriberEntityToConfigurationMapper.cs
```diff
@@ -13,6 +12,6 @@ public interface ISubscriberEntityToConfigurationMapper
         /// </summary>
         /// <param name="entity">A subscriber entity</param>
         /// <returns>A subscriber configuration result or error</returns>
-        Task<OperationResult<IEnumerable<SubscriberConfiguration>>> MapSubscriberAsync(SubscriberEntity entity);
```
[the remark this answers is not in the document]
tabman83: I think you're right, I will re-look at that

src/CaptainHook.Application.Infrastructure/Mappers/SubscriberEntityToConfigurationMapper.cs
```diff
@@ -28,23 +28,85 @@ public async Task<OperationResult<IEnumerable<SubscriberConfiguration>>> MapSubs
...
+                subscriberConfig.Callback = callbackResult.Data;
+            }
+
+            return subscriberConfig;
+        }
+
+        private async Task<OperationResult<WebhookConfig>> MapCallbacksAsync(SubscriberEntity entity)
+        {
+            if (string.IsNullOrEmpty(entity.Callbacks?.SelectionRule) &&
+                entity.Webhooks?.Endpoints?.Count() == 1 &&
```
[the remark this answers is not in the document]
tabman83: I'll double check what I can do there

src/CaptainHook.Application.Infrastructure/Mappers/ISubscriberEntityToConfigurationMapper.cs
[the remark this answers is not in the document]
tabman83: This has been changed.

tabman83:
I will be adding tests with a subsequent PR

dawidkeshop (approved):

src/CaptainHook.Application.Infrastructure/DirectorService/DirectorServiceProxy.cs
```diff
@@ -21,22 +21,22 @@ public DirectorServiceProxy(ISubscriberEntityToConfigurationMapper entityToConfi
             _directorService = directorService;
         }
 
-        public async Task<OperationResult<bool>> CreateReaderAsync(SubscriberEntity subscriber)
+        public async Task<OperationResult<IEnumerable<SubscriberEntity>>> CreateReaderAsync(SubscriberEntity subscriber)
```
chomick, line 24: Why do we return list of subscriber entities and passing a single entity as input?

chomick: Did you mean to return subscriber configs (KV models?)

src/CaptainHook.Application.Infrastructure/DirectorService/IDirectorServiceProxy.cs
```diff
@@ -14,40 +15,37 @@ public interface IDirectorServiceProxy
...
+        Task<OperationResult<IEnumerable<SubscriberEntity>>> CreateReaderAsync(SubscriberEntity subscriber);
 
         /// <summary>
         /// Updates reader service for the given subscriber.
         /// </summary>
         /// <param name="subscriber">Subscriber entity defining ReaderService to be updated</param>
         /// <returns>
-        /// True if update has been invoked
-        /// False if reader does not exist
+        /// A IEnumerable<SubscriberConfiguration> if update has been invoked
```
chomick, line 30: Even here is the right type listed :) I think it doesn't make sense to return multiple entities (there will always be just one - the same as you passed as an input) but it does to return many configs as DLQ may be another one

src/CaptainHook.Application.Infrastructure/Mappers/SubscriberEntityToConfigurationMapper.cs
```diff
@@ -21,87 +21,81 @@ public SubscriberEntityToConfigurationMapper(ISecretProvider secretProvider)
...
+            var webhooksResult = await MapWebhooksAsync(entity, entity.Webhooks);
             if (webhooksResult.IsError)
             {
                 return webhooksResult.Error;
             }
 
-            /*
-            var callbacksResult = await MapCallbacksAsync(entity);
+            SubscriberConfiguration subscriberConfiguration = SubscriberConfiguration.FromWebhookConfig(webhooksResult.Data);
+            subscriberConfiguration.SubscriberName = entity.Name;
```
chomick, line 31: Do we give any special name to callbacks?
Currently, in configs, we assign the name with "-callback" suffix. Is that used in anyway or can we safely ignore that?

chomick: Also, is it safe not to create "webhookConfig" level and have only "subscribers" list in the KV models?

src/CaptainHook.Common/Configuration/EventHandlerConfig.cs
```diff
@@ -187,6 +187,19 @@ public static SubscriberConfiguration FromWebhookConfig(WebhookConfig webhookCon
             };
         }
 
+        public static SubscriberConfiguration FromWebhookConfig(WebhookConfig webhookConfig)
```
chomick, line 190: You are duplicating the logic - can you use this method inside the other one so those mappings are implemented

src/CaptainHook.Common/Configuration/EventHandlerConfig.cs
```diff
@@ -187,6 +187,19 @@ public static SubscriberConfiguration FromWebhookConfig(WebhookConfig webhookCon
...
+        {
+            return new SubscriberConfiguration
+            {
+                Name = webhookConfig.Name,
+                EventType = webhookConfig.EventType,
+                Uri = webhookConfig.Uri,
+                HttpVerb = webhookConfig.HttpVerb,
```
chomick, line 197: Which one do we use: verb or method?

src/CaptainHook.Common/Configuration/EventHandlerConfig.cs
chomick: And you missed a few fields to map

src/CaptainHook.Application.Infrastructure/Mappers/SubscriberEntityToConfigurationMapper.cs
chomick: Ignore - I thought we are mapping this in a different way - all good.

src/CaptainHook.Application.Infrastructure/Mappers/SubscriberEntityToConfigurationMapper.cs
```diff
@@ -21,87 +21,81 @@ public SubscriberEntityToConfigurationMapper(ISecretProvider secretProvider)
...
 
-            return new SubscriberConfiguration
+            return new WebhookConfig
             {
-                Name = entity.Id,
-                SubscriberName = entity.Name,
-                EventType = entity.ParentEvent.Name,
-                Uri = entity.Webhooks?.Endpoints?.FirstOrDefault()?.Uri,
-                HttpVerb = entity.Webhooks?.Endpoints?.FirstOrDefault()?.HttpVerb,
+                Name = subscriberEntity.Id,
```
chomick, line 71: So this will be a duplicate now? Callbacks and webhooks will have the same name?
Maybe we should only pass what is needed in the mappings, i.e.:
- webhook entity
- its name (string)
- event type (string)

src/CaptainHook.Application.Infrastructure/Mappers/SubscriberEntityToConfigurationMapper.cs
```diff
@@ -21,87 +21,81 @@ public SubscriberEntityToConfigurationMapper(ISecretProvider secretProvider)
...
+                subscriberConfiguration.Callback = callbackResult.Data;
             }
-            */
 
-            return new[] { webhooksResult.Data, /* callbacksResult, dlqResult */ };
+            return new SubscriberConfiguration[] { subscriberConfiguration };
         }
 
-        private async Task<OperationResult<SubscriberConfiguration>> MapWebhooksAsync(SubscriberEntity entity)
+        private async Task<OperationResult<WebhookConfig>> MapWebhooksAsync(SubscriberEntity subscriberEntity, WebhooksEntity webhooksEntity)
```
chomick, line 46: We do not need entire subscriber entity here, you just need a webhook, the name you want to give to the webhook and event type

src/CaptainHook.Common/Configuration/EventHandlerConfig.cs
chomick, line 190: Do you have unit tests for that?

src/CaptainHook.Application.Infrastructure/DirectorService/DirectorServiceProxy.cs
chomick, line 24: Do you have unit tests verifying now that callbacks are properly passed in?

chomick (changes requested):
Few concerns:
- return types (list of entity vs config)
- possibly missing unit tests (ok if you plan to do it in subsequent PRs)
- possibly incorrect mappings to subscriber config

src/CaptainHook.Application.Infrastructure/DirectorService/IDirectorServiceProxy.cs
tabman83: Changed

src/CaptainHook.Application.Infrastructure/DirectorService/DirectorServiceProxy.cs
tabman83: Changed

src/CaptainHook.Common/Configuration/EventHandlerConfig.cs
tabman83: verb - method is gathered from verb

src/CaptainHook.Application.Infrastructure/Mappers/SubscriberEntityToConfigurationMapper.cs
tabman83: there is no subscriber for callback. the SB subscription is created only for webhooks and dlq

src/CaptainHook.Common/Configuration/EventHandlerConfig.cs
tabman83: I have used this method inside the other one

tabman83: Added

src/CaptainHook.Application.Infrastructure/DirectorService/DirectorServiceProxy.cs
tabman83: The mapper is thoroughly tested!

src/CaptainHook.Application.Infrastructure/Mappers/SubscriberEntityToConfigurationMapper.cs
chomick: ok, so I understand the webhook name (in callback) is not used in any way.

src/CaptainHook.Common/Configuration/EventHandlerConfig.cs
chomick: ok, maybe we are not passing them just now (yet) but when you map from webhook config to subscriber, I would expect to have _all_ fields mapped, even if you think some of them may not be set right now.
It will save you a bug next time when timeout gets introduced to DTO and you will wonder why it is not mapped? :)

tabman83: I see your point I will change that!

src/CaptainHook.Common/Configuration/EventHandlerConfig.cs
chomick: Is one of them a calculated property? Maybe we can make one of them readonly to avoid mistakes? Does not need to be part of this PR but if quick, maybe worth doing...
If we use Verb as an input, maybe Method should be only a calculated property.

src/Tests/CaptainHook.Application.Tests/Handlers/Subscribers/DeleteSubscriberRequestHandlerTests.cs
```diff
@@ -67,7 +68,7 @@ public DeleteSubscriberRequestHandlerTests()
             _repositoryMock.Setup(x => x.GetSubscriberAsync(It.Is<SubscriberId>(id => id.Equals(new SubscriberId("event", "subscriber")))))
                .ReturnsAsync(subscriber);
             _directorServiceMock.Setup(x => x.DeleteReaderAsync(It.IsAny<SubscriberEntity>()))
-                .ReturnsAsync(true);
+                .ReturnsAsync(new List<SubscriberConfiguration>());
```
chomick, line 71: You know you cna also return Enumerable.Empty<..> - cleaner intentions and better for GC :)

src/Tests/CaptainHook.Application.Tests/Infrastructure/DirectorServiceProxyTests.cs
```diff
@@ -36,7 +36,7 @@ public async Task When_Success_Then_TrueIsReturned()
             var result = await Proxy.CreateReaderAsync(new SubscriberBuilder().Create());
 
             result.IsError.Should().BeFalse();
-            result.Data.Should().BeTrue();
+            result.Data.Should().NotBeNullOrEmpty();
```
chomick, line 39: This could have been like it before but do we have any test that verifies the actual output of the method as well?

chomick (approved):

src/Tests/CaptainHook.Application.Tests/Handlers/Subscribers/DeleteSubscriberRequestHandlerTests.cs
tabman83: I actually tried that and it wouldn't compile!

src/Tests/CaptainHook.Application.Tests/Infrastructure/DirectorServiceProxyTests.cs
tabman83: I will have it as a separate PR as that result is not being used

PR #262, opened 2020-09-14
DTO and validator updated for DLQ

dawidkeshop:


src/CaptainHook.Application/Validators/Dtos/SubscriberDtoValidator.cs
```diff
@@ -13,6 +13,9 @@ public SubscriberDtoValidator()
 
             RuleFor(x => x.Callbacks)
                 .SetValidator(new WebhooksDtoValidator("Callbacks"));
+
+            RuleFor(x => x.Dlq)
+                .SetValidator(new WebhooksDtoValidator("Callbacks"));
```
tabman83, line 18: I'd say DLQ here

tabman83 (changes requested):

src/CaptainHook.Application/Validators/Dtos/SubscriberDtoValidator.cs
dawidkeshop: sure

tabman83 (commented):
May I suggest we call them DlqHooks to reinforce the concept they are still webhooks, just fired when an event ends up in DLQ? @chomick @dawidkeshop @mszymeshop @nikhilgirraj

src/CaptainHook.Application/Validators/Dtos/SubscriberDtoValidator.cs
```diff
@@ -13,6 +13,9 @@ public SubscriberDtoValidator()
 
             RuleFor(x => x.Callbacks)
                 .SetValidator(new WebhooksDtoValidator("Callbacks"));
+
+            RuleFor(x => x.Dlq)
+                .SetValidator(new WebhooksDtoValidator("DlqHooks"));
         }
```
tabman83, line 19: Shouldnt this be DlqHooks as well?

tabman83 (approved):

src/CaptainHook.Application/Validators/Dtos/SubscriberDtoValidator.cs
dawidkeshop: Of course it should, by mistake I just didn't save all changes in this file.

PR #264, opened 2020-09-15
Delete Topic if the last subscription is being deleted.

nikhilgirraj:
Added code to delete the topic when last subscription is being deleted. Added a call to `TypeExtensions.GetEntityName(` in `ServiceBusManager.DeleteSubscription` to get the right topic name on local cluster.
Talked to @dawidkeshop to finish this in my absence.

### DIFF
--- a/src/CaptainHook.Common/ServiceBus/ServiceBusManager.cs
+++ b/src/CaptainHook.Common/ServiceBus/ServiceBusManager.cs
@@ -43,9 +43,17 @@ namespace CaptainHook.Common.ServiceBus
 
         public async Task DeleteSubscriptionAsync(string topicName, string subscriptionName, CancellationToken cancellationToken)
         {
+            var internalTopicName = TypeExtensions.GetEntityName(topicName);
             var serviceBusNamespace = await _serviceBusNamespace;
-            var topic = await FindTopicAsync(serviceBusNamespace, topicName, cancellationToken);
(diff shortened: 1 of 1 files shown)


src/CaptainHook.Common/ServiceBus/ServiceBusManager.cs
```diff
@@ -43,9 +43,17 @@ public async Task CreateTopicAndSubscriptionAsync(string subscriptionName, strin
...
         {
+            var internalTopicName = TypeExtensions.GetEntityName(topicName);
             var serviceBusNamespace = await _serviceBusNamespace;
-            var topic = await FindTopicAsync(serviceBusNamespace, topicName, cancellationToken);
-            if (topic != null)
+            var topic = await FindTopicAsync(serviceBusNamespace, internalTopicName, cancellationToken);
+            if (topic == null) return;
+
+            if (topic.SubscriptionCount == 1) // SubscriptionCount is not updated immediately after Subscriptions.DeleteByNameAsync so can't check for " == 0"
+            {
```
tabman83, line 52: We cannot work on this assumption as that might be someone else's subscription. We should find a way to determine if there are no other subscriptions besides the CH ones. For example we could think of prefixing all CH subscriptions with `ch-`.

tabman83 (changes requested):

PR #265, opened 2020-09-16
#85668 Fix an issue with SB token expiration

tabman83:


src/CaptainHook.Common/ServiceBus/ServiceBusManager.cs
```diff
@@ -109,5 +115,37 @@ private static async Task<ITopic> FindTopicAsync(IServiceBusNamespace sbNamespac
             var topicsList = await sbNamespace.Topics.ListAsync(cancellationToken: cancellationToken);
             return topicsList.SingleOrDefault(t => string.Equals(t.Name, name, StringComparison.OrdinalIgnoreCase));
         }
+
+        private async Task<IServiceBusNamespace> GetServiceBusNamespaceAsync(CancellationToken cancellationToken = default)
```
chomick, line 119: Was that the original code?

src/CaptainHook.EventReaderService/Program.cs
```diff
@@ -29,7 +29,6 @@ private static async Task Main()
                 var builder = new ContainerBuilder();
                 builder.RegisterInstance(configurationSettings).SingleInstance();
                 builder.RegisterType<MessageProviderFactory>().As<IMessageProviderFactory>().SingleInstance();
-                builder.ConfigureServiceBusManager(configurationSettings);
```
chomick, line 32: Is that our method or from MS?

tabman83: was our method in an extension class. that was injecting the cached version of servicebusnamespace object

mszymeshop: it's our extension method

src/CaptainHook.Common/ServiceBus/ServiceBusManager.cs
tabman83: Not exactly - we still need the servicebusmanager to be extracted from reader to allow the director to use it.

src/CaptainHook.Common/ServiceBus/ServiceBusManager.cs
```diff
@@ -20,16 +26,16 @@ public class ServiceBusManager : IServiceBusManager
         private const int RetryCeilingSeconds = 30;
 
         private readonly IMessageProviderFactory _factory;
-        private readonly AsyncLazy<IServiceBusNamespace> _serviceBusNamespace;
         private readonly AsyncRetryPolicy<ITopic> _findTopicPolicy;
+        private readonly ConfigurationSettings _configurationSettings;
 
-        public ServiceBusManager(IMessageProviderFactory factory, AsyncLazy<IServiceBusNamespace> serviceBusNamespace)
+        public ServiceBusManager(IMessageProviderFactory factory, ConfigurationSettings configurationSettings)
```
mszymeshop, line 32: why don't we just use a factory method for the namespace so that we can resolve it every time it is needed? We will get the separation which Nikhil achieved without the side effects?

mszymeshop: Actually, it makes sense to move the code to the class from the extensions.

src/CaptainHook.EventReaderService/Program.cs
chomick: I think that was method introduced by Nikhil - any reason it stays in the code?
Also, it was using GetServiceBusNamespaceAsync () method - is it the same method you implemented or is there another one that does the same (similar?) job?

mszymeshop (approved):

src/CaptainHook.EventReaderService/Program.cs
chomick: I see it was removed with the whole file :) my bad

chomick (approved):

PR #267, opened 2020-09-16
#85983 - move configure eda command to a new namespace

mszymeshop:
Since the command has changed, its namespaces should too.


tabman83 (approved):
Great stuff!

dawidkeshop (approved):

PR #268, opened 2020-09-16
Task 85983 merge CLI to master

mszymeshop:
The goal is to merge the Configure EDA Cli changes with master and start to export the CLI app.


tabman83 (approved):

mszymeshop:
@dawidkeshop I would like to ask for your feedback here as well as I want to make sure nothing goes to master which shouldn't be there and your approval would be advised.

dawidkeshop (approved):

PR #269, opened 2020-09-17
#86042 Improve entities and validation

tabman83:


src/CaptainHook.Domain/Entities/SubscriberEntity.cs
```diff
@@ -124,27 +103,47 @@ private OperationResult<SubscriberEntity> OperationOnWebhooks(Func<OperationResu
...
-            Webhooks = new WebhooksEntity(webhooks.SelectionRule, webhooks.Endpoints, webhooks.UriTransform);
+                WebhooksEntityType.Webhooks => SetWebhooks(webhooks),
+                WebhooksEntityType.Callbacks => SetCallbacks(webhooks),
+                _ => new ValidationError("Invalid entity")
+            };
+        }
 
+        private OperationResult<SubscriberEntity> SetWebhooks(WebhooksEntity webhooks)
+        {
+            Webhooks.SetSelectionRule(webhooks.SelectionRule);
```
dawidkeshop, line 118: Is all that code really required? The `Webhooks` property has type of `WebhooksEntity` same as the parameter `webhooks`. Wouldn't a simple assignment be enough?

dawidkeshop (approved):
Little question/suggestion.

src/CaptainHook.Domain/Entities/SubscriberEntity.cs
tabman83: This would give us the opportunity to introduce some validation at this stage

src/CaptainHook.Domain/Entities/WebhooksEntity.cs
```diff
@@ -12,51 +12,66 @@ namespace CaptainHook.Domain.Entities
...
+            SelectionRule = selectionRule;
+            Endpoints = endpoints?.ToList();
+            UriTransform = uriTransform;
+        }
 
-        public WebhooksEntity(string selectionRule, IEnumerable<EndpointEntity> endpoints, UriTransformEntity uriTransform)
+        public WebhooksEntity SetSelectionRule(string selectionRule)
         {
             SelectionRule = selectionRule;
+            return this;
```
chomick, line 64: Later on you could consider if there should be any rules e.g. can I change this to null if I have many endpoint?

src/CaptainHook.Domain/Entities/SubscriberEntity.cs
mszymeshop: `SetWebhooks` and `SetCallbacks` are the same. If you want to have a hook-in, callback mechanism, make that bit extensible. Code duplication will come up in Sonar as a smell.

src/CaptainHook.Domain/Entities/WebhooksEntity.cs
```diff
@@ -12,51 +12,66 @@ namespace CaptainHook.Domain.Entities
...
         }
 
         public OperationResult<WebhooksEntity> SetEndpoint(EndpointEntity endpoint)
         {
-            var extendedEndpointsCollection = _endpoints
+            var extendedEndpointsCollection = Endpoints
                 .Except(new [] { endpoint }, SelectorEndpointEntityEqualityComparer)
                 .Concat(new[] { endpoint });
-            var endpointsEntityForValidation = new WebhooksEntity(SelectionRule, extendedEndpointsCollection, UriTransform);
+            var endpointsEntityForValidation = new WebhooksEntity(Type, SelectionRule, extendedEndpointsCollection, UriTransform);
```
chomick, line 78: This is were validators approach is lacking imo - I have to create a new instance of the entity just to be able to validate it.  In my world, the validation would be done before I perform operation on my entity (within the entity).

But I do accept this was the existing code.

src/CaptainHook.Domain/Entities/SubscriberEntity.cs
dawidkeshop: But we don't validate at this stage. And manual mapping from some type to the exactly same type looks very strange.

src/Tests/CaptainHook.Application.Tests/Infrastructure/SubscriberEntityToConfigurationMapperTests.cs
```diff
@@ -50,7 +50,7 @@ public SubscriberEntityToConfigurationMapperTests()
         public async Task MapSubscriberAsync_WithSingleWebhookAndNoUriTransformDefined_MapsToSingleWebhook(string httpVerb, AuthenticationEntity authentication, AuthenticationConfig expectedAuthenticationConfig)
         {
             var subscriber = new SubscriberBuilder()
-                .WithWebhook("https://blah-blah.eshopworld.com/webhook/", httpVerb, string.Empty, authentication: authentication)
+                .WithWebhook("https://blah-blah.eshopworld.com/webhook/", httpVerb, "*", authentication: authentication)
```
chomick, line 53: Any reason we are changing from no-selector to any-selector?

dawidkeshop (approved):

src/Tests/CaptainHook.Domain.Tests/Entities/EndpointsCollectionValidatorTests.cs
```diff
@@ -10,12 +10,13 @@ namespace CaptainHook.Domain.Tests.Entities
 {
     public class EndpointsCollectionValidatorTests
     {
-        private readonly IValidator<IEnumerable<EndpointEntity>> _validator = new EndpointsCollectionValidator();
+        private IValidator<IEnumerable<EndpointEntity>> _validator;
```
chomick, line 13: What's the purpose/reason for the validator to be the class field?

src/Tests/CaptainHook.Domain.Tests/Entities/EndpointsCollectionValidatorTests.cs
```diff
@@ -62,14 +79,17 @@ public void Validate_MultipleEndpointsInCollectionWithoutSelector_CollectionIsNo
...
             var result = _validator.TestValidate(endpoints);
 
             result.ShouldHaveValidationErrorFor(x => x)
                 .WithErrorMessage("There cannot be multiple endpoints with the same selector");
         }
 
-        [Fact, IsUnit]
-        public void Validate_MultipleEndpointsInCollectionWithDuplicatedSelector_CollectionIsNotValid()
+        [Theory, IsUnit]
+        [InlineData(WebhooksEntityType.Webhooks)]
```
mszymeshop, line 90: Since the list will grow to Dlq shortly, does it make sense to put it as a class/property instead already?

chomick (approved):

src/Tests/CaptainHook.Domain.Tests/Entities/EndpointsCollectionValidatorTests.cs
dawidkeshop: DLQ is my story, so I think Nino left this as excercise for me.

src/CaptainHook.Application/Handlers/Subscribers/UpsertSubscriberRequestHandler.cs
```diff
@@ -92,15 +92,16 @@ private async Task<OperationResult<bool>> UpdateSubscriber(SubscriberEntity subs
...
+            var webhooks = _dtoToEntityMapper.MapWebooks(request.Subscriber.Webhooks, WebhooksEntityType.Webhooks);
 
-            var subscriberEntity = new SubscriberEntity(request.SubscriberName, new EventEntity(request.EventName))
-                .AddWebhooks(webhooks);
+            var subscriberEntity = new SubscriberEntity(request.SubscriberName, new EventEntity(request.EventName));
 
-            if(request.Subscriber.Callbacks != null)
+            subscriberEntity.SetHooks(webhooks);
+
+            if(request.Subscriber.Callbacks?.Endpoints?.Count > 0)
```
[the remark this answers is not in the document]
tabman83: On the entity, yes. This is checking the DTO instead.

src/Tests/CaptainHook.Domain.Tests/Entities/EndpointsCollectionValidatorTests.cs
tabman83: Done

src/CaptainHook.Domain/Entities/SubscriberEntity.cs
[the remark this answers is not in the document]
tabman83: Yes, moved

src/CaptainHook.Domain/Entities/WebhooksEntity.cs
```diff
@@ -12,51 +12,66 @@ namespace CaptainHook.Domain.Entities
...
+        /// <summary>
+        /// Type of entity
+        /// </summary>
+        public WebhooksEntityType Type { get; }
 
         /// <summary>
         /// Webhook endpoints
         /// </summary>
-        public IEnumerable<EndpointEntity> Endpoints => _endpoints;
+        public IList<EndpointEntity> Endpoints { get; private set; } = new List<EndpointEntity>();
```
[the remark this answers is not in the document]
tabman83: To allow remove and add operations

src/CaptainHook.Domain/Entities/WebhooksEntity.cs
tabman83: Yes, just tiny steps :)

src/Tests/CaptainHook.Application.Tests/Infrastructure/SubscriberEntityToConfigurationMapperTests.cs
tabman83: this now getting validated (thanks to changes to the entity)
the entity accepts default selector (*) or no selector where no selector is null (no empty string)

src/Tests/CaptainHook.Domain.Tests/Entities/EndpointsCollectionValidatorTests.cs
tabman83: to get the entity type when constructing the instance

mszymeshop (approved):

src/CaptainHook.Domain/Entities/WebhooksEntity.cs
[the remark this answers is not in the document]
chomick: but surely that would be managed within the class, not outside, so class cotract does not need the list?

src/Tests/CaptainHook.Domain.Tests/Entities/EndpointsCollectionValidatorTests.cs
chomick: i mean, why do you need it on class, if you initialise it for every test anyway? it just makes impression that you use it in more than one method?

src/CaptainHook.Application/Handlers/Subscribers/UpsertSubscriberRequestHandler.cs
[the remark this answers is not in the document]
chomick: are we doing the same validation in 2 places with differenr code? there should be one place we validate it; i also believe that one place belong to the domain layer, not dto validation

src/CaptainHook.Domain/Entities/SubscriberEntity.cs
chomick: actually, it depends where we believe such validation should take place. if wr had time and need, we would want to allow to update the selection rules via api call. in such case, i would expect the domain layer to be responsible for validation of the new rule and the best place would be in the entity method.
i think validators solve some problems, while intriducing more :) the important thing is that we are consistent and we do not duplicate the validations (which already seems to happen)
cc @dawidkeshop

PR #270, opened 2020-09-17
#86177 - rename the output cli to platform.eda.cli.exe

mszymeshop:
With this change, I would like to suggest that:

1. The generated CLI will be `Platform.Eda.Cli.exe`
1. The project name remains `CaptainHook.Cli` in VS
1. The default namespace remains `CaptainHook.Cli`

Points 2 and 3 are mainly there to limit the changes, as by aligning everything we really don't benefit in anything.

Let me know what you think!


dawidkeshop (approved):

tabman83 (approved):
I wish all PRs were like this!

With the risk of introducing some confusion, I still think it's much more valuable to name our tool like that for the benefit of visibility across the organization.

PR #271, opened 2020-09-17
#86026 - process subscribers independently

mszymeshop:
The goal is to be able for the CLI to process files/subscribers individually where error in one does not affect processing of the other. The output is also improved to reflect the fact that we now look into all the children folders and its files as well:

image

Remark:
No proper error handling is included at the moment, I will try to circle back here.


tabman83 (approved):
Minor comment - the `====` header and footer is used quite often so it could benefit from being produced from a separate function (e.g.: `WriteWithBox`)

mszymeshop:
> Minor comment - the `====` header and footer is used quite often so it could benefit from being produced from a separate function (e.g.: `WriteWithBox`)

Definitely makes sense 👍

PR #276, opened 2020-09-18
#86304 - align project names and namespaces for Platform.Eda.Cli

mszymeshop:
@tabman83 @dawidkeshop especially for you :) Let me know what you think :)

image


mszymeshop:
@tabman83 @dawidkeshop do you know if we should update some documentation somewhere as the name of the CLI has changed? The json generation thing was mainly used internally right?

tabman83 (approved):
Thanks for the effort!

I could've lived without it but it is definitely much nicer to have it!

dawidkeshop (approved):

PR #277, opened 2020-09-18
#86332 Append callback rules when mapping to KV model

tabman83:


dawidkeshop (approved):

mszymeshop (approved):

src/Tests/CaptainHook.Tests/Director/ConfigurationMergerTests.cs
```diff
@@ -47,6 +47,40 @@ public class ConfigurationMergerTests
                 new object[] { "PUT", BasicAuthenticationEntity, BasicAuthenticationConfig }
             };
 
+        private static readonly WebhookRequestRule StatusCodeRequestRule = new WebhookRequestRule
```
chomick, line 50: I would expect the tests to verify this only once? Any reason we also do it in the config merger?

PR #278, opened 2020-09-21
#86028 - cover ApiConsumer with unit tests

mszymeshop:
With this PR I am covering ApiConsumer with unit tests. The reported coverage is 100% for dotCover.

1. SleepTimes are configurable to so that unit tests do not need to wait too long
2. App has been run locally to verify that it's working due to some refactorings made.

### DIFF
--- a/src/Platform.Eda.Cli/Commands/ConfigureEda/ApiConsumer.cs
+++ b/src/Platform.Eda.Cli/Commands/ConfigureEda/ApiConsumer.cs
@@ -19,11 +19,16 @@ namespace Platform.Eda.Cli.Commands.ConfigureEda
 
         private readonly ICaptainHookClient _captainHookClient;
 
+        private readonly TimeSpan[] _sleepDurations;
+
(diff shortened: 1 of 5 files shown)


src/Platform.Eda.Cli/Commands/ConfigureEda/ApiConsumer.cs
```diff
@@ -47,35 +52,39 @@ public async IAsyncEnumerable<ApiOperationResult> CallApiAsync(IEnumerable<PutSu
...
+                    yield return new ApiOperationResult
+                    {
+                        File = file.File,
+                        Response = await BuildExecutionError(response.Response)
+                    };
+                }
             }
         }
 
         private static async Task<CliExecutionError> BuildExecutionError(HttpResponseMessage response)
```
tabman83, line 66: I would prefer the -Async suffix being it an awaited method

mszymeshop: I guess that makes sense, will update

tabman83 (approved):

PR #279, opened 2020-09-21
#86475 Deal with readonly dictionary in domain entity

tabman83:
The issue originally comes from the Domain layer that is directly taking the cosmos replace dictionary to instantiate a new UriTransform entity. This entity will then carry over an instance of ReadOnlyDictionary which will fail when the EntityToSubscriberConfiguration mapper tries to add the selector to the replacements dictionary.

With this PR I have enforced the concept of having a ReadOnlyDictionary even at the domain layer level (so the tests that we're not failing before because of improper mapping, will fail then)

Then I have delegated the EntityToSubscriberConfiguration mapper to make a copy of the replacements dictionary before the manipulation. This brought the tests back to passing.

### DIFF
--- a/src/CaptainHook.Application.Infrastructure/Mappers/SubscriberEntityToConfigurationMapper.cs
+++ b/src/CaptainHook.Application.Infrastructure/Mappers/SubscriberEntityToConfigurationMapper.cs
@@ -12,6 +12,8 @@ namespace CaptainHook.Application.Infrastructure.Mappers
 {
     public class SubscriberEntityToConfigurationMapper : ISubscriberEntityToConfigurationMapper
     {
+        private static readonly Dictionary<string, string> EmptyReplacementsDictionary = new Dictionary<string, string>();
+
         private static readonly WebhookRequestRule StatusCodeRequestRule = new WebhookRequestRule
         {
             Source = new SourceParserLocation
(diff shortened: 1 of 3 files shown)


src/CaptainHook.Application.Infrastructure/Mappers/SubscriberEntityToConfigurationMapper.cs
```diff
@@ -129,7 +131,7 @@ private async Task<OperationResult<WebhookConfig>> MapSingleWebhookWithNoUriTran
 
         private async Task<OperationResult<WebhookConfig>> MapForUriTransformAsync(string name, string eventType, WebhooksEntity webhooksEntity)
         {
-            var replacements = webhooksEntity.UriTransform?.Replace ?? new Dictionary<string, string>();
+            var replacements = new Dictionary<string, string>(webhooksEntity.UriTransform?.Replace ?? EmptyReplacementsDictionary);
```
mszymeshop, line 134: @tabman83 can you explain how does this change the behaviour? I can see the old code also had the fallback to empty dictionary.

mszymeshop: I think I get it, the key is to have a unique new dictionary to work with

dawidkeshop (approved):

mszymeshop (approved):

src/CaptainHook.Application.Infrastructure/Mappers/SubscriberEntityToConfigurationMapper.cs
tabman83: The webhooksEntity.UriTransform?.Replace was being passed from the Cosmos entity and it is a ReadOnlyDictionary there.

src/CaptainHook.Domain/Entities/UriTransformEntity.cs
```diff
@@ -11,7 +12,7 @@ public class UriTransformEntity
 
         public UriTransformEntity(IDictionary<string, string> replace)
         {
-            Replace = replace;
+            Replace = new ReadOnlyDictionary<string, string>(replace ?? new Dictionary<string, string>());
```
chomick, line 15: Why do we use read-only in the first place?

PR #280, opened 2020-09-21
85193 mapping DLQ data to send it to DirectorService

dawidkeshop:


src/CaptainHook.Application/Handlers/Subscribers/DirectorServiceChanger.cs
```diff
@@ -0,0 +1,57 @@
+﻿using System.Threading.Tasks;
+using CaptainHook.Application.Infrastructure.DirectorService;
+using CaptainHook.Common.Configuration;
+using CaptainHook.Domain.Entities;
+using CaptainHook.Domain.Results;
+
+namespace CaptainHook.Application.Handlers.Subscribers
+{
+    public class DirectorServiceChanger : IDirectorServiceChanger
```
chomick, line 9: I think we lost the way somewhere :) this is not what we discussed

chomick: This class should only determine what changes are to be applied but it should not be applying them.
Instead, it was supposed to return the list of operations to be executed by handler on director service.

chomick (changes requested):

src/CaptainHook.Application.Infrastructure/DirectorService/DirectorServiceProxy.cs
```diff
@@ -21,54 +19,57 @@ public DirectorServiceProxy(ISubscriberEntityToConfigurationMapper entityToConfi
             _directorService = directorService;
         }
 
-        public async Task<OperationResult<IEnumerable<SubscriberConfiguration>>> CreateReaderAsync(SubscriberEntity subscriber)
+        public async Task<OperationResult<SubscriberConfiguration>> CreateReaderAsync(SubscriberEntity subscriber)
```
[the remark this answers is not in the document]
dawidkeshop: This proxy is a border between API and `DirectorService.ApplyReaderChange` - other classes don't have to know the DirectorService contract, don't have to map to KV models and can operate only on entities. If something changes in DirectorService, we will have to update this class only.
If we make proxy to accept list of operations, it means that handlers and other classes will have to deal with KV models. It's potentially more places to update in case of any change.

Moreover, methods from proxy are used by many handlers. Should all these flows be replaced with the new approach?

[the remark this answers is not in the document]
chomick: But you are still leaking the SubscriberConfiguration out of the proxy (return values). Also, by design, the other class we were meant to create was also responsible knowing how to convert and interpret the entity class into KV config classes.

Right now, there is a false impression that we can deal with Entities only talking to Director but this is not correct - you still need to create separate calls to Director as Readers/KV models are not 100% compatible with entity models.

Also, in relation to other usages - why not use a single interaction point with the director in a coherent way? You define the operations and you pass them to the director.

src/CaptainHook.Application.Infrastructure/DirectorService/DirectorServiceProxy.cs
```diff
@@ -21,54 +19,57 @@ public DirectorServiceProxy(ISubscriberEntityToConfigurationMapper entityToConfi
...
-                    ReaderChangeResult.Success => subscriberConfig,
-                    ReaderChangeResult.NoChangeNeeded => subscriberConfig,
-                    ReaderChangeResult.CreateFailed => new ReaderCreateError(subscriber),
-                    ReaderChangeResult.DeleteFailed => new ReaderDeleteError(subscriber),
-                    ReaderChangeResult.DirectorIsBusy => new DirectorServiceIsBusyError(),
-                    ReaderChangeResult.ReaderAlreadyExist => new ReaderAlreadyExistsError(subscriber),
-                    ReaderChangeResult.ReaderDoesNotExist => new ReaderDoesNotExistError(subscriber),
-                    _ => new BusinessError("Director Service returned unknown result.")
-                };
+        public async Task<OperationResult<SubscriberConfiguration>> DeleteDlqReaderAsync(SubscriberEntity subscriber)
```
chomick, line 52: Here, you still leak the fact that DLQ from an entity cannot be processed with a normal call - I think this looks like a Francenstain a bit :)

src/CaptainHook.Application.Infrastructure/Mappers/SubscriberEntityToConfigurationMapper.cs
```diff
@@ -83,12 +84,31 @@ public async Task<OperationResult<IEnumerable<SubscriberConfiguration>>> MapSubs
...
             }
 
-            return new SubscriberConfiguration[] { subscriberConfiguration };
+            return subscriberConfiguration;
+        }
+
+        public async Task<OperationResult<SubscriberConfiguration>> MapToDlqAsync(SubscriberEntity entity)
+        {
+            if (!entity.HasDlqHooks)
+                throw new ArgumentException("Entity must contain Dlqhooks", nameof(entity));
```
mszymeshop, line 93: nit: nameof(PROPERTY)?

src/CaptainHook.Application/Handlers/Subscribers/DirectorServiceChanger.cs
mszymeshop: Such an approach would simplify the tests as well as they base on interactions at the moment and really should base on data. The application of the change-request is then a small subset of what you need to test.

src/CaptainHook.Application.Infrastructure/DirectorService/DirectorServiceProxy.cs
```diff
@@ -21,54 +19,39 @@ public DirectorServiceProxy(ISubscriberEntityToConfigurationMapper entityToConfi
...
-        public async Task<OperationResult<IEnumerable<SubscriberConfiguration>>> UpdateReaderAsync(SubscriberEntity subscriber)
+        public async Task<OperationResult<SubscriberConfiguration>> UpdateReaderAsync(SubscriberEntity subscriber)
         {
-            return await CallDirector(s => new UpdateReader { Subscriber = s }, subscriber);
+            return await _entityToConfigurationMapper.MapToWebhookAsync(subscriber)
+                   .Then(async sc => await CallDirectorService(new UpdateReader { Subscriber = sc }));
         }
 
-        public async Task<OperationResult<IEnumerable<SubscriberConfiguration>>> DeleteReaderAsync(SubscriberEntity subscriber)
+        public async Task<OperationResult<SubscriberConfiguration>> DeleteReaderAsync(SubscriberEntity subscriber)
```
chomick, line 34: I suggest we mark those methods as Obsolete/legacy in some way so they are no longer used as the preferred way of usage will be via CallDirectorService.

src/CaptainHook.Application.Infrastructure/Mappers/SubscriberEntityToConfigurationMapper.cs
dawidkeshop: But which property?

src/Tests/CaptainHook.Application.Tests/Handlers/Subscribers/DirectorServiceRequestsGeneratorTests.cs
```diff
@@ -0,0 +1,124 @@
...
+            array[1].Should().BeOfType<CreateReader>();
+        }
+
+        [Fact, IsUnit]
+        public async Task When_UpdatingDlqInExistingSubscriber_ThenWebhookUpdatedAndDlqReaderUpdated()
+        {
+            var result = await Generator.DefineChangesAsync(SubscriberWithDlq, SubscriberWithDlq);
+
+            var array = result.Data.ToArray();
+            array[0].Should().BeOfType<UpdateReader>();
```
chomick, line 110: Would be good to verify at least the names of the subscribers?

src/Tests/CaptainHook.Application.Tests/Handlers/Subscribers/DirectorServiceRequestsGeneratorTests.cs
```diff
@@ -0,0 +1,124 @@
...
+        [Fact, IsUnit]
+        public async Task When_NewSubscriberWithoutDlq_ThenOnlyWebhookReaderCreated()
+        {
+            var result = await Generator.DefineChangesAsync(SubscriberWithoutDlq, null);
+
+            result.Data.Single().Should().BeOfType<CreateReader>();
+        }
+
+        [Fact, IsUnit]
+        public async Task When_NewSubscriberWithDlq_ThenWebhookAndDlqReadersCreated()
```
chomick, line 84: Can you add also tests for updating subscriber webhooks without DLQ?

chomick (approved):

src/CaptainHook.Application.Infrastructure/DirectorService/DirectorServiceProxy.cs
dawidkeshop: Sure

src/Tests/CaptainHook.Application.Tests/Handlers/Subscribers/DirectorServiceRequestsGeneratorTests.cs
dawidkeshop: Sure

src/Tests/CaptainHook.Application.Tests/Handlers/Subscribers/DirectorServiceRequestsGeneratorTests.cs
dawidkeshop: Good catch!

PR #283, opened 2020-09-22
Task 86409 CLI unit tests

nikhilgirraj:


src/Tests/Platform.Eda.Cli.Tests/Commands/ConfigureEda/ConfigureEdaCommandTests.cs
```diff
@@ -0,0 +1,115 @@
...
+            _configureEdaCommand.NoDryRun = true;
+            var response = new HttpOperationResponse<object>
+            {
+                Response = new HttpResponseMessage(HttpStatusCode.Accepted)
+                {
+                    Content = new StringContent("Request accepted.")
+                }
+            };
+
+            _mockCaptainHookClient.Setup(client =>
```
mszymeshop, line 44: I think you're testing too many things at the moment, ApiConsumer should be moved to abstraction in order to mock it.  At the moment you use the concrete implementation (as we don't use DI for it).

src/Tests/Platform.Eda.Cli.Tests/Commands/ConfigureEda/ConfigureEdaCommandTests.cs
```diff
@@ -0,0 +1,115 @@
...
+        private readonly ConfigureEdaCommand _configureEdaCommand;
+        private readonly Mock<ICaptainHookClient> _mockCaptainHookClient;
+
+        public ConfigureEdaCommandTests(ITestOutputHelper output) : base(output)
+        {
+            _mockCaptainHookClient = new Mock<ICaptainHookClient>();
+            _configureEdaCommand = new ConfigureEdaCommand(SubscribersDirectoryProcessorTests.GetMockFileSystem(), _mockCaptainHookClient.Object);
+        }
+
+        [Fact]
```
mszymeshop, line 30: you are missing IsUnit attributes

src/Tests/Platform.Eda.Cli.Tests/Commands/ConfigureEda/SubscribersDirectoryProcessorTests.cs
```diff
@@ -0,0 +1,88 @@
...
+                result.Data.Should().HaveCount(1);
+                result.Data.Single().Request.Should().NotBeNull();
+                result.Data.Single().Request.SubscriberName.Should().Be("test-sub");
+            }
+        }
+
+        public static Dictionary<string, MockFileData> GetMockFiles()
+        {
+            Dictionary<string, MockFileData> mockFiles = new Dictionary<string, MockFileData>();
+            mockFiles.Add("TestFiles/sample1.json", new MockFileData(@"
```
mszymeshop, line 63: We also want to support files from different folders so at least some tests should try to cover that.

src/Tests/Platform.Eda.Cli.Tests/Commands/ConfigureEda/SubscribersDirectoryProcessorTests.cs
```diff
@@ -0,0 +1,88 @@
...
+        readonly SubscribersDirectoryProcessor _subscribersDirectoryProcessor;
+
+        public SubscribersDirectoryProcessorTests()
+        {
+            _subscribersDirectoryProcessor = new SubscribersDirectoryProcessor(GetMockFileSystem());
+        }
+
+        public static MockFileSystem GetMockFileSystem()
+        {
+            return new MockFileSystem(GetMockFiles(), Environment.CurrentDirectory);
```
mszymeshop, line 24: Is `Environment.CurrentDirectory` required or can we use a dummy location instead?

src/Tests/Platform.Eda.Cli.Tests/Commands/ConfigureEda/SubscribersDirectoryProcessorTests.cs
```diff
@@ -0,0 +1,88 @@
...
+        }
+
+        [Theory]
+        [InlineData(null, typeof(ArgumentNullException))]
+        [InlineData("", typeof(ArgumentException))]
+        public void ProcessDirectory_WhenPathEmptyOrNull_ThrowsException(string s, Type exceptionType)
+        {
+            Assert.Throws(exceptionType, () =>
```
mszymeshop, line 32: Might be worth sticking to FluentAssertions since you use those below already.

src/Tests/Platform.Eda.Cli.Tests/Commands/ConfigureEda/SubscribersDirectoryProcessorTests.cs
```diff
@@ -0,0 +1,88 @@
...
+                _subscribersDirectoryProcessor.ProcessDirectory(s));
+        }
+
+        [Theory]
+        [InlineData("a-nonexistent-path")]
+        public void ProcessDirectory_WithNonExistentPath_ReturnsError(string s)
```
mszymeshop, line 38: `s` doesn't tell much at the moment, can you name that parameter so that it is clear what kind of value it represents?

src/Tests/Platform.Eda.Cli.Tests/Commands/ConfigureEda/SubscribersDirectoryProcessorTests.cs
```diff
@@ -0,0 +1,88 @@
...
+        {
+            var result = _subscribersDirectoryProcessor.ProcessDirectory(s);
+            result.IsError.Should().BeTrue();
+            result.Error.Should().BeOfType<CliExecutionError>();
```
mszymeshop, line 42: Could you consider checking the message available in CLIError? What's in there might be important to the flow.

src/Tests/Platform.Eda.Cli.Tests/Commands/ConfigureEda/SubscribersDirectoryProcessorTests.cs
```diff
@@ -0,0 +1,88 @@
...
+        [InlineData("TestFiles")]
+        public void ProcessDirectory_WithValidFile_ReturnsNonNullObject(string s)
+        {
+            var result = _subscribersDirectoryProcessor.ProcessDirectory(s);
+
+            using (new AssertionScope())
+            {
+                result.IsError.Should().BeFalse();
+                result.Data.Should().HaveCount(1);
+                result.Data.Single().Request.Should().NotBeNull();
```
mszymeshop, line 55: If the Data collection doesn't have a single element but has 0 or 2, this will throw an exception. Ok, the test will still fail which is what you want but Assertionscope will not catch it and will not catch issues below. Try to change this block of code into a single assertion with Fluent Assertions. I can help if needed.

src/Tests/Platform.Eda.Cli.Tests/Commands/ConfigureEda/SubscribersDirectoryProcessorTests.cs
nikhilgirraj: I think I can change that to a hypothetical directory as well. Let me change it.

src/Tests/Platform.Eda.Cli.Tests/Commands/ConfigureEda/SubscribersDirectoryProcessorTests.cs
nikhilgirraj: Could not do that because FluentAssertions do(es) not allow an option to provide `exceptionType` parameter.

src/Tests/Platform.Eda.Cli.Tests/Commands/ConfigureEda/ConfigureEdaCommandTests.cs
nikhilgirraj: Another PR :)

src/Tests/Platform.Eda.Cli.Tests/Commands/ConfigureEda/SubscribersDirectoryProcessorTests.cs
nikhilgirraj: Changed

mszymeshop (approved):

PR #285, opened 2020-09-22
Upgrade dependencies and get rid of warnings

tabman83:
### This PR achieves the 0-warnings target for Captain Hook!

- Updated some dependency packages to the latest version (except for Service Fabric related which cannot be update until we update the SF cluster)
- Downgraded SF mocks library to avoid the dependency on newer unsupported version of SF libraries
- Added few comments to get rid of warnings
- Removed unused code to get rid of warnings
- Some projects now target x64 to get rid of warnings (as they had x64 dependencies)
- Swagger updated to the latest version which now supports polymorphism out of the box


dawidkeshop (approved):

nikhilgirraj (approved):

src/Tests/CaptainHook.Api.Tests/appsettings.DEVELOPMENT.json
```diff
@@ -4,5 +4,5 @@
   "RefreshingTokenProviderOptions": {
     "Issuer": "https://security-sts.ci.eshopworld.net"
   },
-  "CaptainHookApiUri": "https://localhost:24010"
+  "CaptainHookApiUri": "https://platform-captainhook-api.ci.eshopworld.net/"
```
nikhilgirraj, line 7: I kind of think this is unnecessary. If this configuration is only going to be active on Development machines then does it not make sense to keep the localhost URL there?

tabman83: It does! It just slipped with the latest commit. Sorry, will revert

tabman83:
This PR has been updated with further changes to Release build to reduce the number of warnings so I'd like to get another review.

src/CaptainHook.Common/CaptainHook.Common.csproj
```diff
@@ -15,11 +15,11 @@
     <PackageReference Include="Autofac.ServiceFabric" Version="3.0.0" />
     <PackageReference Include="Azure.Extensions.AspNetCore.Configuration.Secrets" Version="1.0.0" />
     <PackageReference Include="Azure.Security.KeyVault.Secrets" Version="4.1.0" />
-    <PackageReference Include="Eshopworld.DevOps" Version="5.0.1" />
+    <PackageReference Include="Eshopworld.DevOps" Version="5.1.7" />
     <PackageReference Include="Eshopworld.Telemetry" Version="3.1.5" />
     <PackageReference Include="Microsoft.Azure.Management.Fluent" Version="1.34.0" />
-  	<PackageReference Include="Microsoft.Azure.ServiceBus" Version="4.1.4" />
-    <PackageReference Include="Microsoft.Extensions.Configuration.Binder" Version="3.1.7" />
+  	<PackageReference Include="Microsoft.Azure.ServiceBus" Version="5.0.0" />
```
chomick, line 21: Do we understand the consequences of the upgrade here?

src/CaptainHook.Api/Startup.cs
```diff
@@ -145,19 +145,11 @@ public void ConfigureServices(IServiceCollection services)
...
-                                {
-                                    return new[] {
-                                        typeof(OidcAuthenticationDto),
-                                        typeof(BasicAuthenticationDto)
-                                    };
-                                }
-                                return Enumerable.Empty<Type>();
-                            });
+                        c.UseOneOfForPolymorphism();
+                        c.SelectSubTypesUsing(baseType =>
```
chomick, line 149: What's the difference between those two definitions (old vs new)?

src/CaptainHook.EventHandlerActor/Handlers/Authentication/MmAuthenticationHandler.cs
```diff
@@ -53,7 +53,7 @@ public override async Task<string> GetTokenAsync(CancellationToken cancellationT
 
             if (authProviderResponse.StatusCode != HttpStatusCode.Created || authProviderResponse.Content == null)
             {
-                throw new Exception("didn't get a token from the provider");
+                throw new InvalidOperationException("didn't get a token from the provider");
```
chomick, line 56: Is that authentication handler ever used?
Should we remove it?
If not, should we allow for this authentication to be available via API as well (for backward compatibility)?

chomick (commented):
Good job!
Have a look at the questions raised.

src/CaptainHook.Api/Startup.cs
tabman83: Old Swagger did not have out of the box support for polymorphism, new one has. I have checked the generated swagger models between this PR locally and old one in CI and they match.

src/CaptainHook.EventHandlerActor/Handlers/Authentication/MmAuthenticationHandler.cs
tabman83: Not sure if this is ever used but this is only aimed at getting rid of the warning (SonarCloud related in this case)

src/CaptainHook.Common/CaptainHook.Common.csproj
tabman83: The only breaking change is around pk in SB which we are not using at the minute.
Rest of changes is bugfixes and small improvements so pretty safe I'd say.
See: https://github.com/Azure/azure-sdk-for-net/blob/Microsoft.Azure.ServiceBus_5.0.0/sdk/servicebus/Microsoft.Azure.ServiceBus/CHANGELOG.md

chomick: Let's make sure we run some e2e tests as part of this change (on CI); perf testing would be better but I think we may struggle with availability of Preprod for that.

src/CaptainHook.EventHandlerActor/Handlers/Authentication/MmAuthenticationHandler.cs
chomick: I know :) I thought we could verify if this class is actually used and in what cases and then quickly verify if those cases are actually configured in our KVs.
I agree this does not block the PR but I thought it is worth checking while we are at it.

chomick (approved):

src/CaptainHook.EventHandlerActor/Handlers/Authentication/MmAuthenticationHandler.cs
tabman83: I'll do that with a separate PR

PR #288, opened 2020-09-23
#86801 Update DTO and validators

tabman83:
The webhook model now takes a `payloadTransform` parameter whose accepted values are:
null, empty string, Request, Response, OrderConfirmation, PlatformOrderConfirmation, EmptyCart
all values are case insensitive.
null value and empty string are considered the default: do not apply any payload transformation.
The `payloadTransform` is only accepter for Webhooks and DLQs. Validation will throw a failure if such parameter is added to callbacks.


chomick (approved):

src/CaptainHook.Application/Validators/Dtos/WebhooksDtoValidator.cs
```diff
@@ -30,6 +34,16 @@ public WebhooksDtoValidator(string subject)
                 .Must(NotContainMultipleEndpointsWithTheSameSelector)
                     .WithMessage("There cannot be multiple endpoints with the same selector");
 
+            RuleFor(x => x.PayloadTransform)
```
mszymeshop, line 37: I can be ok with this approach but I think a bit of segregation would help readability and extensibility since this is now a very complex validator. Consider this: https://docs.fluentvalidation.net/en/latest/including-rules.html

mszymeshop (approved):

src/CaptainHook.Application/Validators/Dtos/WebhooksDtoValidator.cs
tabman83: I will raise separate PR for this one.

tabman83: See here: https://github.com/eShopWorld/captain-hook/pull/297

PR #289, opened 2020-09-24
Task 86703 switch to security token

mszymeshop:
With this PR I am:

1. Removing the use of Services.Testing package for token requests
1. Relying on Security.Rest to request token with the production-grade package
1. Removing Autofac for CLI and relying on ServicesCollection instead
1. Tests to be refined in future PRs


src/Platform.Eda.Cli/Program.cs
```diff
@@ -74,14 +80,20 @@ public static int Main(string[] args)
...
+        private static IServiceProvider SetupServices()
         {
-            var serviceProviderFactory = new AutofacServiceProviderFactory();
+            var collection = new ServiceCollection();
 
-            var builder = new ContainerBuilder();
-            builder.RegisterAssemblyModules(Assembly.GetExecutingAssembly());
-            builder.RegisterInstance(new ApiClientFixture().GetApiClient());
-            return serviceProviderFactory.CreateServiceProvider(builder);
+            collection.AddHttpClient();
```
mszymeshop, line 87: I have moved to ServiceCollection as we don't really need Autofac and it is easier for HttpClient registration as those extension methods come in from MS.

src/Platform.Eda.Cli/appsettings.CI.json
```diff
@@ -1,8 +1,11 @@
...
-  "RefreshingTokenProviderOptions": {
-    "Issuer": "https://security-sts.ci.eshopworld.net"
-  },
-  "CaptainHookApiUri": "https://platform-captainhook-api.ci.eshopworld.net/"
+  "ConfigureEdaCommand": {
+    "CaptainHookUrl": "https://platform-captainhook-api.ci.eshopworld.net",
+    "RefreshingTokenProviderOptions": {
+      "Scopes": [ "" ],
+      "ClientId": "platform.captainhook.api.client",
+      "ClientSecret": "",
```
mszymeshop, line 7: Due to the lack of client-secret, this will not work when run. As a separate PR I am going to embed those in the KV and read from there.

src/Platform.Eda.Cli/Program.cs
```diff
@@ -74,14 +80,20 @@ public static int Main(string[] args)
...
+            collection.AddTransient<IFileSystem, FileSystem>();
+
+            collection.AddSingleton<BuildCaptainHookProxy>(serviceProvider =>
+            {
+                var clientFactory = serviceProvider.GetRequiredService<IHttpClientFactory>();
+                return environment => ApiConsumer.BuildApiConsumer(clientFactory, environment);
+            });
```
tabman83, line 94: Can we not simply add the ApiConsumer as a dependency and it can get the client factory from the injected services and the environment can either be loaded with devopssdk or injected as well.

src/Platform.Eda.Cli/Commands/ConfigureEda/BuildCaptainHookProxy.cs
```diff
@@ -0,0 +1,4 @@
+﻿namespace Platform.Eda.Cli.Commands.ConfigureEda
+{
+    public delegate ApiConsumer BuildCaptainHookProxy(string environment);
```
chomick, line 3: Suggest to add `Delegate` suffix for this class as usage is bit confusing.

src/Platform.Eda.Cli/Commands/ConfigureEda/ConfigureEdaCommand.cs
```diff
@@ -18,12 +17,12 @@ namespace Platform.Eda.Cli.Commands.ConfigureEda
     public class ConfigureEdaCommand
     {
         private readonly IFileSystem _fileSystem;
-        private readonly ICaptainHookClient _captainHookClient;
+        private readonly BuildCaptainHookProxy _captainHookBuilder;
 
-        public ConfigureEdaCommand(IFileSystem fileSystem, ICaptainHookClient captainHookClient)
+        public ConfigureEdaCommand(IFileSystem fileSystem, BuildCaptainHookProxy captainHookBuilder)
```
chomick, line 22: Can you just inject the client/proxy and add correct instance to the IoC?

src/Platform.Eda.Cli/Commands/ConfigureEda/ConfigureEdaConfig.cs
```diff
@@ -0,0 +1,12 @@
+﻿using System;
+using EShopworld.Security.Services.Rest;
+
+namespace Platform.Eda.Cli.Commands.ConfigureEda
+{
+    public class ConfigureEdaConfig
```
chomick, line 6: `ConfigureEdaCommandConfig`?

chomick (approved):

src/Platform.Eda.Cli/Commands/ConfigureEda/ApiConsumer.cs
```diff
@@ -25,6 +31,19 @@ public class ApiConsumer
...
+        {
+            var configuration = EswDevOpsSdk.BuildConfiguration(AssemblyLocation, environment);
+            var configureEdaConfig = configuration.GetSection(nameof(ConfigureEdaCommand)).Get<ConfigureEdaConfig>();
+
+            var tokenProvider = new RefreshingTokenProvider(clientFactory, BigBrother.CreateDefault("", ""), configureEdaConfig.RefreshingTokenProviderOptions);
+
+            var client = new CaptainHookClient(configureEdaConfig.CaptainHookUrl, new TokenCredentials(tokenProvider));
+            return new ApiConsumer(client, null);
+        }
+
```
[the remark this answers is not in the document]
mszymeshop: I am going to move big brother out so that we can still inject it. At the moment it isn't used in CLI at all, so wanted to have something here so that we can continue with the features.

As you can see in the Program.cs I am registering a delegate as I need to defer the creation of the class as the environment isn't known until the command gets run.

src/Platform.Eda.Cli/Commands/ConfigureEda/BuildCaptainHookProxy.cs
mszymeshop: Will do

src/Platform.Eda.Cli/Commands/ConfigureEda/ConfigureEdaCommand.cs
mszymeshop: I am afraid I cannot do this as the environment which I need for this to read configuration isn't known until the command gets run. It is only then when I can build the configuration and create the proxy based on what I find in the config.

src/Platform.Eda.Cli/Commands/ConfigureEda/ApiConsumer.cs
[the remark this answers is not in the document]
tabman83: That makes sense thank you.

src/Platform.Eda.Cli/Commands/ConfigureEda/ConfigureEdaCommand.cs
mszymeshop: FYI I checked in the cli documentation (we use a package available in github) but there was nothing there that would directly solve my problem.

tabman83 (approved):

src/Platform.Eda.Cli/Commands/ConfigureEda/ConfigureEdaConfig.cs
mszymeshop: will do

src/Platform.Eda.Cli/Program.cs
mszymeshop: at this moment the environment isn't known until the command gets run. I didn't find a better way to address this apart from deferring the config load and apiconsumer creation like this.

src/Platform.Eda.Cli/Commands/ConfigureEda/ConfigureEdaConfig.cs
mszymeshop: Addressed in https://github.com/eShopWorld/captain-hook/pull/292

PR #290, opened 2020-09-24
#86802 Update entities and DTO-entity-DTO mappers to support payload …

tabman83:
…transformations


src/Tests/CaptainHook.TestsInfrastructure/TestsData/ValidHttpVerbs.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System.Collections;
+using System.Collections.Generic;
+
+namespace CaptainHook.TestsInfrastructure.TestsData
+{
+    public class ValidHttpVerbs : IEnumerable<object[]>
+    {
+        public IEnumerator<object[]> GetEnumerator()
```
chomick, line 8: Why do we need list of lists of verbs?

chomick (approved):

mszymeshop (approved):

PR #292, opened 2020-09-24
#86839 - Read client secret from the Key Vaults

mszymeshop:


tabman83 (approved):

src/Platform.Eda.Cli/Commands/ConfigureEda/ApiConsumer.cs
```diff
@@ -36,9 +36,13 @@ public class ApiConsumer
...
             var configuration = EswDevOpsSdk.BuildConfiguration(AssemblyLocation, environment);
-            var configureEdaConfig = configuration.GetSection(nameof(ConfigureEdaCommand)).Get<ConfigureEdaConfig>();
+            var configureEdaConfig = configuration.BindSection<ConfigureEdaCommandConfig>(nameof(ConfigureEdaCommandConfig));
 
-            var tokenProvider = new RefreshingTokenProvider(clientFactory, BigBrother.CreateDefault("", ""), configureEdaConfig.RefreshingTokenProviderOptions);
+            var refreshingTokenProviderOptions = configuration.BindSection<RefreshingTokenProviderOptions>(
+                $"{nameof(ConfigureEdaCommandConfig)}:{nameof(RefreshingTokenProviderOptions)}",
+                c => c.AddMapping(m => m.ClientSecret, configureEdaConfig.ClientKeyVaultSecretName));
+
+            var tokenProvider = new RefreshingTokenProvider(clientFactory, BigBrother.CreateDefault("", ""), refreshingTokenProviderOptions);
```
chomick, line 45: @macuistin - this gimnastics are needed if we use the refreshing token structure from STS package - maybe you could think how this can be improved so we don't have to map it twice between the structs?

src/Platform.Eda.Cli/appsettings.CI.json
```diff
@@ -1,10 +1,11 @@
 {
-  "ConfigureEdaCommand": {
+  "KeyVaultInstanceName": "esw-platform-ci-we",
+  "ConfigureEdaCommandConfig": {
+    "ClientKeyVaultSecretName": "CaptainHookSubscriberApi--ClientSecret",
```
chomick, line 4: Can we use the correct naming convention, please?

chomick: Also, this name should be defined on the models attribute so it is clear what value ends up in the model

src/Platform.Eda.Cli/appsettings.json
```diff
@@ -1,10 +1,11 @@
 {
-  "ConfigureEdaCommand": {
+  "KeyVaultInstanceName": "esw-platform-ci-we",
+  "ConfigureEdaCommandConfig": {
+    "ClientKeyVaultSecretName": "CaptainHookSubscriberApi--ClientSecret",
```
chomick, line 4: As above- the name should not be part of config - all env's should have the KV entry defined under the same name and it should be explicitly referenced in the code

chomick (approved):
Don't want to block PR but we will need to update the KV fetching to be in line with new naming conventions/standards.

src/Platform.Eda.Cli/appsettings.CI.json
mszymeshop: @chomick this is changed in https://github.com/eShopWorld/captain-hook/pull/293 as this PR came earlier you still see some early version when it comes to config.

src/Platform.Eda.Cli/appsettings.json
mszymeshop: See https://github.com/eShopWorld/captain-hook/pull/293 and let me know if this matches your requirements

src/Platform.Eda.Cli/Commands/ConfigureEda/ApiConsumer.cs
mszymeshop: the reason it looks like that is that I need to treat the `refreshingTokenProviderOptions` as a separate thing as `AddMapping` does not work for embedded references. And since I am using `EswDevOpsSdk.BuildConfiguration` I cannot modify the `ConfigurationBuilder` as I don't have access to it.

PR #293, opened 2020-09-24
#86704 - prepare configuration for environments

mszymeshop:
The secrets are not in key-vaults yet. As soon as this is accepted I am going to prepare a PR in the other repo to push them.

### DIFF
--- a/src/Platform.Eda.Cli/appsettings.json
+++ b/src/Platform.Eda.Cli/appsettings.json
@@ -1,10 +1,10 @@
 {
-  "KeyVaultInstanceName": "esw-platform-ci-we",
+  "KeyVaultInstanceName": "esw-platform-ci",
   "ConfigureEdaCommandConfig": {
-    "ClientKeyVaultSecretName": "CaptainHookSubscriberApi--ClientSecret",
+    "ClientKeyVaultSecretName": "sts--sts-secret--platform-captainhook-api-client",
     "CaptainHookUrl": "https://platform-captainhook-api.ci.eshopworld.net",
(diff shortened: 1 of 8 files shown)


tabman83 (approved):

src/Platform.Eda.Cli/appsettings.json
```diff
@@ -1,10 +1,10 @@
 {
-  "KeyVaultInstanceName": "esw-platform-ci-we",
+  "KeyVaultInstanceName": "esw-platform-ci",
   "ConfigureEdaCommandConfig": {
-    "ClientKeyVaultSecretName": "CaptainHookSubscriberApi--ClientSecret",
+    "ClientKeyVaultSecretName": "sts--sts-secret--platform-captainhook-api-client",
```
chomick, line 4: Still, would not have it in the config; instead, would reference it in the attribute of the config model in the code

mszymeshop: I don't have access to that config model, it comes from the package. I guess it has been foreseen by the creators that is why they have provided an option to use `AddMapping`. For that one, I decided it is better to have it in the config instead of hard-coding that ONLY value in the code.

chomick: Eventually we will use enums etc so this will have to go from config anyway and be passed to (better) methods in the dev-ops package. The fact that we will construct that string eventually makes me believe this should end up in the code anyway but will not block PR for that.

chomick (approved):

PR #299, opened 2020-09-27
#73570 Avoid double DLQ suffix in service name

tabman83:
This change avoids the creation of event reader services with a double `-DLQ` suffix in the name. The code in `CaptainHook.Common\ServiceNaming.cs` is already taking care of appending said suffix for DLQ readers.


dawidkeshop (changes requested):
It's not that easy. 

DirectorService internally holds a Dictionary with keys build from  `EventType` and `SubscriberName`, which means that combination of these values must be unique. After this change you'll get same key twice for every subscriber with DLQ defined, so DirectorService will fail.

tabman83:
> It's not that easy.
> 
> DirectorService internally holds a Dictionary with keys build from `EventType` and `SubscriberName`, which means that combination of these values must be unique. After this change you'll get same key twice for every subscriber with DLQ defined, so DirectorService will fail.

Thank you for the clarification, I am happy to abandon this

tabman83:
Closing as not needed.

PR #300, opened 2020-09-28
Fixed - EDA CLI - subscriber data not fully passed

nikhilgirraj:
Regenrated API client and Added test to cover the Json parsing bug.

### DIFF
--- a/src/Tests/Platform.Eda.Cli.Tests/Commands/ConfigureEda/SubscribersDirectoryProcessorTests.cs
+++ b/src/Tests/Platform.Eda.Cli.Tests/Commands/ConfigureEda/SubscribersDirectoryProcessorTests.cs
@@ -135,7 +135,59 @@ namespace Platform.Eda.Cli.Tests.Commands.ConfigureEda
             result.IsError.Should().BeTrue();
             result.Error.Should().BeOfType<CliExecutionError>();
         }
-        
+
+
+        [Fact, IsUnit]
+        public void ProcessDirectory_JsonFileWithCallbacksDlqHooks_ReturnsValidObject()
+        {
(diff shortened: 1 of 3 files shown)


src/Tests/Platform.Eda.Cli.Tests/Commands/ConfigureEda/SubscribersDirectoryProcessorTests.cs
```diff
@@ -135,7 +135,59 @@ public void ProcessDirectory_WithInvalidJsonFile_ReturnsError()
             result.IsError.Should().BeTrue();
             result.Error.Should().BeOfType<CliExecutionError>();
         }
-        
+
+
+        [Fact, IsUnit]
+        public void ProcessDirectory_JsonFileWithCallbacksDlqHooks_ReturnsValidObject()
```
chomick, line 141: Is there a way to test it without mocking file loading?
Should we be able to define the model (content) of the file somehow and when loaded successfully, that model should be used within the app, not reference to the physical file?

chomick: I think when you load your file, you should have:
- name of the event
- name of the subscriber
- request content
- source file name (for reporting etc)

This model is what we should be using within the app - that would nicely reduce your dependencies and make the whole processing more modular

chomick (approved):

src/Tests/Platform.Eda.Cli.Tests/Commands/ConfigureEda/SubscribersDirectoryProcessorTests.cs
nikhilgirraj: Do you mean passing the file contents directly to some method and checking the return value? I think we can do that, a small bit of refactoring is required, that's all.

PR #303, opened 2020-09-28
87403 delete DLQ reader during subscriber removal

dawidkeshop:


src/CaptainHook.Application/Handlers/Subscribers/DeleteSubscriberRequestHandler.cs
```diff
@@ -53,10 +59,31 @@ private async Task<OperationResult<SubscriberDto>> DeleteSubscriberAsync(DeleteS
...
+                return existingItem.Error;
+            }
+
+            var subscriberEntity = existingItem.Data;
+
+            return await _entityToConfigurationMapper.MapToWebhookAsync(subscriberEntity)
+                .Then(async webhookConfig => await _directorService.CallDirectorService(new DeleteReader(webhookConfig)))
+                .Then<SubscriberConfiguration, SubscriberEntity>(async _ => await DeleteDlqIfExists(subscriberEntity))
+                .Then(async _ => await _subscriberRepository.RemoveSubscriberAsync(subscriberId))
+                .Then<SubscriberId, SubscriberDto>(_ => _entityToDtoMapper.MapSubscriber(subscriberEntity));
```
tabman83, line 74: Super! Don't we have an overload of Then which accepts task-async ?

src/CaptainHook.Application/Handlers/Subscribers/DeleteSubscriberRequestHandler.cs
```diff
@@ -53,10 +59,31 @@ private async Task<OperationResult<SubscriberDto>> DeleteSubscriberAsync(DeleteS
...
+        }
+
+        private async Task<SubscriberEntity> DeleteDlqIfExists(SubscriberEntity subscriberEntity)
+        {
+            if (subscriberEntity.HasDlqHooks)
+            {
+                return await _entityToConfigurationMapper.MapToDlqAsync(subscriberEntity)
+                    .Then(async dlqConfig => await _directorService.CallDirectorService(new DeleteReader(dlqConfig)))
```
tabman83, line 82: Same as before

tabman83 (approved):
See the comments above

src/CaptainHook.Application/Handlers/Subscribers/DeleteSubscriberRequestHandler.cs
dawidkeshop: Sure, you're right!

PR #306, opened 2020-09-29
#87391 - update Key Vault loading for custom config

mszymeshop:
The configuration setup has been changed. The previous approach would fail for KeyVault secret loading as with the previous implementation it would re-create a default configuration builder. We cannot use the default builder as our environment specification comes as a command line parameter rather than env-variable which is the default.


dawidkeshop (approved):

src/Platform.Eda.Cli/Commands/ConfigureEda/ApiConsumer.cs
```diff
@@ -32,21 +29,26 @@ public class ApiConsumer : IApiConsumer
...
+                .UseDefaultConfigs(environment: environment)
+                .AddKeyVaultSecrets(new Dictionary<string, string> {
+                    { "sts--sts-secret--platform-captainhook-api-client", "ConfigureEdaCommandConfig:RefreshingTokenProviderOptions:ClientSecret" }
+                })
+                .Build();
+
+            var configureEdaCommandConfig = configuration
+                .GetSection(nameof(ConfigureEdaCommandConfig))
+                .Get<ConfigureEdaCommandConfig>();
+
```
tabman83, line 45: Would it be better to use BindSection and move `sts--sts-secret--platform-captainhook-api-client` to an attribute on the relevant property of the ConfigureEdaCommandConfig class?

tabman83: Here's what I mean: https://github.com/eShopWorld/devops/#bind-to-a-class-specifying-key-vault-mappings

mszymeshop: I'm afraid I have to do the mapping @tabman83 as in line 38 as I don't have access to the configuration entity since it comes from the security nuget package.

tabman83: Happy to approve then. Thanks for clarifying.

tabman83 (approved):

PR #307, opened 2020-09-29
Bug 87386 CLI meaningful output

nikhilgirraj:
- Did not change the multi line output for error files since it made sense to use multiple lines to output the details like file name, event name, subscriber name, error code and message.
- Could not find a direct way to club all messages about a single file because the CLI does its job in two steps by design - read all and process all.


src/Platform.Eda.Cli/Commands/ConfigureEda/ConfigureEdaCommand.cs
```diff
@@ -104,11 +107,27 @@ await foreach (var apiResult in api.CallApiAsync(subscriberFiles.Where(f => !f.I
...
+                            operationDescription = "created";
+                            break;
+                        case HttpStatusCode.Accepted:
+                            operationDescription = "updated";
+                            break;
+                        default:
+                            operationDescription = "unknown result";
+                            break;
+                    }
+
```
tabman83, line 128: You can use switch expressions here: https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/operators/switch-expression

tabman83 (changes requested):

tabman83 (approved):

src/Platform.Eda.Cli/Commands/ConfigureEda/ConfigureEdaCommand.cs
```diff
@@ -104,11 +107,20 @@ await foreach (var apiResult in api.CallApiAsync(subscriberFiles.Where(f => !f.I
...
+                        $" subscriber '{apiResult.Request.SubscriberName}'. Error details: ", apiResultResponse.Error.Message);
                 }
                 else
                 {
-                    writer.WriteNormal($"File '{fileRelativePath}' has been processed successfully");
+                    string operationDescription = apiResult.Response.Data.Response.StatusCode switch
+                    {
+                        HttpStatusCode.Created => "created",
+                        HttpStatusCode.Accepted => "updated",
+                        _ => "unknown result"
```
chomick, line 119: I would include the status code if we can't recognise it

nikhilgirraj: That makes sense. Let me change it.

chomick (approved):

PR #310, opened 2020-09-30
#87968 Fix no-auth validation

tabman83:


dawidkeshop (approved):

src/Tests/CaptainHook.Tests/Json/AuthenticationDtoJsonConverterTests.cs
```diff
@@ -89,7 +106,7 @@ public void WhenAuthenticationTypeIsNone_ThenItIsDeserializedProperly()
         [InlineData(null)]
         [InlineData("")]
         [IsUnit]
-        public void WhenAuthenticationTypeIsInvalid_ThenItIsDeserializedAsNull(string authType)
+        public void WhenAuthenticationTypeIsInvalid_ThenItIsDeserializedAsInvalidAuthentication(string authType)
```
nikhilgirraj, line 109: I am still getting a grasp on the test method naming convention. These ones feel like they should be `AuthenticationType_WhenInvalid_DeserializedAsInvalidAuthentication`

nikhilgirraj (approved):

src/Tests/CaptainHook.Tests/Json/AuthenticationDtoJsonConverterTests.cs
tabman83: You are most probably right but I am not gonna change all the test names now as it outside of the scope for this PR. I will address those separately if that's ok.

PR #311, opened 2020-09-30
Added method to parse variables from the JSON file.

nikhilgirraj:
Intended flow of parsing -
1. Parse Vars as Dict
2. Replace vars in content string 
3. Replace params in content string
4. Parse Content string as PutSubscriberRequest as before
5. Profit?


src/Platform.Eda.Cli/Commands/ConfigureEda/SubscribersDirectoryProcessor.cs
```diff
@@ -3,8 +3,11 @@
 using System.IO;
 using System.IO.Abstractions;
 using System.Linq;
+using System.Text.RegularExpressions;
 using CaptainHook.Domain.Results;
+using JetBrains.Annotations;
```
dawidkeshop, line 8: What's that? `JetBrains.Annotations` sounds like something very internal, non-standard and something which can't be trusted. :)

src/Platform.Eda.Cli/Commands/ConfigureEda/SubscribersDirectoryProcessor.cs
```diff
@@ -69,5 +75,39 @@ private PutSubscriberFile ProcessFile(string fileName)
...
+        public static Dictionary<string, Dictionary<string, string>> GetFileVars(JObject fileContent)
+        {
+            if (fileContent.ContainsKey("vars"))
+            {
+                var varsDictionary = JsonConvert.DeserializeObject<Dictionary<string, Dictionary<string, JToken>>>(
+                        fileContent["vars"].ToString());
+
+                var outputVarsDictionary = new Dictionary<string, Dictionary<string, string>>();
+
+                foreach (var (key1, innerDict) in varsDictionary)
```
dawidkeshop, line 88: I would use a name which is more descriptive than `key1`. Like `propertyKey` maybe?

src/Platform.Eda.Cli/Commands/ConfigureEda/SubscribersDirectoryProcessor.cs
```diff
@@ -69,5 +75,39 @@ private PutSubscriberFile ProcessFile(string fileName)
...
+                {
+                    outputVarsDictionary[key1] = new Dictionary<string, string>();
+                    foreach (var (key2, val) in innerDict)
```
dawidkeshop, line 91: I would use a name which is more descriptive than `key2`. Like `environmentsKey` maybe?
And `envsValues` or `valuesForEnv` than `innerDict`?

I mean to use a name which tells the purpose not the internal file location.

src/Platform.Eda.Cli/Commands/ConfigureEda/SubscribersDirectoryProcessor.cs
```diff
@@ -69,5 +75,39 @@ private PutSubscriberFile ProcessFile(string fileName)
...
+                    {
+                        if (key2.Contains(","))
+                        {
+                            foreach (var key in key2.Split(','))
```
dawidkeshop, line 95: `singleEnv` instead of `key`?

src/Platform.Eda.Cli/Commands/ConfigureEda/SubscribersDirectoryProcessor.cs
```diff
@@ -69,5 +75,39 @@ private PutSubscriberFile ProcessFile(string fileName)
...
+                            {
+                                outputVarsDictionary[key1][key] = val.ToString(Formatting.None);
+                            }
+                        }
+                        else // convert to string
+                        {
+                            outputVarsDictionary[key1][key2] = val.Type == JTokenType.String ? val.ToString() : val.ToString(Formatting.None);
```
dawidkeshop, line 102: Why this logic is different than in the previous `if` statement?
I think we can have a `string` and `node` regardless if this is set for one or several environments.

dawidkeshop (commented):
Couple naming suggestions. Maybe not crucial, but in couple weeks or months will help us to understand what and why is happening in the code.

src/Platform.Eda.Cli/Commands/ConfigureEda/SubscribersDirectoryProcessor.cs
nikhilgirraj: Yes, I was trying something and ended up not using this namespace. This should have been automatically removed. :/

src/Platform.Eda.Cli/Commands/ConfigureEda/SubscribersDirectoryProcessor.cs
nikhilgirraj: Wanted to use something that would say key at level 1 and key at level 2. Later understood that the second level is always env. Will change it now.

src/Platform.Eda.Cli/Commands/ConfigureEda/SubscribersDirectoryProcessor.cs
```diff
@@ -69,5 +74,39 @@ private PutSubscriberFile ProcessFile(string fileName)
...
+                var varsDict = JsonConvert.DeserializeObject<Dictionary<string, Dictionary<string, JToken>>>(fileContent["vars"]!.ToString());
+
+                var outputVarsDict = new Dictionary<string, Dictionary<string, string>>();
+
+                foreach (var (propertyKey, innerDict) in varsDict)
+                {
+                    outputVarsDict[propertyKey] = new Dictionary<string, string>();
+                    foreach (var (envKey, val) in innerDict)
+                    {
+                        var stringVal = val.Type == JTokenType.String ? val.ToString() : val.ToString(Formatting.None);
```
chomick, line 91: I would not make this distinction - why not always toString (none)?

src/Platform.Eda.Cli/Commands/ConfigureEda/SubscribersDirectoryProcessor.cs
dawidkeshop: `ctrl + R, ctrl + G` is a friend. ;)

src/Platform.Eda.Cli/Commands/ConfigureEda/SubscribersDirectoryProcessor.cs
```diff
@@ -69,5 +74,39 @@ private PutSubscriberFile ProcessFile(string fileName)
...
+                        if (envKey.Contains(","))
+                        {
+                            foreach (var singleEnv in envKey.Split(','))
+                            {
+                                outputVarsDict[propertyKey][singleEnv] = stringVal;
```
chomick, line 96: Just be sure to:
- trim spaces
- check for empty env names

src/Platform.Eda.Cli/Commands/ConfigureEda/SubscribersDirectoryProcessor.cs
```diff
@@ -69,5 +74,39 @@ private PutSubscriberFile ProcessFile(string fileName)
...
+
+        public static Dictionary<string, Dictionary<string, string>> GetFileVars([NotNull] JObject fileContent)
+        {
+            if (fileContent.ContainsKey("vars"))
+            {
+                var varsDict = JsonConvert.DeserializeObject<Dictionary<string, Dictionary<string, JToken>>>(fileContent["vars"]!.ToString());
+
+                var outputVarsDict = new Dictionary<string, Dictionary<string, string>>();
+
+                foreach (var (propertyKey, innerDict) in varsDict)
```
chomick, line 86: proeprtyKey == variableName?

src/Platform.Eda.Cli/Commands/ConfigureEda/SubscribersDirectoryProcessor.cs
```diff
@@ -69,5 +74,39 @@ private PutSubscriberFile ProcessFile(string fileName)
                 };
             }
         }
+
+        public static Dictionary<string, Dictionary<string, string>> GetFileVars([NotNull] JObject fileContent)
+        {
+            if (fileContent.ContainsKey("vars"))
```
chomick, line 80: See if you can break down this method - it got rather complex

src/Platform.Eda.Cli/Commands/ConfigureEda/SubscribersDirectoryProcessor.cs
```diff
@@ -52,7 +54,10 @@ private PutSubscriberFile ProcessFile(string fileName)
             try
             {
                 var content = _fileSystem.File.ReadAllText(fileName);
-                var request = JsonConvert.DeserializeObject<PutSubscriberRequest>(content);
+                var contentJObject = JObject.Parse(content);
+                var varsDictionary = GetFileVars(contentJObject);
+
+                var request = JsonConvert.DeserializeObject<PutSubscriberRequest>(contentJObject.ToString());
```
chomick, line 60: Are you planning to run the vars substitution here or later?
You may want to think how the substitution would work - that may help you with the definition of what data structures you actually will need to hold the vars in.

Because we are expecting correct Json as an input, we have to push the variable name into a string in our Json request definition. That means you can't simply apply string-replace for the variables as you would also have to remove the string literals from the output.

That leaves you with only one option: iterate over all elements in your JObject (recursively) and if any node has a string value, apply vars replacement in that string (that would be nice, little service you could have: input is a string & vars dictionary and output is converted string or value/JToken). 
The result would have to be attached to the node instead original string. 

Here is a second problem: you will need to differentiate between simple string replacement vs substituting string variable usage with complex object the variable represents.

We wanted to be able to support entries like:

```
  "vars": { "name" : "random string" }

  ...
  "some-property": "regular string {var:name} with variable inside",
  ...
```

but also this:

```
  "vars": { "name" : { "complex": "object"}  }

  ...
  "some-property": "{var:name}",
  ...
```
So when you attempt to substitute the variable name with a value, there must be some logic applied that will detect what is the value you try to inject and whether it is a valid substitution in a given context. For the case below, the usage of variable below would be incorrect:

```
  "vars": { "name" : { "complex": "object"}  }

  ...
  "some-property": "my string here and {var:name}", // we can't really do that - complex object can be substituted only when the variable is used on its own
  ...
```

Also, on top of all that, you will have parameters substitution on the variables and other content as well :)

src/Tests/Platform.Eda.Cli.Tests/Commands/ConfigureEda/SubscribersDirectoryProcessorTests.cs
```diff
@@ -193,14 +194,66 @@ public void ProcessDirectory_JsonFileWithCallbacksDlqHooks_ReturnsValidObject()
...
             result.Should().BeEquivalentTo(expected, opt => opt
                 .Excluding(info => info.SelectedMemberInfo.Name == "Authentication")
-                .Excluding(info => info.SelectedMemberInfo.MemberType == typeof(FileInfoBase)));   
+                .Excluding(info => info.SelectedMemberInfo.MemberType == typeof(FileInfoBase)));
+
+        }
+
+
+        [Fact, IsUnit]
+        public void ProcessDirectory_JsonFileWithVars_ReturnsValidObject()
```
chomick, line 203: You definitively want to break this functionality out into some service/class.
As you start adding more logic, testing this will be impossible.

src/Platform.Eda.Cli/Commands/ConfigureEda/SubscribersDirectoryProcessor.cs
tabman83: I suggest you have separate services for
- variables substitution
- parameters substitution
This way the entire code is more testable (you can test the different substitution services in isolation)

Also I'd suggest you use JSONPath to retrieve the variables and parameters

tabman83 (changes requested):
I think that if you look at the set of operations needed to perform an event configuration you can define a pipeline of stages:
1. read the file and parse the json
2. replace params
3. replace variables
4 call the API

Right now everything is done in one single place for all files at the same time. If you were to change your code in a way where you have different services, and each service only takes care of one stage per one file, then you could have your pipeline in place. Once this is achieved you would ideally process one file at a time (thus having a single log as requested by @chomick ) - but since the pipeline can be asynchronous as soon as one file moves to the second stage you can start processing another one.

I don't expect all the stages to take a similar amount of time so this won't be an extreme parallelization.

Hope I was clear.

src/Platform.Eda.Cli/Commands/ConfigureEda/SubscribersDirectoryProcessor.cs
dawidkeshop: This task is only about reading vars from JSON file, variables and parameters substitutions have separated, follow-up tasks.

src/Platform.Eda.Cli/Commands/ConfigureEda/JsonProcessor/SubscriberFileParser.cs
```diff
@@ -0,0 +1,69 @@
...
+            catch (Exception ex)
+            {
+                return new CliExecutionError(ex.Message);
+            }
+        }
+
+        public Dictionary<string, Dictionary<string, string>> GetFileVars(JObject fileContent)
+        {
+            if (fileContent.ContainsKey("vars"))
+            {
```
tabman83, line 38: If this is going to be part of a separate service (step) then you can remove it from here.

src/Platform.Eda.Cli/Commands/ConfigureEda/JsonProcessor/SubscribersDirectoryProcessor.cs
```diff
@@ -8,15 +8,17 @@
...
+        private readonly ISubscriberFileParser _subscriberFileParser;
 
-        public SubscribersDirectoryProcessor(IFileSystem fileSystem)
+        public SubscribersDirectoryProcessor(IFileSystem fileSystem, ISubscriberFileParser subscriberFileParser)
         {
             _fileSystem = fileSystem;
+            _subscriberFileParser = subscriberFileParser;
         }
 
         public OperationResult<IEnumerable<PutSubscriberFile>> ProcessDirectory(string inputFolderPath)
```
tabman83, line 24: This is currently returning an empty collection of PutSubscriberFile. Is it just a stub?

src/Platform.Eda.Cli/Commands/ConfigureEda/JsonProcessor/SubscriberFileParser.cs
nikhilgirraj: Just noticed. Removing it.

src/Platform.Eda.Cli/Commands/ConfigureEda/JsonProcessor/SubscribersDirectoryProcessor.cs
nikhilgirraj: It is a stub for now. The next task is to chain all the components together. As part of that - 
1. This one (SubscriberDirectoryProcessor) will be changed to return a IEnumerable<string> list of files instead.
2. Control code will be moved to a new class (that chains all steps) from [ConfigureEdaCommand](https://github.com/eShopWorld/captain-hook/blob/5359812b7d8c0d74429ced3c30046414aeaefc87/src/Platform.Eda.Cli/Commands/ConfigureEda/ConfigureEdaCommand.cs)

src/Platform.Eda.Cli/Commands/ConfigureEda/JsonProcessor/IJsonVarsParser.cs
```diff
@@ -0,0 +1,10 @@
+﻿using System.Collections.Generic;
+using Newtonsoft.Json.Linq;
+
+namespace Platform.Eda.Cli.Commands.ConfigureEda.JsonProcessor
+{
+    public interface IJsonVarsParser
+    {
+        public Dictionary<string, JToken> GetFileVars(JObject fileContent, string environmentName);
```
chomick, line 8: Can you add comments/documentation for the interface/method, please?
We need to start doing this for all public interfaces

src/Platform.Eda.Cli/Commands/ConfigureEda/JsonProcessor/JsonVarsParser.cs
```diff
@@ -0,0 +1,42 @@
...
+            {
+                foreach (var (envKey, val) in innerDict)
+                {
+                    if (string.Equals(environmentName, envKey, StringComparison.OrdinalIgnoreCase))
+                    {
+                        outputVarsDict[propertyKey] = val;
+                    }
+                    else if (envKey.Contains(","))
+                    {
+                        foreach (var singleEnv in envKey.Split(','))
```
chomick, line 29: Trim spaces and check for empty env name?
Also, should we have a list of valid env's or do we accept any string here?

src/Platform.Eda.Cli/Commands/ConfigureEda/JsonProcessor/JsonVarsParser.cs
```diff
@@ -0,0 +1,42 @@
...
+                        {
+                            if (string.Equals(environmentName, singleEnv, StringComparison.OrdinalIgnoreCase))
+                                outputVarsDict[propertyKey] = val;
+                        }
+                    }
+                }
+            }
+
+            fileContent.Remove("vars");
```
chomick, line 38: Are we happy with this side-effect? Will it be clear for the consumer of the interface that this will happen?

chomick (approved):

src/Platform.Eda.Cli/Commands/ConfigureEda/JsonProcessor/JsonVarsParser.cs
nikhilgirraj: Maybe renaming the method / interface to extract vars would be more appropriate?

src/Platform.Eda.Cli/Commands/ConfigureEda/JsonProcessor/IJsonVarsParser.cs
nikhilgirraj: Yes, I was doing that earlier but I don't know why I stopped. :/

src/Platform.Eda.Cli/Commands/ConfigureEda/JsonProcessor/JsonVarsParser.cs
nikhilgirraj: Would it be better to check against a finite list of valid env names?

chomick: I think we should have a list of valid names - that may be used in many places, actually, so should be shared.

src/Platform.Eda.Cli/Commands/ConfigureEda/JsonProcessor/JsonVarsParser.cs
chomick: I just wonder why we have to modify the input? You can always, outside of this class, extract the subscriber section and use it for further processing - no need to do it here.

tabman83 (approved):

src/Platform.Eda.Cli/Commands/ConfigureEda/JsonProcessor/JsonVarsParser.cs
```diff
@@ -0,0 +1,42 @@
...
+    {
+        public Dictionary<string, JToken> GetFileVars(JObject fileContent, string environmentName)
+        {
+            if (!fileContent.ContainsKey("vars")) 
+                return new Dictionary<string, JToken>(); // no vars
+            
+            var varsDict = JsonConvert.DeserializeObject<Dictionary<string, Dictionary<string, JToken>>>(fileContent["vars"]!.ToString());
+            var outputVarsDict = new Dictionary<string, JToken>();
+
+            foreach (var (propertyKey, innerDict) in varsDict)
```
tabman83, line 19: If this (or the rest of the process) breaks for any reason it will blow up. I would leverage OperationResult to report back the error.

tabman83 (changes requested):

src/Platform.Eda.Cli/Commands/ConfigureEda/JsonProcessor/JsonVarsExtractor.cs
```diff
@@ -0,0 +1,53 @@
...
+            if (varsJObject == null)
+                return new Dictionary<string, JToken>(); // no vars
+
+
+            if (!ConfigureEdaConstants.EnvironmentNames.Contains(environmentName))
+                return new CliExecutionError($"Cannot extract vars for environment {environmentName}.");
+
+            var varsDict = varsJObject.ToObject<Dictionary<string, Dictionary<string, JToken>>>();
+
+            if (varsDict == null)
```
tabman83, line 22: If this section does not fit into your model it will blow up. You should catch it and leverage OperationResult to return it.

tabman83 (changes requested):

src/Platform.Eda.Cli/Commands/ConfigureEda/JsonProcessor/JsonVarsExtractor.cs
nikhilgirraj: oh, now I understand. I did not get it the last time.

nikhilgirraj: Fixed

tabman83 (approved):

PR #312, opened 2020-09-30
#88014 Apply fix for reader crash

tabman83:
There seem to be a race condition when the Director Service attempts to create more than one reader at the same time. The issue is when 2 or more reader service try to re-create the same topic and subscription at the same time, then SB gives out.

The proposed fix just swallows the exception which is only due to the race condition by assuming that the topic and subscription will be retrieved on the next step.

This does not want to be the best solution possible but at least a way to move forward at this stage. I will take the time to investigate separately a better solution to this issue.

### DIFF
--- a/src/CaptainHook.Common/ServiceBus/ServiceBusManager.cs
+++ b/src/CaptainHook.Common/ServiceBus/ServiceBusManager.cs
@@ -63,12 +63,16 @@ namespace CaptainHook.Common.ServiceBus
             var subscription = subscriptionsList.SingleOrDefault(s => string.Equals(s.Name, subscriptionName, StringComparison.OrdinalIgnoreCase));
             if (subscription != null) return subscription;
 
-            await topic.Subscriptions
-                .Define(subscriptionName.ToLowerInvariant())
-                .WithMessageLockDurationInSeconds(60)
-                .WithExpiredMessageMovedToDeadLetterSubscription()
-                .WithMessageMovedToDeadLetterSubscriptionOnMaxDeliveryCount(10)
-                .CreateAsync(cancellationToken);
+            try
+            {
+                await topic.Subscriptions
+                    .Define(subscriptionName.ToLowerInvariant())
+                    .WithMessageLockDurationInSeconds(60)
+                    .WithExpiredMessageMovedToDeadLetterSubscription()
+                    .WithMessageMovedToDeadLetterSubscriptionOnMaxDeliveryCount(10)
+                    .CreateAsync(cancellationToken);
+            }
+            catch { }
 
             await topic.RefreshAsync(cancellationToken);
             subscriptionsList = await topic.Subscriptions.ListAsync(cancellationToken: cancellationToken);
@@ -90,10 +94,14 @@ namespace CaptainHook.Common.ServiceBus
 
             if (topic != null) return topic;
 
-            await serviceBusNamespace.Topics
-                .Define(topicName.ToLowerInvariant())
-                .WithDuplicateMessageDetection(TimeSpan.FromMinutes(10))
-                .CreateAsync(cancellationToken);
+            try
+            {
+                await serviceBusNamespace.Topics
+                    .Define(topicName.ToLowerInvariant())
+                    .WithDuplicateMessageDetection(TimeSpan.FromMinutes(10))
+                    .CreateAsync(cancellationToken);
+            }
+            catch { }
 
             return await _findTopicPolicy.ExecuteAsync(ct => FindTopicAsync(serviceBusNamespace, topicName, ct), cancellationToken);
         }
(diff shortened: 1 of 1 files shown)


src/CaptainHook.Common/ServiceBus/ServiceBusManager.cs
```diff
@@ -63,12 +63,16 @@ public async Task DeleteSubscriptionAsync(string topicName, string subscriptionN
...
+            try
+            {
+                await topic.Subscriptions
+                    .Define(subscriptionName.ToLowerInvariant())
+                    .WithMessageLockDurationInSeconds(60)
+                    .WithExpiredMessageMovedToDeadLetterSubscription()
+                    .WithMessageMovedToDeadLetterSubscriptionOnMaxDeliveryCount(10)
+                    .CreateAsync(cancellationToken);
+            }
+            catch { }
```
chomick, line 75: I think we should at least log it here. What do you think?

src/CaptainHook.Common/ServiceBus/ServiceBusManager.cs
```diff
@@ -90,10 +94,14 @@ private async Task<ITopic> CreateTopicIfNotExistsAsync(string topicName, Cancell
...
-                .WithDuplicateMessageDetection(TimeSpan.FromMinutes(10))
-                .CreateAsync(cancellationToken);
+            try
+            {
+                await serviceBusNamespace.Topics
+                    .Define(topicName.ToLowerInvariant())
+                    .WithDuplicateMessageDetection(TimeSpan.FromMinutes(10))
+                    .CreateAsync(cancellationToken);
+            }
+            catch { }
```
chomick, line 104: Same here

src/CaptainHook.Common/ServiceBus/ServiceBusManager.cs
```diff
@@ -63,12 +63,16 @@ public async Task DeleteSubscriptionAsync(string topicName, string subscriptionN
...
+                    .Define(subscriptionName.ToLowerInvariant())
+                    .WithMessageLockDurationInSeconds(60)
+                    .WithExpiredMessageMovedToDeadLetterSubscription()
+                    .WithMessageMovedToDeadLetterSubscriptionOnMaxDeliveryCount(10)
+                    .CreateAsync(cancellationToken);
+            }
+            catch { }
 
             await topic.RefreshAsync(cancellationToken);
             subscriptionsList = await topic.Subscriptions.ListAsync(cancellationToken: cancellationToken);
```
chomick, line 78: btw - is there any reason we retrieve the full list of all topics to pick one? Can we use GetByName() instead? It is case-sensitive (not like our search) but I wonder if this is a problem for us (totally unrelated to this PR, though! :) )

tabman83: There is GetByName which is case sensitive but since we drive the naming this should not be an issue

src/CaptainHook.Common/ServiceBus/ServiceBusManager.cs
tabman83: Done

src/CaptainHook.Common/ServiceBus/ServiceBusManager.cs
tabman83: Done

src/CaptainHook.Common/ServiceBus/ServiceBusManager.cs
tabman83: Separate PR?

dawidkeshop (approved):

src/CaptainHook.Common/ServiceBus/ServiceBusManager.cs
chomick: Oh yes, that would be something separate.

src/CaptainHook.Common/ServiceBus/ServiceBusManager.cs
```diff
@@ -63,12 +69,19 @@ public async Task DeleteSubscriptionAsync(string topicName, string subscriptionN
...
-            await topic.Subscriptions
-                .Define(subscriptionName.ToLowerInvariant())
-                .WithMessageLockDurationInSeconds(60)
-                .WithExpiredMessageMovedToDeadLetterSubscription()
-                .WithMessageMovedToDeadLetterSubscriptionOnMaxDeliveryCount(10)
-                .CreateAsync(cancellationToken);
+            try
+            {
+                await topic.Subscriptions
+                    .Define(subscriptionName.ToLowerInvariant())
```
chomick, line 75: Do you know, Nino, which SB entity was causing 429 exception for you in your tests?
Is this exception logged in AI?
Are we certain that when that exception happens for one of the call (DLQ or normal subscriber), both will be properly created in the service bus (e.g. subscriptions)?

tabman83: We know that it will be created in the end because outside the try-catch we still try and list the subscriptions to find the relevant one. if it then goes ahead there. it's just a race condition due to both readers being spawned at the same time. should anything else wrong happen there, there still is an outer block where we catch the exception and try to recover the SB connection (it was already there).

As for SB entity, I could see locally that happens for both topic and subscription.

chomick (approved):

PR #313, opened 2020-10-01
CLI accepts params replacements

dawidkeshop:
There is no way to have a profit from this code currently, but as for now it can accept, validate and parse parameters from command line.

The syntax is: `configure-eda -i "C:/Temp" -p evo-host=test.eshopworld.com -p sample=value`

### DIFF
--- a/src/Tests/Platform.Eda.Cli.Tests/Commands/ConfigureEda/OptionsValidation/ReplacementParamsDictionaryAttributeTests.cs
+++ b/src/Tests/Platform.Eda.Cli.Tests/Commands/ConfigureEda/OptionsValidation/ReplacementParamsDictionaryAttributeTests.cs
@@ -8,10 +8,12 @@ namespace Platform.Eda.Cli.Tests.Commands.ConfigureEda.OptionsValidation
 {
     public class ReplacementParamsDictionaryAttributeTests
     {
-        [Fact, IsUnit]
-        public void ForValidParam_ReturnsTrue()
+        [Theory, IsUnit]
+        [InlineData("abc=def")]
+        [InlineData("my-domain=\"test.domain.com\"")]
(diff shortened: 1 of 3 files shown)


src/Tests/Platform.Eda.Cli.Tests/Commands/ConfigureEda/OptionsValidation/ReplacementParamsDictionaryAttributeTests.cs
```diff
@@ -0,0 +1,30 @@
...
+        [InlineData("=def")]
+        [InlineData("abc")]
+        public void ForInvalidParam_ReturnsError(string invalidValue)
+        {
+            var result = new ReplacementParamsDictionaryAttribute().GetValidationResult(invalidValue, new ValidationContext(new object()));
+
+            result.Should().NotBe(ValidationResult.Success);
+        }
+    }
+}
```
tabman83, line 30: What happens if we have blanks like `abc =def` or commas `evo-domain="test.domain.com"`

dawidkeshop: The first one is invalid, the second one is valid. I've updated regex and tests.

tabman83 (approved):
Happy to approve with just a suggestion. Should we just not trim the spaces? Not saying that it is the correct behavior but let's think about that.

dawidkeshop:
> 
> 
> Happy to approve with just a suggestion. Should we just not trim the spaces? Not saying that it is the correct behavior but let's think about that.

Actually it won't be an issue. If CLI is executed with arguments containing spaces, like that:
 `configure-eda -i "C:/Temp" -p evo-host= test.eshopworld.com` it will just fail due to unrecognized syntax.

PR #314, opened 2020-10-05
#87849 JSON replacement service

tabman83:
This implements a JSON replacement service which can be used for both vars and params.

Given the following `Dictionary<string, JToken>`:
```
{
	{ "{vars:sts-settings}", JToken.Parse("{ \"type\": \"OIDC\" } ") },
	{ "{params:evo-host}", (JToken)"aaa" },
}
```

the service is capable of doing the proper replacements. Please see tests attached.


src/Platform.Eda.Cli/Commands/ConfigureEda/JsonProcessor/IJsonReplacer.cs
```diff
@@ -0,0 +1,17 @@
...
+{
+    public interface IJsonReplacer
+    {
+        /// <summary>
+        /// Replace a dictionary of keys into the source JSON object
+        /// </summary>
+        /// <param name="source">The source JSON object</param>
+        /// <param name="replacements">A dictionary of replacements</param>
+        /// <returns></returns>
+        public OperationResult<JObject> Replace(JObject source, Dictionary<string, JToken> replacements);
```
chomick, line 15: Do we use OperationResult in other places for command processing?
If not, I would be happy to throw specific exceptions to handle error cases

src/Platform.Eda.Cli/Commands/ConfigureEda/JsonProcessor/JsonReplacer.cs
```diff
@@ -0,0 +1,75 @@
...
+            }
+        }
+
+        private IEnumerable<JToken> GetTokens(JObject obj)
+        {
+            var toSearch = new Stack<JToken>(obj.Children());
+            while (toSearch.Count > 0)
+            {
+                var inspected = toSearch.Pop();
+                yield return inspected;
```
chomick, line 67: If it is a complex object (i.e. not a string), you will not have to return it and can simplify the processing above.

src/Platform.Eda.Cli/Commands/ConfigureEda/JsonProcessor/JsonReplacer.cs
```diff
@@ -0,0 +1,75 @@
...
+
+        private void ReplaceProperties(JToken token, Dictionary<string, JToken> replacements)
+        {
+            if (token.Type == JTokenType.Property && 
+                ((JProperty)token).Value.Type == JTokenType.String)
+            {
+                var property = (JProperty)token;
+
+                var value = (string)property.Value;
+                foreach (var replacement in replacements)
```
chomick, line 38: With this implementation, it will be difficult to introduce variables validation (or at least, you will have to duplicate the efforts).
A different implementation would be that for every token you scan for any "{var:" string with matching closing bracket and a name inside and you look for a variable in your dictionary. You will be able to handle properly any errors related to missing declarations etc. Also, you will be able to handle case -insensitive search, which I would expect to be preferred.

src/Platform.Eda.Cli/Commands/ConfigureEda/JsonProcessor/JsonReplacer.cs
```diff
@@ -0,0 +1,75 @@
...
+                    {
+                        // replaced variable is a string
+                        if (replacement.Value.Type == JTokenType.String)
+                        {
+                            var replacedValue = value.Replace(replacement.Key, 
+                                $"{(string)replacement.Value}");
+
+                            property.Value = (JToken)replacedValue;
+                        }
+
```
chomick, line 50: else?

src/Tests/Platform.Eda.Cli.Tests/Commands/ConfigureEda/JsonReplacerTests.cs
```diff
@@ -0,0 +1,134 @@
+﻿using Eshopworld.Tests.Core;
+using FluentAssertions;
+using FluentAssertions.Execution;
+using Newtonsoft.Json.Linq;
+using Platform.Eda.Cli.Commands.ConfigureEda.JsonProcessor;
+using System.Collections.Generic;
+using Xunit;
+
+namespace Platform.Eda.Cli.Tests.Commands.ConfigureEda
```
chomick, line 9: I would like to see more tests, not following subscriber's schema, to ensure different combinations of structures/values/vars/params are handled properly.

src/Platform.Eda.Cli/Commands/ConfigureEda/JsonProcessor/JsonReplacer.cs
tabman83: Could you please elaborate more on that?

tabman83:
Closing as not needed anymore.

PR #315, opened 2020-10-05
Task 87849 replace vars in template

nikhilgirraj:
some parts co-authored by @dawidkeshop,
includes @tabman83's tests from the other #314


src/Platform.Eda.Cli/Commands/ConfigureEda/JsonProcessor/JsonTemplateValuesValuesReplacer.cs
```diff
@@ -0,0 +1,53 @@
...
+using System.Linq;
+using System.Text;
+using System.Text.RegularExpressions;
+using CaptainHook.Domain.Results;
+using Newtonsoft.Json.Linq;
+using Platform.Eda.Cli.Common;
+
+namespace Platform.Eda.Cli.Commands.ConfigureEda.JsonProcessor
+{
+    public class JsonTemplateValuesValuesReplacer : IJsonTemplateValuesReplacer
```
dawidkeshop, line 11: Duplicated `Values` in the name.

dawidkeshop, line 11: Duplicated Values in the name.

src/Tests/Platform.Eda.Cli.Tests/Commands/ConfigureEda/Processor/JsonTemplateValuesValuesReplacerParams.cs
```diff
@@ -0,0 +1,104 @@
...
+using FluentAssertions;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+using Platform.Eda.Cli.Commands.ConfigureEda.JsonProcessor;
+using Platform.Eda.Cli.Commands.ConfigureEda.Models;
+using Xunit;
+
+namespace Platform.Eda.Cli.Tests.Commands.ConfigureEda.Processor
+{
+    public class JsonTemplateValuesValuesReplacerParams
```
dawidkeshop, line 13: Duplicated Values in the name, also name should end with Tests.

src/Tests/Platform.Eda.Cli.Tests/Commands/ConfigureEda/Processor/JsonTemplateValuesValuesReplacerVars.cs
```diff
@@ -0,0 +1,306 @@
...
+using FluentAssertions.Execution;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+using Platform.Eda.Cli.Commands.ConfigureEda.JsonProcessor;
+using Platform.Eda.Cli.Commands.ConfigureEda.Models;
+using Xunit;
+
+namespace Platform.Eda.Cli.Tests.Commands.ConfigureEda.Processor
+{
+    public class JsonTemplateValuesValuesReplacerVars
```
dawidkeshop, line 14: Duplicated Values in the name, also name should end with Tests.

dawidkeshop (commented):
Couple naming suggestions

src/Platform.Eda.Cli/Commands/ConfigureEda/JsonProcessor/JsonTemplateValuesValuesReplacer.cs
nikhilgirraj: :|

src/Platform.Eda.Cli/Commands/ConfigureEda/JsonProcessor/JsonTemplateValuesReplacer.cs
```diff
@@ -0,0 +1,53 @@
...
+        };
+
+        public OperationResult<string> Replace(TemplateReplacementType replacementType, string fileContent, Dictionary<string, JToken> variablesDictionary)
+        {
+            var sb = new StringBuilder(fileContent);
+
+            var replacementPrefix = ReplacementTypeToPrefix[replacementType];
+
+            var regexPattern = $@"{{{replacementPrefix}:[a-zA-Z]+[a-zA-Z0-9-_]+}}";
+            var vars = Regex.Matches(fileContent, regexPattern);
```
tabman83, line 26: You can compile the 2 regex beforehand and gather a small improvement. Not a biggie.

src/Platform.Eda.Cli/Commands/ConfigureEda/JsonProcessor/JsonTemplateValuesReplacer.cs
```diff
@@ -0,0 +1,53 @@
...
+                var value = tempDictionary[variableMatch.Value];
+
+                if (value.Type == JTokenType.String)
+                {
+                    sb = sb.Replace(variableMatch.Value, value.ToString());
+                }
+                else if (value.Type == JTokenType.Object)
+                {
+                    sb = sb.Replace($"\"{variableMatch.Value}\"", value.ToString());
+                }
```
tabman83, line 47: Let's make it case insensitive here

src/Platform.Eda.Cli/Commands/ConfigureEda/JsonProcessor/IJsonTemplateValuesReplacer.cs
```diff
@@ -0,0 +1,11 @@
+﻿using System.Collections.Generic;
+using CaptainHook.Domain.Results;
+using Newtonsoft.Json.Linq;
+
+namespace Platform.Eda.Cli.Commands.ConfigureEda.JsonProcessor
+{
+    public interface IJsonTemplateValuesReplacer
+    {
+        public OperationResult<string> Replace(TemplateReplacementType replacementType, string fileContent, Dictionary<string, JToken> variablesDictionary);
```
chomick, line 9: Summary comment for the interface & method, please.

src/Platform.Eda.Cli/Commands/ConfigureEda/JsonProcessor/JsonTemplateValuesReplacer.cs
```diff
@@ -0,0 +1,53 @@
...
+            {TemplateReplacementType.Vars, "vars"}
+        };
+
+        public OperationResult<string> Replace(TemplateReplacementType replacementType, string fileContent, Dictionary<string, JToken> variablesDictionary)
+        {
+            var sb = new StringBuilder(fileContent);
+
+            var replacementPrefix = ReplacementTypeToPrefix[replacementType];
+
+            var regexPattern = $@"{{{replacementPrefix}:[a-zA-Z]+[a-zA-Z0-9-_]+}}";
```
chomick, line 25: You may want to build a compiled version of the regexp so it won't have to be compiled every time you use it

tabman83 (commented):
Again can you have a try-catch block, if anything goes wrong it will just blow up. We should leverage the OperationResult<> return type.

src/Platform.Eda.Cli/Commands/ConfigureEda/JsonProcessor/JsonTemplateValuesReplacer.cs
```diff
@@ -0,0 +1,53 @@
...
+            var vars = Regex.Matches(fileContent, regexPattern);
+
+            var tempDictionary = variablesDictionary.ToDictionary(k => $"{{{replacementPrefix}:{k.Key}}}", k => k.Value);
```
chomick, line 28: If you define your regex differently, you can extract the variable name alone (without the crap around it) and use it in the given dictionary - no need to rebuild it.
Try it out but I think this pattern will do the job:

`{replacementPrefix}:([a-zA-Z]+[a-zA-Z0-9-_]+)}`

this will result in 2nd group to contain your result, I think. 
Also, you can probably replace the var name with `(\w+)`?

src/Platform.Eda.Cli/Commands/ConfigureEda/JsonProcessor/JsonTemplateValuesReplacer.cs
```diff
@@ -0,0 +1,53 @@
...
+            {
+                var value = tempDictionary[variableMatch.Value];
+
+                if (value.Type == JTokenType.String)
+                {
+                    sb = sb.Replace(variableMatch.Value, value.ToString());
+                }
+                else if (value.Type == JTokenType.Object)
+                {
+                    sb = sb.Replace($"\"{variableMatch.Value}\"", value.ToString());
```
chomick, line 46: You actually have no guarantee that your variable in this case was used in the correct way, e.g. you can have:
```
  x: "blah{var:my-object}blah"
```

and `my-object` variable would be defined as an object.
In this case you will simply miss the replacement.

To fix that, you probably want to ensure that `"` is just before and after the variable usage and if not, thrown an error.

src/Tests/Platform.Eda.Cli.Tests/Commands/ConfigureEda/Processor/JsonTemplateValuesReplacerParamsTests.cs
```diff
@@ -0,0 +1,206 @@
...
+using FluentAssertions.Execution;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+using Platform.Eda.Cli.Commands.ConfigureEda.JsonProcessor;
+using Platform.Eda.Cli.Commands.ConfigureEda.Models;
+using Xunit;
+
+namespace Platform.Eda.Cli.Tests.Commands.ConfigureEda.Processor
+{
+    public class JsonTemplateValuesReplacerParamsTests
```
chomick, line 14: As I made the comment on Nino's PR as well, can you ensure we test it with other cases, just like some I gave above? Try to break it and define usage that do not necessarily mirror CLI usage -  this class is generic enough, your tests could be simpler and broader, rather than testing very few, specific cases for subscriber Json.

src/Tests/Platform.Eda.Cli.Tests/Commands/ConfigureEda/Processor/JsonTemplateValuesReplacerVarsTests.cs
```diff
@@ -0,0 +1,212 @@
...
+using FluentAssertions.Execution;
+using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+using Platform.Eda.Cli.Commands.ConfigureEda.JsonProcessor;
+using Platform.Eda.Cli.Commands.ConfigureEda.Models;
+using Xunit;
+
+namespace Platform.Eda.Cli.Tests.Commands.ConfigureEda.Processor
+{
+    public class JsonTemplateValuesReplacerVarsTests
```
chomick, line 14: I would avoid duplicating tests like this - they are almost identical. Again, you want to test generic functionality of your class rather than specific usage in CLI case :) 
Maybe we can ditch the dictionary of enums and allow for string prefix to be passed instead to the method?

src/Platform.Eda.Cli/Commands/ConfigureEda/JsonProcessor/JsonTemplateValuesReplacer.cs
nikhilgirraj: I'm not aware of any ways to make a case-insensitive replacement in `StringBuilder`. Besides, wouldn't the Match obtained from Regex already have the correct case?

src/Platform.Eda.Cli/Commands/ConfigureEda/JsonProcessor/JsonTemplateValuesReplacer.cs
dawidkeshop: The question is, do we need the exact name of unknown variable from the file, like: `{params:some-url}` or `some-url` is enough? I thought the first one is required, so that's the reason for that dictionary conversion.

src/Platform.Eda.Cli/Commands/ConfigureEda/JsonProcessor/JsonTemplateValuesReplacer.cs
```diff
@@ -0,0 +1,73 @@
...
+            try
+            {
+                var sb = new StringBuilder(fileContent);
+
+                var replacementPrefix = ReplacementTypeToPrefix[replacementType];
+                var regexPattern = ReplacementTypeToRegex[replacementType];
+
+                var vars = regexPattern.Matches(fileContent);
+
+                var tempDictionary =
```
chomick, line 37: I thought we can avoid the temp dictionary now that you use regexp groups?

src/Platform.Eda.Cli/Commands/ConfigureEda/JsonProcessor/JsonTemplateValuesReplacer.cs
```diff
@@ -0,0 +1,73 @@
...
+                {
+                    var value = tempDictionary[variableMatch.Value];
+
+                    if (value.Type == JTokenType.String)
+                    {
+                        sb.Replace(variableMatch.Value, value.ToString());
+                    }
+                    else if (value.Type == JTokenType.Object)
+                    {
+                        if (sb[variableMatch.Index - 1] != '"' || sb[variableMatch.Index + variableMatch.Value.Length] != '"')
```
chomick, line 57: Add extra checks for boundaries (variable at start/end of the input string without `"`)

src/Platform.Eda.Cli/Commands/ConfigureEda/JsonProcessor/JsonTemplateValuesReplacer.cs
```diff
@@ -0,0 +1,76 @@
...
+                var unknownVars = vars.Where(x => !tempDictionary.ContainsKey(x.Value)).Select(x => x.Value).ToArray();
+                if (unknownVars.Any())
+                {
+                    return new CliExecutionError(
+                        $"Template has an undeclared {replacementPrefix}: {string.Join(',', unknownVars)}");
+                }
+
+                foreach (var variableMatch in vars.Reverse())
+                {
+                    var value = tempDictionary[variableMatch.Value];
```
chomick, line 49: So your solution is still case sensitive - any chance we can make it case insensitive?

src/Platform.Eda.Cli/Commands/ConfigureEda/JsonProcessor/JsonTemplateValuesReplacer.cs
nikhilgirraj: I'm expecting to return an error in such a scenario.  Do you mean to add a range check to return a more specific error message?

src/Platform.Eda.Cli/Commands/ConfigureEda/JsonProcessor/JsonTemplateValuesReplacer.cs
nikhilgirraj: I can convert the vars dictionary to a case insensitive one - `new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase)`

src/Platform.Eda.Cli/Commands/ConfigureEda/JsonProcessor/JsonTemplateValuesReplacer.cs
```diff
@@ -0,0 +1,64 @@
...
+        private static readonly Regex ValidReplacementPrefix = new Regex("^\\w+$", RegexOptions.Compiled);
+        public OperationResult<string> Replace(string replacementPrefix, string fileContent, Dictionary<string, JToken> variablesDictionary)
+        {
+            try
+            {
+                if (!ValidReplacementPrefix.IsMatch(replacementPrefix))
+                    return new CliExecutionError($"Invalid replacement prefix '{replacementPrefix}'");
+
+                replacementPrefix = replacementPrefix.ToLowerInvariant();
+
```
tabman83, line 23: Can we not just do a case insensitive check

src/Platform.Eda.Cli/Commands/ConfigureEda/JsonProcessor/JsonTemplateValuesReplacer.cs
```diff
@@ -0,0 +1,64 @@
...
+{
+    public class JsonTemplateValuesReplacer : IJsonTemplateValuesReplacer
+    {
+        private static readonly Regex ValidReplacementPrefix = new Regex("^\\w+$", RegexOptions.Compiled);
+        public OperationResult<string> Replace(string replacementPrefix, string fileContent, Dictionary<string, JToken> variablesDictionary)
+        {
+            try
+            {
+                if (!ValidReplacementPrefix.IsMatch(replacementPrefix))
+                    return new CliExecutionError($"Invalid replacement prefix '{replacementPrefix}'");
```
tabman83, line 20: I am a bit unclear here. If the replacementPrefix is either `var` or `params` we can just cache & compile the respective regex.

src/Platform.Eda.Cli/Commands/ConfigureEda/JsonProcessor/JsonTemplateValuesReplacer.cs
```diff
@@ -0,0 +1,64 @@
...
+
+                var sb = new StringBuilder(fileContent);
+                foreach (var variableMatch in vars.Reverse())
+                {
+                    var value = variablesDictionary[variableMatch.Groups[1].Value];
+
+                    if (value.Type == JTokenType.String)
+                    {
+                        sb.Remove(variableMatch.Index, variableMatch.Length);
+                        sb.Insert(variableMatch.Index, value);
```
tabman83, line 43: That's 2 operations. Is it more efficient than just a replace?

tabman83 (approved):

src/Tests/Platform.Eda.Cli.Tests/Commands/ConfigureEda/Processor/JsonTemplateValuesReplacerTests.cs
```diff
@@ -0,0 +1,191 @@
...
+            };
+
+            var source = @"{ ""authentication"": ""{vars:sts-settings}"" }";
+
+            // Act
+            var result = _jsonVarsValuesReplacer.Replace("vars", source, replacements);
+
+            // Assert
+            result.Should().BeEquivalentTo(
+                    new OperationResult<string>($@"{{ ""authentication"": {JToken.Parse("{ \"type\": \"OIDC\" } ")} }}"));
```
mszymeshop, line 37: This is impossible to read for a human :) Can you do the ultimate final string here as the expected result? No concatenation and no interpolation?

src/Tests/Platform.Eda.Cli.Tests/Commands/ConfigureEda/Processor/JsonTemplateValuesReplacerTests.cs
```diff
@@ -0,0 +1,191 @@
...
+                ["val-var1"] = "blah.bla",
+                ["obj-var1"] = objValue
+            };
+
+            var result = _jsonVarsValuesReplacer.Replace("vars", SimpleJsonWithVars, varsDictionary);
+
+            result.Should().BeEquivalentTo(new OperationResult<string>($@"
+{{  
+    ""eventName"": ""event"",
+    ""objectType"": {objValue},
```
mszymeshop, line 74: I think it would make sense not to use string interpolation here

src/Tests/Platform.Eda.Cli.Tests/Commands/ConfigureEda/Processor/JsonTemplateValuesReplacerTests.cs
```diff
@@ -0,0 +1,191 @@
...
+                ["val-var1"] = "blah.bla",
+            };
+
+            var result = _jsonVarsValuesReplacer.Replace("vars", SimpleJsonWithVars, varsDictionary);
+            var expected = new OperationResult<string>(new CliExecutionError("Template has undeclared vars: obj-var1"));
+            result.Should().BeEquivalentTo(expected);
+        }
+
+        [Fact, IsUnit]
+        public void Replace_UndeclaredVariableDifferentReplacementType_ReturnsValidString()
```
mszymeshop, line 126: Does it return "valid string" or "unchanged string"

src/Tests/Platform.Eda.Cli.Tests/Commands/ConfigureEda/Processor/JsonTemplateValuesReplacerTests.cs
```diff
@@ -0,0 +1,191 @@
...
+        [InlineData("(")]
+        [InlineData(" ")]
+        [InlineData("")]
+        [InlineData(null)]
+        public void Replace_InvalidReplacementTypes_ReturnsError(string replacementPrefix)
+        {
+            var varsDictionary = new Dictionary<string, JToken>();
+            var result = _jsonVarsValuesReplacer.Replace(replacementPrefix, SimpleJsonWithVars, varsDictionary);
+
+            result.IsError.Should().BeTrue();
```
mszymeshop, line 160: What kind of error will be displayed then? Might be worth asserting on the Error property here.

src/Tests/Platform.Eda.Cli.Tests/Commands/ConfigureEda/Processor/JsonTemplateValuesReplacerTests.cs
```diff
@@ -0,0 +1,191 @@
...
+        }
+
+        [Fact, IsUnit]
+        public void Replace_NonJsonTemplate_SuccessfulReplacement()
+        {
+            var result = _jsonVarsValuesReplacer.Replace("param", "Hi, {param:person}!", new Dictionary<string, JToken>{["person"] = "Captain Hook" });
+            result.Should().BeEquivalentTo(new OperationResult<string>("Hi, Captain Hook!"));
```
mszymeshop, line 167: :100:

src/Tests/Platform.Eda.Cli.Tests/Commands/ConfigureEda/Processor/JsonTemplateValuesReplacerTests.cs
```diff
@@ -0,0 +1,191 @@
...
+        }
+
+        [Theory, IsUnit]
+        [InlineData("{param:value")]
+        [InlineData("{param: value")]
+        [InlineData("param:value}")]
+        [InlineData("{param;value}")]
+        public void Replace_MalformedTemplate_NotReplaced(string badTemplate)
+        {
+            var result = _jsonVarsValuesReplacer.Replace("param", $@"{{ ""type"": Hi, {badTemplate} }}", new Dictionary<string, JToken> { ["value"] = "ValidValue" });
```
mszymeshop, line 177: Do we assume all parameters need to have the prefix? This is something I will need to check with validation.

src/Platform.Eda.Cli/Commands/ConfigureEda/JsonProcessor/JsonTemplateValuesReplacer.cs
```diff
@@ -0,0 +1,64 @@
...
+
+                var regexPattern = new Regex($@"{{{Regex.Escape(replacementPrefix)}:([a-zA-Z]+[a-zA-Z0-9-_]+)}}", RegexOptions.Compiled);
+
+                var vars = regexPattern.Matches(fileContent);
+
+                var unknownVars = vars.Where(x => !variablesDictionary.ContainsKey(x.Groups[1].Value)).Select(x => x.Groups[1].Value).ToArray();
+                if (unknownVars.Any())
+                {
+                    return new CliExecutionError(
+                        $"Template has undeclared {replacementPrefix}: {string.Join(',', unknownVars)}");
```
mszymeshop, line 32: consider ", " instead to introduce a clearer message when the list has more than a single element. Might be worth unit testing that case as well (there is a test for a single unknown var only)

src/Platform.Eda.Cli/Commands/ConfigureEda/JsonProcessor/JsonTemplateValuesReplacer.cs
```diff
@@ -0,0 +1,64 @@
...
+        public OperationResult<string> Replace(string replacementPrefix, string fileContent, Dictionary<string, JToken> variablesDictionary)
+        {
+            try
+            {
+                if (!ValidReplacementPrefix.IsMatch(replacementPrefix))
+                    return new CliExecutionError($"Invalid replacement prefix '{replacementPrefix}'");
+
+                replacementPrefix = replacementPrefix.ToLowerInvariant();
+
+                var regexPattern = new Regex($@"{{{Regex.Escape(replacementPrefix)}:([a-zA-Z]+[a-zA-Z0-9-_]+)}}", RegexOptions.Compiled);
```
mszymeshop, line 24: If we're having these match requirements for variable names, we should unit test them

src/Platform.Eda.Cli/Commands/ConfigureEda/JsonProcessor/JsonTemplateValuesReplacer.cs
```diff
@@ -0,0 +1,64 @@
...
+                    var value = variablesDictionary[variableMatch.Groups[1].Value];
+
+                    if (value.Type == JTokenType.String)
+                    {
+                        sb.Remove(variableMatch.Index, variableMatch.Length);
+                        sb.Insert(variableMatch.Index, value);
+                    }
+                    else if (value.Type == JTokenType.Object)
+                    {
+                        if (sb[variableMatch.Index - 1] != '"' || sb[variableMatch.Index + variableMatch.Value.Length] != '"')
```
mszymeshop, line 47: Is it possible for `variableMatch.Index` to be `0`? This would result in an exception.

src/Tests/Platform.Eda.Cli.Tests/Commands/ConfigureEda/Processor/JsonTemplateValuesReplacerTests.cs
nikhilgirraj: The description (https://eshopworld.atlassian.net/wiki/spaces/PLAT/pages/2026602872/Decentralised+EDA+Subscribers+Onboarding) does not say anything about it specifically. I have coded this under the assumption that a prefix is indeed required.

src/Platform.Eda.Cli/Commands/ConfigureEda/JsonProcessor/JsonTemplateValuesReplacer.cs
nikhilgirraj: `.Replace` replaces all the occurrences of the template at once which did not allow checking for quotation marks around each variable usage. Remove and Insert lets me do that when done in reverse order.

src/Platform.Eda.Cli/Commands/ConfigureEda/JsonProcessor/JsonTemplateValuesReplacer.cs
nikhilgirraj: @chomick's suggestion I removed the enum to make the class more generic and testing easier.

src/Platform.Eda.Cli/Commands/ConfigureEda/JsonProcessor/JsonTemplateValuesReplacer.cs
nikhilgirraj: Handled

src/Platform.Eda.Cli/Commands/ConfigureEda/JsonProcessor/JsonTemplateValuesReplacer.cs
nikhilgirraj: Added tests.

nikhilgirraj: Changed the regex as you had suggested. Added tests.

mszymeshop (approved):

PR #319, opened 2020-10-06
Temporarily removed empty directory error.

nikhilgirraj:
Fixes the error in CLI pipeline. The complete code is a WIP on task 88419.

### DIFF
--- a/src/Platform.Eda.Cli/Commands/ConfigureEda/JsonProcessor/SubscribersDirectoryProcessor.cs
+++ b/src/Platform.Eda.Cli/Commands/ConfigureEda/JsonProcessor/SubscribersDirectoryProcessor.cs
@@ -36,10 +36,11 @@ namespace Platform.Eda.Cli.Commands.ConfigureEda.JsonProcessor
                 var subscriberFiles =
                     _fileSystem.Directory.GetFiles(sourceFolderPath, "*.json", SearchOption.AllDirectories);
 
-                if (!subscribers.Any())
-                {
-                    return new CliExecutionError($"No files have been found in '{sourceFolderPath}'");
-                }
+                // TODO: Part of next PR
+                //if (!subscribers.Any())
+                //{
+                //    return new CliExecutionError($"No files have been found in '{sourceFolderPath}'");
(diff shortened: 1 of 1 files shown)


tabman83 (approved):

src/Platform.Eda.Cli/Commands/ConfigureEda/JsonProcessor/SubscribersDirectoryProcessor.cs
```diff
@@ -36,10 +36,11 @@ public OperationResult<IEnumerable<PutSubscriberFile>> ProcessDirectory(string i
                 var subscriberFiles =
                     _fileSystem.Directory.GetFiles(sourceFolderPath, "*.json", SearchOption.AllDirectories);
 
-                if (!subscribers.Any())
-                {
-                    return new CliExecutionError($"No files have been found in '{sourceFolderPath}'");
-                }
+                // TODO: Part of next PR
+                //if (!subscribers.Any())
```
dawidkeshop, line 40: I believe that the problem is that condition uses `subscribers .Any()` instead of `subscribersFiles.Any()`.

mszymeshop: I think it makes sense to address it the way @dawidkeshop suggests

mszymeshop (changes requested):

src/Platform.Eda.Cli/Commands/ConfigureEda/JsonProcessor/SubscribersDirectoryProcessor.cs
nikhilgirraj: Fixing that would not fix the error since the rest of the code is still a WIP and would just transfer the error to the next step - file parsing. I intend to change the return type of the method from `IEnumerable<PutSubscriberRequest>` to `IEnumerable<string> filePaths` in my next PR. So I want to spend less time here.

mszymeshop (approved):

PR #322, opened 2020-10-06
#88477 - validate environments for configure-eda cli

mszymeshop:
For parameter validation the attached approach is suggested. When the validation fails, the command will not be executed at all. For advanced json contents validation, we are going to use Fluent Validation.


src/Platform.Eda.Cli/Commands/ConfigureEda/OptionsValidation/EnvironmentValidationAttribute.cs
```diff
@@ -0,0 +1,28 @@
...
+        {
+            if (value is string s && (AllowedEnvironments.Contains(s, StringComparer.Ordinal) || string.IsNullOrEmpty(s)))
+            {
+                return ValidationResult.Success;
+                
+            }
+
+            var message = string.Format(ErrorMessageString, value);
+            return new ValidationResult(message);
+        }
```
tabman83, line 26: do we really need to be case-sensitive here?

tabman83 (approved):
Once this is merged, the CLI will break because it is being invoked with -env=VERIFY.

Would you please address this in the platform-config-events YAML file, I suppose we can just remove the env param and let it take the default there.

tabman83:
> Once this is merged, the CLI will break because it is being invoked with -env=VERIFY.
> 
> Would you please address this in the platform-config-events YAML file, I suppose we can just remove the env param and let it take the default there.

Nevermind, there already is a PR for that.

dawidkeshop (approved):

src/Platform.Eda.Cli/Commands/ConfigureEda/OptionsValidation/EnvironmentValidationAttribute.cs
mszymeshop: The main reason I require upper case is that they already come upper case from the pipeline and the config files have the env in upper case, so kept it synchronised that way. Saved me time to check how it would all behave for the other case.

PR #323, opened 2020-10-07
#88561 Sequence diagram of startup flow

tabman83:
[Updated Preview Here](http://www.plantuml.com/plantuml/png/ZP9FpzCm3CRl-HJMxaZZpylXFQ2Nbmc1k20ckBkDrqMaSUMa6_lsE5swT80gTkZ1psC_FtjvYWabZRqt3O_1ebnWJt9oBJsF4LWi2RpT6bL_mUO33WbTW4_CFpS6e6NlgKsEGuHlXEej7GzRLGSTx5etO4YmSw96dlKtV-hlltukqki7Gy5gdAb4N-Zo0qUVPilMM4pOO2HbujXpXDtpfgYlb1QdhE_n1lGo5tswbfAuhYCXQ-1QUF5k0dZKXSB1TI1q48f7bRMiQiLVM3oZ1KDU8VOV0OT1QzCnjIrm8k0JoQApj9G1UKbZp2nhIDqoMwjnQLbkU-yo_upiIJgwMJpp0EWznB69hRY69CuYW07LKjslrs6A2Fgh2CfG_kX7iBgky0LaAcJZhAbj-OggjGGF-SdaqVbwSs9xnD1f8i1APgjXYKead5qwGg1p6V7ViAMJVJtVFGjQkxxxEiVAAKhdtvS8cfS_J_ePU_yR)


docs/CS-CH-SD-startup.uml
```diff
@@ -0,0 +1,29 @@
...
+
+box "Captain Hook"
+  collections Reader order 50
+  participant Director order 40
+endbox
+
+boundary KeyVault order 30
+database "Cosmos DB" order 20
+
+KeyVault -> Director : read config
```
mszymeshop, line 12: The vector should be the other way round, Directors reads from KeyVault and CosmosDb

docs/CS-CH-SD-startup.uml
```diff
@@ -0,0 +1,29 @@
...
+
+loop for all subscribers config
+
+  Director --> Reader : destroy old or changed reader
+  Reader --> ServiceBus : remove subscription
+
+  Director --> Reader : create new reader
+  Reader --> ServiceBus : create subscription
+
+  Director --> Reader : create new DLQ reader
```
mszymeshop, line 25: DLQ is just a special case, the loop goes over subscribers and DLQ is a subscriber

docs/CS-CH-SD-startup.uml
```diff
@@ -0,0 +1,29 @@
...
+
+KeyVault -> Director : read config
+"Cosmos DB" -> Director : read config
+
+note over Director : merge config
+
+loop for all subscribers config
+
+  Director --> Reader : destroy old or changed reader
+  Reader --> ServiceBus : remove subscription
```
mszymeshop, line 20: The removal of a subscription is not part of the main flow, is it

docs/CS-CH-SD-startup.uml
```diff
@@ -0,0 +1,29 @@
...
+database "Cosmos DB" order 20
+
+KeyVault -> Director : read config
+"Cosmos DB" -> Director : read config
+
+note over Director : merge config
+
+loop for all subscribers config
+
+  Director --> Reader : destroy old or changed reader
```
mszymeshop, line 19: This suggests that destroy and creation always take place. They do, but only if config has changed. Maybe it is worth identifying that.

docs/CS-CH-SD-startup.uml
tabman83: Yes changed

docs/CS-CH-SD-startup.uml
tabman83: Yes changed

docs/CS-CH-SD-startup.uml
tabman83: I think it is in the sense that is done during the startup which is what I am showing here

docs/CS-CH-SD-startup.uml
tabman83: Yes changed to better highlight what is happening

docs/CS-CH-SD-startup.uml
```diff
@@ -0,0 +1,40 @@
...
+loop all subscribers config
+  
+  alt config removed
+
+    Director --> Reader : destroy reader
+    Reader --> ServiceBus : create subscription
+
+  else config changed
+  
+    Director --> Reader : create reader with new config
```
chomick, line 26: Probably you may want to swap in places the reader flows (create vs destroy)

docs/CS-CH-SD-startup.uml
```diff
@@ -0,0 +1,40 @@
...
+boundary ServiceBus order 60
+
+box "Captain Hook"
+  collections Reader order 50
+  participant Director order 40
+endbox
+
+boundary KeyVault order 30
+database "Cosmos DB" order 20
+
```
chomick, line 11: Can you add some start-point, e.g. app-startup or config re-load? Looking at the diagram was hard to say where the thing starts 
The object that starts the reload/start-up could be SF if you need something

chomick (approved):

docs/CS-CH-SD-startup.uml
tabman83: of course :)

docs/CS-CH-SD-startup.uml
tabman83: changed as well to highlight both the app start reload and user triggered refresh

mszymeshop (approved):

PR #328, opened 2020-10-08
Task 88613 validate structure of files

mszymeshop:
This is my approach to json file structure validation. I have decided to use what's available (JObject) and FluentAssertions to validate it. This gave me more control over what I can validate - for instance, if I deserialised the object I wouldn't be able to catch string vs object validation etc. Json Schema validation would on the other hand require translation to the validation errors returned by Fluent Validation framework.

Here is an example of what it outputs:
image


dawidkeshop (approved):

dawidkeshop:
Please remember to Squash commits, otherwise we will have a big mess in `master`.

PR #332, opened 2020-10-08
Task 89143 CLI output messages

nikhilgirraj:
Changes to CLI output + Tests

Sample -
image


src/Platform.Eda.Cli/Commands/ConfigureEda/JsonProcessor/PutSubscriberProcessChain.cs
```diff
@@ -67,6 +66,9 @@ public async Task<int> ProcessAsync(string inputFolderPath, string env, Dictiona
                 _console.WriteNormal($"File '{Path.GetRelativePath(inputFolderPath, subscriberFilePath)}' has been found");
 
                 // Step 1 - Read file
+                var fileRelativePath = Path.GetRelativePath(inputFolderPath, subscriberFilePath);
+                _console.WriteNormal($"Processing file: '{fileRelativePath}'");
```
mszymeshop, line 70: Isn't the above information (line 66) enough?

mszymeshop: even if, the same value is calculated twice

src/Tests/Platform.Eda.Cli.Tests/Commands/ConfigureEda/Processor/PutSubscriberProcessChainTests.cs
```diff
@@ -157,11 +160,127 @@ async IAsyncEnumerable<ApiOperationResult> EmptyAsyncEnumerable()
...
+            VerifyWrite(_outWriter, "Processing file: '..\\file1.json'");
+            VerifyWrite(_outWriter, "Processing file: '..\\file2.json'");
+            VerifyWrite(_outWriter, "Extracting variables", Times.Exactly(2));
+            VerifyWrite(_outWriter, "Replacing vars in template", Times.Exactly(2));
+            VerifyWrite(_outWriter, "Replacing params in template", Times.Exactly(2));
+            VerifyWrite(_outWriter, "File successfully parsed", Times.Exactly(2));
+            VerifyWrite(_outWriter, "By default the CLI runs in 'dry-run' mode. If you want to run the configuration against Captain Hook API use the '--no-dry-run' switch");
+        }
+
+        private static void VerifyWrite(Mock<TextWriter> mockWriter, string text, Times times)
```
dawidkeshop, line 276: That's very nice. I think would be good to make this method an extension method and use in all other tests which checks what is write on console.
What do you think @tabman83 @mszymeshop  ?

dawidkeshop (approved):

mszymeshop (approved):

src/Platform.Eda.Cli/Commands/ConfigureEda/JsonProcessor/PutSubscriberProcessChain.cs
nikhilgirraj: Line 66 should be sufficient. But somehow that was not part of my first commit to the branch and I'm not able to track which commit added that. :/

nikhilgirraj: Removing the one on line 66 since that way I don't have to change the tests.

PR #342, opened 2020-10-09
Task 89423 fix console errors

dawidkeshop:
image

1. Thanks to removing `Environment.NewLine`, now whole box will have same colour in the pipeline (yellow for warning, red for error).
2. Multi value options now displays all values in console (not only the first one as previously).
3. No value options (flags) now displays `SET` or `NOT SET` depending on real usage.


src/Platform.Eda.Cli/Extensions/CommandOptionsListExtensions.cs
```diff
@@ -1,15 +1,33 @@
...
+            string JoinValues() => string.Join(";", option.Values);
+
+            var valueToPrint = option.OptionType switch
+            {
+                CommandOptionType.NoValue => (option.HasValue() ? "is SET" : "is NOT SET"),
+                CommandOptionType.SingleOrNoValue => (option.HasValue() ? $"'{JoinValues()}'" : "is NOT SET"),
+                _ => $"'{JoinValues()}'"
+            };
+
+            return $"--{option.LongName}: {valueToPrint}";
```
tabman83, line 30: I would just say:
- `SET`
- `UNSET`

tabman83 (approved):

PR #343, opened 2020-10-12
Return 1 if any file processing fails.

nikhilgirraj:
plus tests

### DIFF
--- a/src/Tests/Platform.Eda.Cli.Tests/Commands/ConfigureEda/Processor/PutSubscriberProcessChainTests.cs
+++ b/src/Tests/Platform.Eda.Cli.Tests/Commands/ConfigureEda/Processor/PutSubscriberProcessChainTests.cs
@@ -88,18 +88,13 @@ namespace Platform.Eda.Cli.Tests.Commands.ConfigureEda.Processor
             _jsonTemplateValuesReplacerMock = new Mock<IJsonTemplateValuesReplacer>();
             _apiConsumerMock = new Mock<IApiConsumer>();
 
-            IApiConsumer captainHookBuilder(string env)
-            {
-                return _apiConsumerMock.Object;
-            }
-
             _sut = new PutSubscriberProcessChain(
                 mockConsole.Object,
                 _subscribersDirectoryProcessorMock.Object,
                 _subscriberFileParserMock.Object,
                 _jsonVarsExtractorMock.Object,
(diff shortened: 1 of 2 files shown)


src/Tests/Platform.Eda.Cli.Tests/Commands/ConfigureEda/Processor/PutSubscriberProcessChainTests.cs
```diff
@@ -88,18 +88,13 @@ public PutSubscriberProcessChainTests()
...
-            }
-
             _sut = new PutSubscriberProcessChain(
                 mockConsole.Object,
                 _subscribersDirectoryProcessorMock.Object,
                 _subscriberFileParserMock.Object,
                 _jsonVarsExtractorMock.Object,
                 _jsonTemplateValuesReplacerMock.Object,
-                captainHookBuilder);
+                (env)=> _apiConsumerMock.Object);
```
tabman83, line 97: much better and I think you can use the discard here

src/Tests/Platform.Eda.Cli.Tests/Commands/ConfigureEda/Processor/PutSubscriberProcessChainTests.cs
```diff
@@ -222,6 +217,34 @@ public async Task ProcessAsync_NoDryRunValue_ControlsIfApiConsumerIsCalled(bool
             _apiConsumerMock.Verify(x => x.CallApiAsync(It.IsAny<IEnumerable<PutSubscriberFile>>()), Times.Exactly(timesApiConsumerIsCalled));
         }
 
+        [Fact, IsUnit]
+        public async Task ProcessAsync_OnValidationFailure_ReturnsValue1()
```
tabman83, line 221: This naming here very much feels like there's a typo. What about `ReturnsNonZeroValue` (where you check that the result is actually different than 0) or `ReturnsOne` ?

tabman83 (approved):

src/Tests/Platform.Eda.Cli.Tests/Commands/ConfigureEda/Processor/PutSubscriberProcessChainTests.cs
nikhilgirraj: I want to be specific about the exit code so I think ReturnsOne is a better choice

PR #345, opened 2020-10-12
Silently ignore files which don't contain variables for specified environment

dawidkeshop:


src/Platform.Eda.Cli/Commands/ConfigureEda/JsonProcessor/EnvironmentNamesExtractor.cs
```diff
@@ -0,0 +1,41 @@
...
+        public static OperationResult<IEnumerable<string>> FindInVars(JObject varsJObject)
+        {
+            if (varsJObject == null)
+                return new HashSet<string>();
+
+            Dictionary<string, Dictionary<string, JToken>> varsDict;
+            try
+            {
+                varsDict = varsJObject.ToObject<Dictionary<string, Dictionary<string, JToken>>>();
+            }
```
tabman83, line 21: I am starting to think that we should have introduced a domain model for that

src/Platform.Eda.Cli/Commands/ConfigureEda/JsonProcessor/EnvironmentNamesExtractor.cs
```diff
@@ -0,0 +1,41 @@
...
+            catch (Exception e)
+            {
+                return new CliExecutionError($"Cannot parse vars. {e.Message}.");
+            }
+
+            if (varsDict == null)
+                return new CliExecutionError($"Cannot parse vars from {varsJObject}.");
+
+            var environmentNames = varsDict.SelectMany(var => var.Value).SelectMany(env => env.Key.Split(',')).Select(name => name.ToLower()).Distinct().ToList();
```
tabman83, line 30: I would trim other than split, in case someone has the brilliant idea of putting a space between env names

tabman83 (approved):

src/Platform.Eda.Cli/Commands/ConfigureEda/JsonProcessor/EnvironmentNamesExtractor.cs
dawidkeshop: I was thinking about that, but it looks like  a bigger refactoring, and didn't want to spend so much time on that right now.

src/Platform.Eda.Cli/Commands/ConfigureEda/JsonProcessor/EnvironmentNamesExtractor.cs
dawidkeshop: I think we don't support such scenario. If we'd like to support that, we must also change vars validation and vars extraction logic, so it's definitely out of scope of this task.

PR #350, opened 2020-10-14
Bug 89989 fix all

dawidkeshop:


tabman83 (approved):

src/Tests/CaptainHook.Application.Tests/Infrastructure/DirectorServiceProxyTests.cs
```diff
@@ -22,43 +22,19 @@ public async Task When_Success_Then_SubscriberConfigurationIsReturned()
...
                 .ReturnsAsync(ReaderChangeResult.Success);
 
-            var result = await Proxy.CallDirectorServiceAsync(new CreateReader(new SubscriberConfigurationBuilder().Create()));
+            var result = await Proxy.CallDirectorServiceAsync(new UpdateReader(new SubscriberConfigurationBuilder().Create()));
 
             result.IsError.Should().BeFalse();
             result.Data.Should().NotBeNull();
         }
 
-        [Fact, IsUnit]
```
chomick, line 31: Should we have those tests (or some other?) to verify that when 'update' for existing or new is called, correct actions are taken?

src/Tests/CaptainHook.Tests/Director/DirectorServiceTests.cs
```diff
@@ -53,25 +53,12 @@ public async Task CreateReader_When_ReaderDoesNotExist_Then_ReaderIsCreated()
...
 
-            var createRequest = new CreateReader { Subscriber = _defaultSubscriberConfig };
+            var createRequest = new UpdateReader { Subscriber = _defaultSubscriberConfig };
             var result = await _directorService.ApplyReaderChange(createRequest);
 
             result.Should().Be(ReaderChangeResult.Success);
         }
 
-        [Fact, IsUnit]
-        public async Task CreateReader_When_ReaderAlreadyExistsForThatConfiguration_Then_ReaderAlreadyExistIsReturned()
```
chomick, line 63: As above - should we have a test that verifies that scenario - only outcome is different

src/Tests/CaptainHook.Tests/Director/DirectorServiceTests.cs
```diff
@@ -102,18 +89,6 @@ public async Task UpdateReader_When_ReaderExistsInSameVersion_Then_NoChangeNeede
             result.Should().Be(ReaderChangeResult.NoChangeNeeded);
         }
 
-        [Fact, IsUnit]
-        public async Task UpdateReader_When_ReaderDoesNotExist_Then_ReaderDoesNotExistIsReturned()
```
chomick, line 106: As above

chomick (approved):
I would leave the tests and update the outcome - unless those cases were explicitly tested with other tests?

src/Tests/CaptainHook.Application.Tests/Infrastructure/DirectorServiceProxyTests.cs
dawidkeshop: With fixed flow, DirectorService will never return `ReaderChangeResult.ReaderAlreadyExist`, so no point in testing that.

dawidkeshop: With fixed flow, DirectorService will never return `ReaderAlreadyExist` or `ReaderDoesNotExist`. For both these cases it will return `Success`, so no point in keeping these tests. Test `When_Success_Then_SubscriberConfigurationIsReturned` already covers a success scenario.

src/Tests/CaptainHook.Tests/Director/DirectorServiceTests.cs
dawidkeshop: Already covered by `UpdateReader_When_ReaderExistsInSameVersion_Then_NoChangeNeeded`

src/Tests/CaptainHook.Tests/Director/DirectorServiceTests.cs
dawidkeshop: Already covered in `UpdateReader_When_ReaderDoesNotExist_Then_ReaderIsCreated`

src/Tests/CaptainHook.Application.Tests/Infrastructure/DirectorServiceProxyTests.cs
dawidkeshop: Actually `ReaderDoesNotExist` is returned in the delete flow, so I've decided to keep test for that.

PR #351, opened 2020-10-14
CLI output readability

nikhilgirraj:
Formatted the console output to single line per file
+ Indentations and tests.
Ouput -
image
image


src/Platform.Eda.Cli/Extensions/IConsoleExtensions.cs
```diff
@@ -13,19 +12,31 @@ namespace Platform.Eda.Cli.Extensions
...
 
         public static void WriteNormal(this IConsole console, params string[] lines)
         {
-            console.Out.WriteInColor(Colors.Cyan, lines);
+            console.Out.WriteInColor(Colors.Default, lines);
+        }
+
+        public static void WriteNormalWithFileName(this IConsole console, string fileName, params string[] lines)
+        {
+            const string ok = "Ok";
```
tabman83, line 33: You can probably move this along with the other const at the top of the class

src/Platform.Eda.Cli/Extensions/IConsoleExtensions.cs
```diff
@@ -38,11 +49,33 @@ public static void WriteWarning(this IConsole console, params string[] lines)
             console.Out.WriteInColor(Colors.Yellow, lines);
         }
 
+        public static void WriteSkippedWithFileName(this IConsole console, string fileName, params string[] lines)
+        {
+            const string skip = "Skip";
```
tabman83, line 54: Same as above

src/Platform.Eda.Cli/Extensions/IConsoleExtensions.cs
```diff
@@ -38,11 +49,33 @@ public static void WriteWarning(this IConsole console, params string[] lines)
...
+        }
+
         public static void WriteError(this IConsole console, params string[] lines)
         {
             console.Out.WriteInColor(Colors.Red, lines);
         }
 
+        public static void WriteErrorWithFileName(this IConsole console, string fileName, params string[] lines)
+        {
+            const string error = "Error";
```
tabman83, line 70: Same as above

tabman83 (approved):

tabman83:
Would you have a screenshot of the result to attach?

src/Platform.Eda.Cli/Extensions/IConsoleExtensions.cs
```diff
@@ -13,19 +12,31 @@ namespace Platform.Eda.Cli.Extensions
...
             public static readonly Color Reset = new Color("\u001b[0m");
         }
 
         public static void WriteNormal(this IConsole console, params string[] lines)
         {
-            console.Out.WriteInColor(Colors.Cyan, lines);
+            console.Out.WriteInColor(Colors.Default, lines);
+        }
+
+        public static void WriteNormalWithFileName(this IConsole console, string fileName, params string[] lines)
```
dawidkeshop, line 31: I'm not sure if this method belongs here. This class should contain reusable methods which allows to print `string` on console. Building messages should be done somewhere else.

src/Platform.Eda.Cli/Extensions/IConsoleExtensions.cs
```diff
@@ -38,11 +49,33 @@ public static void WriteWarning(this IConsole console, params string[] lines)
             console.Out.WriteInColor(Colors.Yellow, lines);
         }
 
+        public static void WriteSkippedWithFileName(this IConsole console, string fileName, params string[] lines)
```
dawidkeshop, line 52: This method also doesn't belong here.

src/Platform.Eda.Cli/Extensions/IConsoleExtensions.cs
```diff
@@ -38,11 +49,33 @@ public static void WriteWarning(this IConsole console, params string[] lines)
...
+
+            WriteWarning(console, lines);
+        }
+
         public static void WriteError(this IConsole console, params string[] lines)
         {
             console.Out.WriteInColor(Colors.Red, lines);
         }
 
+        public static void WriteErrorWithFileName(this IConsole console, string fileName, params string[] lines)
```
dawidkeshop, line 68: This method also doesn't belong here.

dawidkeshop (changes requested):
I think a bit of refactoring is required.

src/Platform.Eda.Cli/Extensions/IConsoleExtensions.cs
nikhilgirraj: I figured that since we have the `WriteValidationResultWithFileName` method here. These would also make sense, do you have any suggestions on where I could put them?

dawidkeshop: `WriteValidationResult` (not `WithFileName`) is something reusable and already used in two places. Methods you've added aren't reusable, these are strictly related to particular flow. I would move message building logic to the chain class, since context of the message is known there.

PR #353, opened 2020-10-15
Removed the EmptyDirectory error + tests

nikhilgirraj:


src/Platform.Eda.Cli/Commands/ConfigureEda/JsonProcessor/SubscribersDirectoryProcessor.cs
```diff
@@ -30,11 +30,6 @@ public OperationResult<IEnumerable<string>> ProcessDirectory(string inputFolderP
                 var subscriberFiles =
```
tabman83, line 30: You can return straight away without declaring the variable.

A note regarding the catch block below (github doesnt allow me to comment there):
use `e.Message` for the `CliExecutionError` arg instead of `ToString()`.

tabman83 (approved):

src/Platform.Eda.Cli/Commands/ConfigureEda/JsonProcessor/SubscribersDirectoryProcessor.cs
nikhilgirraj: I put the `ToString` sicne it gives a combined Message + StackTrace information. Is it okay to just have the message?

tabman83: I think we aim at the message only otherwise we're gonna pollute to console output with too much information. At the end of the day the message will be enough to identify which piece of the code is originating it.

mszymeshop (approved):

dawidkeshop (approved):

PR #357, opened 2020-10-19
Removed File OK validation output when running against environment

nikhilgirraj:
Prints the File Validation success message when NoDryRun is false.

### DIFF
--- a/src/Platform.Eda.Cli/Commands/ConfigureEda/JsonProcessor/PutSubscriberProcessChain.cs
+++ b/src/Platform.Eda.Cli/Commands/ConfigureEda/JsonProcessor/PutSubscriberProcessChain.cs
@@ -76,7 +76,7 @@ namespace Platform.Eda.Cli.Commands.ConfigureEda.JsonProcessor
                 var parseFileResult = _subscriberFileParser.ParseFile(subscriberFilePath);
                 if (parseFileResult.IsError)
                 {
-                    WriteErrorWithFileName(fileRelativePath, parseFileResult.Error.Message);
+                    WriteErrorProcessingResult(fileRelativePath, parseFileResult.Error.Message);
                     putSubscriberFiles.Add(new PutSubscriberFile
                     {
                         Error = parseFileResult.Error.Message,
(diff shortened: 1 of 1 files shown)


src/Platform.Eda.Cli/Commands/ConfigureEda/JsonProcessor/PutSubscriberProcessChain.cs
```diff
@@ -183,7 +183,11 @@ public async Task<int> ProcessAsync(string inputFolderPath, string env, Dictiona
                 }
 
                 // Step 4 - Create PutSubscriberFile object for further processing
-                WriteNormalWithFileName(fileRelativePath);
+                if (!noDryRun) // output File Ok only on dry run
```
chomick, line 186: So this will only be an 'OK' output for the validation, warning/error are not handled by this line?

chomick (approved):

src/Platform.Eda.Cli/Commands/ConfigureEda/JsonProcessor/PutSubscriberProcessChain.cs
```diff
@@ -183,7 +183,11 @@ public async Task<int> ProcessAsync(string inputFolderPath, string env, Dictiona
...
+                {
+                    WriteNormalWithFileName(fileRelativePath);
```
mszymeshop, line 188: It isn't entirely clear to me what this method does. Is it a valid name? Maybe a screenshot of the output would help.

dawidkeshop: I agree, would be great to have more meaningful names. I believe in this case it should be something like `WriteSuccessProcessingResult` without `FileName` in the name of the method, because it's the name of one of parameters.

src/Platform.Eda.Cli/Commands/ConfigureEda/JsonProcessor/PutSubscriberProcessChain.cs
nikhilgirraj: No, those are separate.

src/Platform.Eda.Cli/Commands/ConfigureEda/JsonProcessor/PutSubscriberProcessChain.cs
nikhilgirraj: Makes sense, I kept the names from when the method was in CLI output extensions class.

PR #361, opened 2020-10-19
#90759 EventReader: use appropriate message lock value

tabman83:


src/CaptainHook.Common/ServiceBus/ServiceBusManager.cs
```diff
@@ -74,7 +75,7 @@ public async Task DeleteSubscriptionAsync(string topicName, string subscriptionN
             {
                 await topic.Subscriptions
                     .Define(subscriptionName.ToLowerInvariant())
-                    .WithMessageLockDurationInSeconds(60)
+                    .WithMessageLockDurationInSeconds(messageLockDurationInSeconds)
```
chomick, line 78: not sure if you have another task for it but the story asks for this time to be updated when reader is refreshed as well

src/CaptainHook.EventReaderService/EventReaderService.cs
```diff
@@ -194,11 +201,25 @@ protected override async Task OnChangeRoleAsync(ReplicaRole newRole, Cancellatio
         /// </summary>
         internal int InFlightMessageCount => HandlerCount - _freeHandlers.Count;
 
+        internal int CalculateMessageLock()
+        {
```
chomick, line 205: is the fact that we select a random settings, correct? what if i define different timeouts and dueations for different endpoibts? the original plan was to have it in subscriber level (webhook classl but if you guys decided to di it on ebdpoint, how will you handle different settings here?

src/CaptainHook.EventReaderService/EventReaderService.cs
```diff
@@ -194,11 +201,25 @@ protected override async Task OnChangeRoleAsync(ReplicaRole newRole, Cancellatio
...
+            var routesDurations = _initData.SubscriberConfiguration.WebhookRequestRules
+                                .FirstOrDefault(x => x.Routes.Any())?
+                                .Routes?
+                                .Select(x => x.RetrySleepDurations);
+                
+            var durations = routesDurations ?? new List<TimeSpan[]> { _initData.SubscriberConfiguration.RetrySleepDurations };
+
+            return durations.Max(d => _messageLockDurationCalculator.CalculateAsSeconds(_initData.SubscriberConfiguration.Timeout, d));
```
chomick, line 213: was the timeout not checked on endpoint level in handler? is that correct value here?

src/CaptainHook.Common/ServiceBus/ServiceBusManager.cs
tabman83: What is a reader refresh? We only have destroy & re-create reader hence why this is placed here.

chomick: But when you create new reader, the subscription is already in place - will it update the lock duration settings?

src/CaptainHook.EventReaderService/EventReaderService.cs
chomick: I can see now that you are getting a list of retry-duration, not a first one from the list - this should be better.

src/CaptainHook.EventReaderService/EventReaderService.cs
chomick: I see now that you use Max() from calculated duration - that should be fine.

src/CaptainHook.Common/ServiceBus/ServiceBusManager.cs
tabman83: Yes thanks. This has now been addressed.

chomick (approved):
Nino, I think we should test it properly on CI or Test before we move to higher env's. Should we attempt to create more e2e tests related to this functionality? Those tests could use API now for setup/tear-down and they can verify (broadly) number of retries, timeouts etc.
It would take a bit of work to write them up but I think this is well spent effort - what do you think?

tabman83:
> Nino, I think we should test it properly on CI or Test before we move to higher env's. Should we attempt to create more e2e tests related to this functionality? Those tests could use API now for setup/tear-down and they can verify (broadly) number of retries, timeouts etc.
> It would take a bit of work to write them up but I think this is well spent effort - what do you think?

I'll devise a plan and share with you

nikhilgirraj (approved):

PR #364, opened 2020-10-21
#91447 Case insensitive deserialization for replacement dictionary

tabman83:
The replacement dictionary is automatically deserialized with the camelCase convention by Cosmos, and transferred to the entity layer. Once we try to add the webhook through the SetHooks method, this fails because the validator cannot rightly find a matching replacement between the token in the URL and the replacement dictionary. This is because the dictionary applies a case-sensitive search by default. I have changed the replacement dictionaries at different levels to be case insensitive and this also required the implementation of a case-insensitive dictionary deserializer for Cosmos.

This is a draft PR to give you an opportunity to look at the fix, will add unit tests shortly.


chomick (commented):
there is no tests to catch the original issue and now prove it is fixed? can you add those, please?

src/CaptainHook.Storage.Cosmos/CaseInsensitiveDictionaryConverter.cs
```diff
@@ -0,0 +1,38 @@
+﻿using Newtonsoft.Json;
+using Newtonsoft.Json.Linq;
+using System;
+using System.Collections.Generic;
+
+namespace CaptainHook.Storage.Cosmos
+{
+    internal class CaseInsensitiveDictionaryConverter : JsonConverter
```
mszymeshop, line 8: I'd love to see some UTs for this

chomick:
i woder, should we minimise the change to the domain scope only? whenever we set the replacement, rather than blibdly copy the vapues, we could apply lower case keys and ensure vars substitution works on lower cases as well. 
If that is still to broad change, simply create the case-invariant dictionary on the domain model alone - why do we care how cosmos reads it?

tabman83:
> i woder, should we minimise the change to the domain scope only? whenever we set the replacement, rather than blibdly copy the vapues, we could apply lower case keys and ensure vars substitution works on lower cases as well.
> If that is still to broad change, simply create the case-invariant dictionary on the domain model alone - why do we care how cosmos reads it?

See here: ([UriTransformEntity.cs](https://github.com/eShopWorld/captain-hook/blob/task-91447-case-sensitive-deserialization/src/CaptainHook.Domain/Entities/UriTransformEntity.cs))
To build an entity object we use the constructor which takes an IDictionary, that comes from Cosmos side, and does not give us the opportunity. to enforce the case-insensitivity, unless, we make a copy inside that constructor, and we copy to a case insensitive dictionary. Opinions? @chomick @mszymeshop

src/Tests/CaptainHook.Domain.Tests/Entities/WebhooksEntityTests.cs
```diff
@@ -0,0 +1,60 @@
...
+                ["order"] = "$.Response.Order"
+            });
+            var entity = new WebhooksEntity(type, "*", new List<EndpointEntity>(), uriTransform);
+            var endpoint = new EndpointEntity("http://url.com/{OrderNumber}", BasicAuthenticationEntity, "POST", "*" );
+
+            // Act
+            var result = entity.SetEndpoint(endpoint);
+
+            // Assert
+            using(new AssertionScope())
```
mszymeshop, line 33: this will work only for the next line, introduce brackets or `using var _ = new AssertionScope();`

src/CaptainHook.Domain/Entities/UriTransformEntity.cs
```diff
@@ -1,18 +1,24 @@
...
         /// A list of replacements
         /// </summary>
         public IDictionary<string, string> Replace { get; }
 
         public UriTransformEntity(IDictionary<string, string> replace)
         {
-            Replace = new ReadOnlyDictionary<string, string>(replace ?? new Dictionary<string, string>());
+            var caseInsensitiveDictionary = new Dictionary<string, string>(
+                replace ?? EmptyReplacementsDictionary,
+                StringComparer.InvariantCultureIgnoreCase);
```
mszymeshop, line 20: I think OrdinalIgnoreCase makes more sense here, your call though

mszymeshop (approved):

src/CaptainHook.Domain/Entities/UriTransformEntity.cs
tabman83: Changed

src/Tests/CaptainHook.Domain.Tests/Entities/WebhooksEntityTests.cs
tabman83: Changed

src/CaptainHook.Domain/Entities/UriTransformEntity.cs
```diff
@@ -1,18 +1,24 @@
...
 namespace CaptainHook.Domain.Entities
 {
     public class UriTransformEntity
     {
+        private static readonly Dictionary<string, string> EmptyReplacementsDictionary = new Dictionary<string, string>();
+
         /// <summary>
         /// A list of replacements
         /// </summary>
         public IDictionary<string, string> Replace { get; }
```
chomick, line 14: We could use IReadOnlyDictionary interface and leave the implementation as we see fit/need - why do we expose IDictionary but implement it with readonly instance only..

src/Tests/CaptainHook.Domain.Tests/Entities/WebhooksEntityTests.cs
```diff
@@ -0,0 +1,60 @@
...
+
+            // Act
+            var result = entity.SetEndpoint(endpoint);
+
+            // Assert
+            using var _ = new AssertionScope();
+            result.IsError.Should().BeTrue();
+            result.Error.Failures.Should().HaveCount(1);
+            result.Error.Failures.First().As<ValidationFailure>().Message.Should()
+                .Be("URI Transform dictionary must contain all the placeholders defined in each URI");
```
chomick, line 37: Do we have a code/type for this error or do we have to check the actual error message?

src/Tests/CaptainHook.Domain.Tests/Entities/WebhooksEntityTests.cs
```diff
@@ -0,0 +1,60 @@
...
+                ["ordernumber"] = "$.Response.Order"
+            });
+            var entity = new WebhooksEntity(type, "*", new List<EndpointEntity>(), uriTransform);
+            var endpoint = new EndpointEntity("http://url.com/{OrderNumber}", BasicAuthenticationEntity, "POST", "*");
+
+            // Act
+            var result = entity.SetEndpoint(endpoint);
+
+            // Assert
+            result.IsError.Should().BeFalse();
```
chomick, line 57: I presume that test would fail before the fix?

src/Tests/CaptainHook.Domain.Tests/Entities/WebhooksEntityTests.cs
```diff
@@ -0,0 +1,60 @@
...
+using CaptainHook.Domain.Results;
+using Eshopworld.Tests.Core;
+using FluentAssertions;
+using FluentAssertions.Execution;
+using Kusto.Cloud.Platform.Utils;
+using Xunit;
+
+namespace CaptainHook.Domain.Tests.Entities
+{
+    public class WebhooksEntityTests
```
chomick, line 13: What about tests in `WebhooksEntityValidatorTests`? Should they not be extended to capture the issue?
Not negating the need for tests on the entity, but maybe this validation should be consistently verified with other validations?

src/Tests/CaptainHook.Domain.Tests/Entities/WebhooksEntityTests.cs
```diff
@@ -0,0 +1,60 @@
...
+            result.IsError.Should().BeTrue();
+            result.Error.Failures.Should().HaveCount(1);
+            result.Error.Failures.First().As<ValidationFailure>().Message.Should()
+                .Be("URI Transform dictionary must contain all the placeholders defined in each URI");
+            result.Error.Message.Should().Be("Invalid request");
+        }
+
+        [IsUnit, Theory]
+        [ClassData(typeof(WebhooksCallbacksDlqHooks))]
+        public void SetEndpoint_When_EndpointWithTokenInReplaceIsAdded_Then_Succeeds(WebhooksEntityType type)
```
chomick, line 43: This case should be added to this test: `Validate_WebhookWithAllTokensInUriTransform_SucceedsValidation` - we never cared for the casing, our tests should be updated (some) to be aware of the casing now and maybe no need for the new ones?

src/Tests/CaptainHook.Domain.Tests/Entities/WebhooksEntityTests.cs
```diff
@@ -0,0 +1,60 @@
...
+            // Arrange
+            var uriTransform = new UriTransformEntity(new Dictionary<string, string>()
+            {
+                ["ordernumber"] = "$.Response.Order"
+            });
+            var entity = new WebhooksEntity(type, "*", new List<EndpointEntity>(), uriTransform);
+            var endpoint = new EndpointEntity("http://url.com/{OrderNumber}", BasicAuthenticationEntity, "POST", "*");
+
+            // Act
+            var result = entity.SetEndpoint(endpoint);
```
chomick, line 54: Is that the only method that we can use to define endpoints on that entity?

chomick (approved):
I would clean up the unit tests - it looks that we focused on testing validations and they should be extended with the new cases now. 
The additional tests for entity methods are def missing but they should (ideally) focus on other aspects rather than repeating validation cases.

src/Tests/CaptainHook.Domain.Tests/Entities/WebhooksEntityTests.cs
tabman83: yes I have checked that

src/Tests/CaptainHook.Domain.Tests/Entities/WebhooksEntityTests.cs
tabman83: No because that was the failure triggered by the case insensitive compare

src/CaptainHook.Domain/Entities/UriTransformEntity.cs
tabman83: Changed as suggested

src/Tests/CaptainHook.Domain.Tests/Entities/WebhooksEntityTests.cs
tabman83: Changed

src/Tests/CaptainHook.Domain.Tests/Entities/WebhooksEntityTests.cs
tabman83: Done

tabman83:
> I would clean up the unit tests - it looks that we focused on testing validations and they should be extended with the new cases now.
> The additional tests for entity methods are def missing but they should (ideally) focus on other aspects rather than repeating validation cases.

will raise separate PR as I want to deploy this fix for now

PR #365, opened 2020-10-22
#91578 - update validation to catch selection-rule issues

mszymeshop:


src/Tests/CaptainHook.Application.Tests/Validators/Dtos/WebhooksDtoValidatorTests.cs
```diff
@@ -0,0 +1,91 @@
...
+using CaptainHook.Application.Validators.Dtos;
+using CaptainHook.Contract;
+using CaptainHook.TestsInfrastructure.Builders;
+using Eshopworld.Tests.Core;
+using FluentValidation.TestHelper;
+using Xunit;
+
+namespace CaptainHook.Application.Tests.Validators.Dtos
+{
+    public class WebhooksDtoValidatorTests
```
chomick, line 11: How is it possible we did not have a single test for this validator?

chomick: Is it possible these are tested elsewhere?

src/Tests/CaptainHook.Application.Tests/Validators/Dtos/WebhooksDtoValidatorTests.cs
```diff
@@ -0,0 +1,91 @@
...
+            result.ShouldHaveValidationErrorFor(x => x.SelectionRule)
+                .WithErrorMessage("'Selection Rule' must not be empty.");
+        }
+
+        [Fact, IsUnit]
+        public void Validate_UriTransformPresent_NoErrorOnSelectionRuleExpected()
+        {
+            // Arrange
+            var webhooksDto = new WebhooksDtoBuilder()
+                .With(x => x.SelectionRule, "$.Test")
```
chomick, line 80: Could we default it to "$" and not require it in the input?

chomick (approved):

src/Tests/CaptainHook.Application.Tests/Validators/Dtos/WebhooksDtoValidatorTests.cs
mszymeshop: I am afraid I couldn't find any other tests for this class. Looks like it has been missed.

src/Tests/CaptainHook.Application.Tests/Validators/Dtos/WebhooksDtoValidatorTests.cs
mszymeshop: If you mean the regular DTO, then this is something we could check as it could work. At the moment I cannot guarantee this isn't going to break for some flow, so some tests would be advised.

PR #366, opened 2020-10-23
#91761 - allow routes not to specify default selector

mszymeshop:
We are introducing a change with this PR which will:

1. stop requiring exactly one default selector * in RouteAndReplace
1. send a UnroutableEvent to AI if a message comes in which doesn't match the configuration (and we don't have * to fall back)


chomick (approved):

tabman83 (approved):
💯

PR #367, opened 2020-10-26
#91261 #91264 Extend subscriber with MaxDeliveryCount functionality

mszymeshop:
I have tested the following scenarios:

- new subscriber added via API with no MaxDeliveryCount (default 10 used)
- new subscriber added via API with valid and invalid MaxDeliveryCount
- subscriber updated via API with MaxDeliveryCount
- subscriber updated via API with no MaxDeliveryCount
- manual change in CosmosDB and restart of CH results in proper values in SB
- regular KV entries with no MaxDeliveryCount result in default 10

For each of the above, I have tested that the subscriber in SB is refreshed with the relevant MaxDeliveryCount value.


tabman83 (approved):
Shall we also update the README to include this bit?

mszymeshop:
> Shall we also update the README to include this bit?

Sounds reasonable but there is no documentation on API whatsoever at the moment in there. Do you mean to start it?

tabman83:
> > Shall we also update the README to include this bit?
> 
> Sounds reasonable but there is no documentation on API whatsoever at the moment in there. Do you mean to start it?

I got confused with the EDA config repo README, nevermind!

PR #372, opened 2020-10-28
#92340 - validator to be case insensitive for uri replacements

mszymeshop:


tabman83 (commented):
Woudln't be better to have the replace dictionary as a case insensitive one?

chomick (approved):

PR #374, opened 2020-10-28
91706 subscriber name length validation

dawidkeshop:


src/CaptainHook.Application/Validators/UpsertSubscriberRequestValidator.cs
```diff
@@ -9,7 +9,9 @@ public class UpsertSubscriberRequestValidator : AbstractValidator<UpsertSubscrib
         public UpsertSubscriberRequestValidator()
         {
             RuleFor(x => x.EventName).NotEmpty();
-            RuleFor(x => x.SubscriberName).NotEmpty();
+            RuleFor(x => x.SubscriberName).Cascade(CascadeMode.Stop)
+               .NotEmpty()
+               .MaximumLength(50);
```
chomick, line 14: It is not the subscriber name that is limited, it is a subscription name in service bus, which is built of event name, subscriber name and few other things

chomick: That's why we need to extract the name building to a separate class so we can reuse it here

dawidkeshop: I've added comment to the story, we use directly subscriber name, we don't append event name or any other things there.

mszymeshop: I guess this checks out @dawidkeshop. The concatenation of event and subscriber is used internally for CH Readers, SB subscribers are just SubscriberNames.

mszymeshop (approved):

src/Tests/CaptainHook.Application.Tests/RequestValidators/UpsertSubscriberRequestValidatorTests.cs
```diff
@@ -94,6 +94,16 @@ public void When_SubscriberIsEmpty_Then_ValidationFails(string invalidString)
             result.ShouldHaveValidationErrorFor(x => x.SubscriberName);
         }
 
+        [Fact, IsUnit]
+        public void When_SubscriberIsTooLong_Then_ValidationFails()
```
tabman83, line 98: When_Subscriber_Name_IsTooLong_Then_ValidationFails

tabman83 (approved):

src/Tests/CaptainHook.Application.Tests/RequestValidators/UpsertSubscriberRequestValidatorTests.cs
dawidkeshop: sure

src/Tests/CaptainHook.Application.Tests/RequestValidators/UpsertWebhookRequestValidatorTests.cs
```diff
@@ -63,6 +63,16 @@ public void When_SubscriberIsEmpty_Then_ValidationFails(string invalidString)
             result.ShouldHaveValidationErrorFor(x => x.SubscriberName);
         }
 
+        [Fact, IsUnit]
+        public void When_SubscriberIsTooLong_Then_ValidationFails()
```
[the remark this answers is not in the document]
dawidkeshop: sure

PR #376, opened 2020-10-29
Extended logging for Actor deserialization issue

dawidkeshop:


src/CaptainHook.Common/Telemetry/Actor/ActorStateError.cs
```diff
@@ -0,0 +1,15 @@
+﻿using Microsoft.ServiceFabric.Actors.Runtime;
+using Newtonsoft.Json;
+
+namespace CaptainHook.Common.Telemetry.Actor
+{
+    public class ActorStateError : ActorTelemetryEvent
+    {
+        public ActorStateError(object state, ActorBase actor) : base(actor)
+        {
+            State = JsonConvert.SerializeObject(state);
```
chomick, line 10: what if state is null? will it serialize well?

dawidkeshop: I've checked that. :)
It will produce string: `"null"`.

nikhilgirraj (approved):

chomick (approved):

src/CaptainHook.EventHandlerActor/EventHandlerActor.cs
```diff
@@ -75,6 +75,8 @@ protected override Task OnDeactivateAsync()
 
         public async Task Handle(MessageData messageData)
         {
+            _bigBrother.Publish(new ActorHandleStarted(messageData, this));
```
chomick, line 78: Not sure if I would log everything (?) as it will be rather heavy for each message.
Unless we will remove it shortly? Alternatively, maybe you can log just key details and/or the fact that things are not null etc? Just conscious this is logged for every message...

dawidkeshop: I was thinking to log everything to detect potential issues when e.g.  authentication details of subscriber configuration aren't passed correctly. I think better to log everything for some time than lost some important information during debugging.

mszymeshop (approved):

PR #379, opened 2020-11-06
92430 stop loading subscribers from KV

dawidkeshop:


src/CaptainHook.DirectorService/Program.cs
```diff
@@ -59,10 +58,6 @@ private static async Task Main()
                     .As<ISubscriberConfigurationLoader>()
                     .SingleInstance();
 
-                builder.RegisterType<ConfigurationMerger>()
```
tabman83, line 62: There's a SubscribersKeyVaultProvider on line 58 which needs to be removed as well, cannot highlight it from here

tabman83 (changes requested):

src/CaptainHook.DirectorService/Program.cs
tabman83: Also both ISubscribersKeyVaultProvider.cs and SubscribersKeyVaultProvider.cs needs to be removed

tabman83: And its test files if any

tabman83 (changes requested):

src/CaptainHook.DirectorService/Program.cs
dawidkeshop: Good idea! Removed, as well as KeyVaultConfigurationLoader

tabman83 (approved):
Please hold on to the merge for now

src/CaptainHook.DirectorService/Infrastructure/SubscriberConfigurationLoader.cs
```diff
@@ -11,32 +16,50 @@ namespace CaptainHook.DirectorService.Infrastructure
...
+            }
+
+            try
+            {
+                await Task.WhenAll(tasks);
+            }
+            catch (Exception ex)
             {
-                return subscribersFromKV.Error;
+                return new MappingError("Cannot map Cosmos DB entries", new ExceptionFailure(ex));
```
chomick, line 52: Should you check here which tasks had exceptions and log those?

dawidkeshop: I've changed that part, so now we know the ID of failed subscriber entity.
Logging is done by DirectorService.

The question is - do we really should fail DirectorService if one subscriber configuration is broken?

src/CaptainHook.DirectorService/Infrastructure/Interfaces/LoadingSubscribersResult.cs
```diff
@@ -0,0 +1,31 @@
...
+using CaptainHook.Common.Configuration;
+using CaptainHook.Domain.Results;
+
+namespace CaptainHook.DirectorService.Infrastructure.Interfaces
+{
+    public class LoadingSubscribersResult
+    {
+        public IEnumerable<ErrorBase> Errors { get; }
+        public IEnumerable<SubscriberConfiguration> Subscribers { get; }
+        public bool HaveErrors => Errors.Any();
```
tabman83, line 13: Not meaning to be pedantic, I'd say `HasErrors`

dawidkeshop: Sure, will fix

src/CaptainHook.DirectorService/Infrastructure/Interfaces/LoadingSubscribersResult.cs
```diff
@@ -0,0 +1,31 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using CaptainHook.Common.Configuration;
+using CaptainHook.Domain.Results;
+
+namespace CaptainHook.DirectorService.Infrastructure.Interfaces
+{
+    public class LoadingSubscribersResult
```
chomick, line 9: I didn't think we have to tell the caller what worked what did not? I though we could simply log the individual errors and return correct subscribers?

chomick (commented):
Before we merge this, is there any way to verify (on any env) that there is no longer ANY keyvault config loaded and all of them are already served from DB only?
Do we have some indicator on GET operation on where the config came from?

PR #380, opened 2020-11-10
Additional logging of subscribers existing only in KV

dawidkeshop:


tabman83 (approved):

chomick (approved):
Let's get it out ASAP so we can verify all envs today before the merge, thx

PR #381, opened 2020-11-23
Fix for DLQ payload wrapping

dawidkeshop:


tabman83 (approved):
Shouldn't we be adding a test that verifies that WrapperContract is correctly set for DLQ? Just in case we change that again later.

chomick (approved):
Did you test it locally to ensure it works as expected?

PR #382, opened 2020-11-24
Fixed reader service hash calculation for DLQ payload wrapping

dawidkeshop:


tabman83 (approved):

nikhilgirraj:
Should the client be regenerated as part of this change?

nikhilgirraj (approved):

tabman83:
> Should the client be regenerated as part of this change?

this is not changing contract or controllers.

PR #383, opened 2020-12-01
New event and logging

dawidkeshop:
Now exception details as well as the invoked uri are logged in `FailedWebhookExceptionEvent`

image

### DIFF
--- a/src/CaptainHook.Common/Telemetry/FailedWebhookExceptionEvent.cs
+++ b/src/CaptainHook.Common/Telemetry/FailedWebhookExceptionEvent.cs
@@ -4,9 +4,9 @@ using Newtonsoft.Json;
 
 namespace CaptainHook.Common.Telemetry
 {
-    public class FailedWebhookExceptionEvent : TelemetryEvent
+    public class HttpSendFailureEvent : TelemetryEvent
     {
-        public FailedWebhookExceptionEvent(string uri, Exception exception)
+        public HttpSendFailureEvent(string uri, Exception exception)
         {
             Uri = uri;
             Exception = JsonConvert.SerializeObject(exception, new JsonSerializerSettings()
--- a/src/CaptainHook.Common/Telemetry/FailedWebhookExceptionEvent.cs
+++ b/src/CaptainHook.Common/Telemetry/FailedWebhookExceptionEvent.cs
@@ -1,0 +1,24 @@
+﻿using System;
+using Eshopworld.Core;
+using Newtonsoft.Json;
+
+namespace CaptainHook.Common.Telemetry
+{
+    public class FailedWebhookExceptionEvent : TelemetryEvent
+    {
+        public FailedWebhookExceptionEvent(string uri, Exception exception)
+        {
+            Uri = uri;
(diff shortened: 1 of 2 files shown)


src/CaptainHook.Common/Telemetry/FailedWebhookExceptionEvent.cs
```diff
@@ -0,0 +1,24 @@
+﻿using System;
+using Eshopworld.Core;
+using Newtonsoft.Json;
+
+namespace CaptainHook.Common.Telemetry
+{
+    public class FailedWebhookExceptionEvent : TelemetryEvent
```
tabman83, line 7: This may get logged also for a callback. Then the name is misleading. I suggest something like HttpSendFailureEvent or HttpSendFailureExceptionEvent.

tabman83 (changes requested):

tabman83:
Is there any additional info that we can gather from the httpsender? Like dns for example. I wonder if that is failing because dns issues.

src/CaptainHook.Common/Telemetry/FailedWebhookExceptionEvent.cs
dawidkeshop: I can do that, I just followed the convention with `FailedWebhookEvent` which has no distinction for webhooks/callbacks.

src/CaptainHook.EventHandlerActor/Handlers/GenericWebhookHandler.cs
```diff
@@ -88,7 +90,7 @@ public virtual async Task<bool> CallAsync<TRequest>(TRequest request, IDictionar
             }
             catch (Exception e)
             {
-                BigBrother.Publish(e.ToExceptionEvent());
+                BigBrother.Publish(new FailedWebhookExceptionEvent(uri?.AbsoluteUri, e));
```
chomick, line 93: Should we catch URI transformation exception only here? There may be other reasons the delivery failed and should this be logged in request builder instead?

chomick: Ah, sorry, the issue is not with url validity, but the delivery problem - it's ok then.

tabman83 (approved):

src/CaptainHook.Common/Telemetry/FailedWebhookExceptionEvent.cs
dawidkeshop: I've decided to name it `HttpSendFailureEvent` because `HttpSendFailureExceptionEvent` suggests that this derives from `ExceptionEvent` which is not true.

src/CaptainHook.EventHandlerActor/Handlers/GenericWebhookHandler.cs
dawidkeshop: @chomick - the issue which is root cause of this was `SocketException`. We're hoping that knowing the URI will help us understand the situation a bit better in the future.
There is `throw` in the next line, so everything will be logged also as `ExceptionEvent` by `EventHandlerActor` in exactly same way as it was before.

PR #386, opened 2020-12-16
Fixed sonar cloud bugs

nikhilgirraj:
Bugs - https://sonarcloud.io/project/issues?id=esw.captain-hook&open=AXLsA35OHE1eOavyJyNB&resolved=false&types=BUG

### DIFF
--- a/src/CaptainHook.EventReaderService/EventReaderService.cs
+++ b/src/CaptainHook.EventReaderService/EventReaderService.cs
@@ -260,6 +260,8 @@ namespace CaptainHook.EventReaderService
                             continue;
                         }
 
+                        // False warning for possibility of messages being null. This is handled 
+                        // ReSharper disable once PossibleNullReferenceException
                         foreach (var message in messages)
                         {
                             var messageData = BuildMessageData(message);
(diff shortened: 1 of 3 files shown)


src/CaptainHook.EventReaderService/EventReaderService.cs
```diff
@@ -260,6 +260,8 @@ protected override async Task RunAsync(CancellationToken cancellationToken)
                             continue;
                         }
 
+                        // False warning for possibility of messages being null. This is handled
```
chomick, line 263: The error is about `messages`, not `message` - it is giving warning because some lines above we expect this collection to be null but we do not check it here.
My suggestion - just do this in the loop:
`messages ?? Enumerable<...>.Empty`

nikhilgirraj: The lines 248 and 253 above ensure that `messages` would not be null at this point in code.

chomick (approved):
I think you need to fix the last point, otherwise looks good

src/CaptainHook.EventReaderService/EventReaderService.cs
chomick: I know - but you still will get the warning - so either you change the code or you update the sonar cloud bug :)

PR #388, opened 2021-01-04
101924 telemetry keys improvements

dawidkeshop:
Added support for separate Instrumentation (Platform AI) and Internal (Tooling AI) keys.
Won't be merged until both keys are properly set on all environments.


tabman83 (changes requested):
this should be an opportunity to:
- switch to correct keyvault
- use correct secret key names (see here: https://eshopworld.atlassian.net/wiki/spaces/OPUS/pages/2045673762/KeyVault+Secret+Naming+Convention)

tabman83 (approved):

nikhilgirraj (approved):

PR #389, opened 2021-01-07
102754 update contract with UseHeaders field

dawidkeshop:


src/CaptainHook.Application/Validators/Dtos/OidcAuthenticationDtoValidator.cs
```diff
@@ -10,13 +12,19 @@ public OidcAuthenticationDtoValidator()
         {
             RuleFor(x => x.ClientId).NotEmpty()
                 .WithMessage("'ClientId' must not be empty.");
-            RuleFor(x => x.Scopes).NotEmpty();
-            RuleForEach(x => x.Scopes).NotEmpty();
+            RuleFor(x => x.Scopes).Must(BeSetAccordingToUseHeaders)
+                 .WithMessage("'Scopes' must be defined only if 'UseHeaders' is true");
```
chomick, line 16: Opposite :) Only when false

chomick: We do not use scopes when passing client/secrets in headers

src/CaptainHook.Domain/Entities/OidcAuthenticationEntity.cs
```diff
@@ -22,12 +22,18 @@ public class OidcAuthenticationEntity : AuthenticationEntity
         /// </summary>
         public string[] Scopes { get; }
 
-        public OidcAuthenticationEntity(string clientId, string clientSecretKeyName, string uri, string[] scopes)
+        /// <summary>
+        /// Use headers
```
chomick, line 26: I would just add one more sentence to explain the purpose of this flag (passing client id/secret in headers and ignoring scopes & grant type)

src/CaptainHook.Application/Validators/Dtos/OidcAuthenticationDtoValidator.cs
```diff
@@ -10,13 +12,19 @@ public OidcAuthenticationDtoValidator()
...
             RuleFor(x => x.ClientSecretKeyName).NotEmpty()
                 .WithMessage("'ClientSecretKeyName' must not be empty.");
             RuleFor(x => x.Uri).Cascade(CascadeMode.Stop)
                 .NotEmpty()
                 .SetValidator(new UriValidator(false));
         }
+
+        private bool BeSetAccordingToUseHeaders(OidcAuthenticationDto authDto, List<string> scopes)
+        {
+            return authDto.UseHeaders ? scopes == null || !scopes.Any() : scopes != null && scopes.Any();
```
chomick, line 27: The implementation seems ok, just the message is wrong

chomick (approved):

PR #392, opened 2021-01-11
102954 Use Azure API to find single topic instead of listing all

dawidkeshop:


tabman83 (commented):
I know it doesn't look like part of the scope of this story, but can we apply the same improvement here? 

https://github.com/eShopWorld/captain-hook/blob/dfae9ba5b90d87b784ef6bb49cceb1bbd9a89d43/src/CaptainHook.Common/ServiceBus/ServiceBusManager.cs#L77

CC: @chomick

tabman83 (approved):

PR #399, opened 2021-01-19
Added DebugView

dawidkeshop:


dawidkeshop:
This was only for test purposes.

PR #400, opened 2021-01-19
KEYVAULT_URL defined for all services and environments

dawidkeshop:


tabman83 (approved):
The KEYVAULT_URL setting is only temporary because we are running off Github pipelines. It will be gone when we move to ADO + DevOps template. Would you please create a TD for such removal when the move happens.

